PR #43, opened 2022-01-06
possible fix to react hydration error? 

nglglhtr:
image

Using the window component in rendering causes this error, possible solution [in the docs](https://nextjs.org/docs/messages/react-hydration-error) is to use useEffect / useState

Not sure why do we need to have a check like this though?

### DIFF
--- a/packages/private/src/connectors/injectedConnector.ts
+++ b/packages/private/src/connectors/injectedConnector.ts
@@ -25,7 +25,7 @@ export class InjectedConnector extends Connector<
     super({ ...config, options: undefined })
 
     let name = 'Injected'
-    if (typeof window !== 'undefined' && window.ethereum) {
+    if (window.ethereum) {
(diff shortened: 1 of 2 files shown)


iDuuck:
> Not sure why do we need to have a check like this though?
The problem here is, that we are having Coinbase Wallet as injected wallet, and also MetaMask as injected wallet (before they nerfed the iOS app, even Trust Wallet shared the same API). Hence we need to check for the name.

Though, my proposal would be just to use the name "injected", since this is, what it is, and just let the developer see what he does with this (i.e. creating two buttons, who both use the injected connector, but just for the sake of user experience)

tmm:
@nglglhtr thanks for spending time to investigate. Added some thoughts on [this issue](https://github.com/tmm/wagmi/issues/51#issuecomment-1008157716).

PR #64, opened 2022-01-10
docs: create remix example

margalit:
This MR adds a [Remix.run](https://remix.run) example.

## Provider support
- [x] Metamask
- [x] WalletConnect
- [x] WalletLink

## Details
[WalletLink](https://github.com/walletlink/walletlink/issues/56) and WalletConnect both rely on polyfills of native Node globals ("process") and modules ("buffer", "events") which are not automatically included by esbuild (Remix’s compiler).

I was able to create a working version by [patching](https://github.com/ds300/patch-package) the Remix compiler code with [@esbuild-plugins/node-globals-polyfill](https://www.npmjs.com/package/@esbuild-plugins/node-globals-polyfill) and [@esbuild-plugins/node-modules-polyfill](https://www.npmjs.com/package/@esbuild-plugins/node-modules-polyfill).

WalletLink relies on an older version of `eth-block-tracker`, which in turn relies on an old version of `safe-event-emitter` which uses an old commonjs syntax require syntax `events/` (with a trailing slash) which doesn't seem to get picked up and polyfilled by the esbuild plugin. I've explicitly added `safe-event-emitter` and patched that too.

## References 🙇🏽
- https://github.com/remix-run/remix/issues/669#issuecomment-1010152216
- https://github.com/evanw/esbuild/issues/1921#issuecomment-1013698430


tmm:
@margalit made it so examples are now standalone so no need to worry about local package resolution.

also, fixed the entrypoints error (I had set it up incorrectly previously), but am now getting this error when trying to run remix locally:

image

Once this is fixed, we are ready to go!

margalit:
Thanks @tmm! I was able to get this error to go away by assigning `window.process = process` as part of the client entry logic but I think I'll need to dig a little more into [Remix’s module constraints](https://remix.run/docs/en/v1/guides/constraints) to figure out how to figure out how to properly expose the `events` and `process` modules that `WalletConnectConnector` and `WalletLinkConnector` rely on:

image

tmm:
> I'll need to dig a little more into [Remix’s module constraints](https://remix.run/docs/en/v1/guides/constraints) to figure out how to figure out how to properly expose the `events` and `process` modules that `WalletConnectConnector` and `WalletLinkConnector` rely on:
> 
> image

I had to do [something similar](https://github.com/tmm/wagmi/blob/main/examples/vite-react/index.html#L9-L15) for the vite-react example.

margalit:
Okay! With a couple of patches this is now working, see MR for details.

PR #66, opened 2022-01-10
feat: add useSigner hook

zakangelle:
This PR adds a `useSigner` convenience hook which provides the `signer` object of the connected account. This streamlines contract w/ signer instantiation by tucking away the async logic needed to access the `signer`.

### Example usage

```tsx
import { useSigner } from 'wagmi'

const App = () => {
  const [{ data, error, loading }, getSigner] = useSigner()

  const contract = useContract({
    addressOrName: '0xa50f703771d04d9e76e03b20720539fc014aaa40',
    contractInterface: ERC20ABI,
    signerOrProvider: signer,
  })

  return null
}
```


tmm (changes requested):
Thanks for taking this on!

We'll want to add a test before merging this in (See [`useAccount.test.ts`](https://github.com/tmm/wagmi/blob/main/packages/react/src/hooks/accounts/useBalance.ts#L97-L107) for connecting account, etc.).

Also, we want the hook to update when the chain changes. Check out how this is done in [`useBalance`](https://github.com/tmm/wagmi/blob/main/packages/react/src/hooks/accounts/useBalance.ts#L97-L107) for more info.

zakangelle:
@tmm PR feedback addressed [here](https://github.com/tmm/wagmi/pull/66/commits/e7dc770dabfd43ac35516487cbea4c92264b096d) and hook tests are [here](https://github.com/tmm/wagmi/pull/66/files#diff-98426becdb6a05357678cf7496e0ed25564fc0991665fe265ca4535422bde270) - lemme know how that looks and if I'm missing anything.

PR #77, opened 2022-01-13
fix: cant install package directly from github

clifton:
fix for `error Can't add "root": invalid package version undefined`


clifton:
ah, my bad. not used to npm hierarchical package structures

PR #112, opened 2022-01-26
Add support for Avalanche chain

rpavlovs:
This adds Avalanche Mainnet and Fuji chains to constants


tmm:
@arjanvillon this was removed when `0.3` was released. Wanted to trim down until there is a scalable process for adding chains. In the meantime, you can create a custom chain by conforming to the `Chain` type.

PR #135, opened 2022-02-05
fix: state update warnings, incomplete hooks reactivity

bpierre:
This PR prevents state to be set on unmounted or outdated render contexts by refactoring the hooks around a new `useCancel()` utility hook, which allows to manage the cancelled state of a query. It automatically calls its cancellation callback when the component unmounts, and can be manually called to invalidate it in other cases (typically by returning it in any `useEffect()` calling the query function defining its callback).

This utility is used by multiple hooks, which removes the warnings caused by the library trying to update state on unmounted components, or on outdated render contexts.

Note that since these callbacks are also [returned by the hooks](https://github.com/tmm/wagmi/blob/a6dc92c769ef64badc6c9d0ea171b2e8abb5baca/packages/react/src/hooks/accounts/useBalance.ts#L142) and can be called independently to fetch results, they don’t interrupt when cancelled. Instead, they just stop updating the local state.

These changes also ensure some `useEffect()` hooks react to changes on all the dependencies they are using, removing the need to disable the `react-hooks/exhaustive-deps` ESLint rule. Some unnecessary cancel checks [have also been removed](https://github.com/tmm/wagmi/pull/135/files#diff-efee3705b67f0357e970b517bc73227ef12d2e0215cd9a161a233d4802e7bdb1L64-L65) from these `useEffect()` hooks.

Note on tests: I tried different ways to test this, but I couldn’t find a reliable way to check if React emits a warning if the state of an unmounted component is set.

@tmm let me know what you think!


packages/react/src/hooks/accounts/useConnect.ts
```diff
@@ -35,35 +42,38 @@ export const useConnect = () => {
...
       }
     },
-    [globalState.connector, setGlobalState, setLastUsedConnector],
+    [cancelQuery, globalState.connector, setGlobalState, setLastUsedConnector],
   )
 
   // Keep connector in sync with global connector
   React.useEffect(() => {
-    let didCancel = false
-    if (didCancel) return
```
bpierre, line 55: `didCancel` will always be false here, so the `if (didCancel)` is unnecessary.

packages/react/src/hooks/accounts/useConnect.ts
```diff
@@ -35,35 +42,38 @@ export const useConnect = () => {
...
-
     setState((x) => ({
       ...x,
       connector: globalState.connector,
       error: undefined,
     }))
-
-    return () => {
-      didCancel = true
-    }
```
bpierre, line 65: And this can be safely removed since the only use of `didCancel` here is in the `if` check right after its definition.

packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -28,29 +28,34 @@ export const useSigner = ({ skip }: Config = {}) => {
...
   React.useEffect(() => {
-    if (skip) return
-
-    let didCancel = false
-    if (didCancel) return
-    getSigner()
-
-    return () => {
-      didCancel = true
+    if (!skip && cacheBuster) {
```
bpierre, line 54: `cacheBuster` is always truthy since it starts at `1`, this is just to make ESLint happy.

tmm:
Looks great and will do a proper review in a bit!

You might also be interested in checking out #85

bpierre:
Thanks! :hugs:

I just opened a discussion that also takes inspiration from React Query, for `refetch()`: https://github.com/tmm/wagmi/discussions/140

tmm:
Looking good! Waiting until the RFC drops (sometime this week) for the discussion (https://github.com/tmm/wagmi/discussions/85) I linked above since that could have some implications for how this is set up.

bpierre:
Regarding https://github.com/tmm/wagmi/pull/162#issuecomment-1038307232:

I am not sure what the merging strategy should be, since #135 is not merged yet (I would rather have all the hooks updated at once rather only this one), and #153 will also be an opportunity to fix these issues in a different way, so we might want to keep going with the same structure in the meantime.

A safe approach could also be to fix the `setState()` calls so the library stops emitting warnings, and remove the unneeded `didCancel` checks, but leave the dependencies fix for later since this will alter the hooks behavior.

/cc @tmm @shunkakinoki

tmm:
Agree! #153 will be great, but in the meantime, this is a nice improvement. Once merged in, #162 can follow along.

PR #153, opened 2022-02-09
rfc: Deterministic states & caching

jxom:
[See formatted](https://github.com/jxom/wagmi-1/blob/ad88f59d4a3fed30e5f2b7a5d8c58cc0132dd62d/rfcs/001-deterministic-states-and-caching.md)

### DIFF
--- a/rfcs/001-deterministic-states-and-caching.md
+++ b/rfcs/001-deterministic-states-and-caching.md
@@ -576,7 +576,7 @@ const [{
 By default, caching on `useWaitForTransaction` will be disabled.
 
 
-### Action hooks
+## Action hooks
 
 The following hooks will see a common set of changes to their APIs:
 
--- a/rfcs/001-deterministic-states-and-caching.md
+++ b/rfcs/001-deterministic-states-and-caching.md
@@ -92,9 +92,15 @@ const [...] = useAccount({ cacheTime: 0 });
 
 # Motivation
 
+The motivation of this RFC can be seen in the below discussion:
+
 https://github.com/tmm/wagmi/discussions/85
 
-TODO
+TLDR; 
+
+- Non-deterministic & implicit states can lead to "impossible" states, which can cause UX bugs quite easily,
+- No idle state & implicit states lead to a confusing DX,
+- Most dapps experience a "flash of loading wallet/resolving ens" experience.
 
 # Detailed design
 
@@ -145,6 +151,28 @@ const [{ data, error, loading }, getBalance] = useBalance({
 ```
 
 By default, if `enabled` is not provided, then it is truthy & it still fetches on mount.
+
+## React Query & `Provider`
+
+This implementation will use [React Query](https://react-query.tanstack.com/), which means we will need to provide a query client (via react-query's `QueryClientProvider`) within wagmi's `Provider`.
+
+## Storage persistence
+
+I think maybe we should have storage persistence an opt-in feature, that people can toggle on from the `Provider`. Storage persistence is available via [react-query's `persistQueryClient`](https://react-query.tanstack.com/plugins/persistQueryClient)
+
+### Usage
+
+```tsx
+import { Provider, createPersistor } from 'wagmi';
+
+const localStoragePersistor = createPersistor({ storage: window.localStorage })
+
+const App = () => (
+  <Provider persistor={localStoragePersistor}>
+    <YourRoutes />
+  </Provider>
+)
+```
 
 ## Read hooks
 
@@ -663,38 +691,21 @@ const [{
 
 # Drawbacks
 
-Why should we *not* do this? Please consider:
+There are 2 immediate drawbacks:
 
-- implementation cost, both in term of code size and complexity
-- whether the proposed feature can be implemented in user space
-- the impact on teaching people React
-- integration of this feature with other existing and planned features
-- cost of migrating existing React applications (is it a breaking change?)
-
-There are tradeoffs to choosing any path. Attempt to identify them here.
+- Bundle size
+  - Adding `react-query` could [add up to 13kB](https://bundlephobia.com/package/react-query@3.34.14) to the bundle size. However, `react-query` is tree-shakable, so it is possible that this could be lower.
+- Increased API surface
+  - With the addition of new async state variables, and caching options, this will increase the API surface layer. But I don't think it's too much of a concern.
 
 # Alternatives
 
-What other designs have been considered? What is the impact of not doing this?
+There are a few alternatives that were mentioned in the [precurser discussion thread](https://github.com/tmm/wagmi/discussions/85). They were:
 
-# Adoption strategy
-
-If we implement this proposal, how will existing React developers adopt it? Is
-this a breaking change? Can we write a codemod? Should we coordinate with
-other projects or libraries?
+- [Rolling our own states & cache](https://github.com/tmm/wagmi/discussions/85#discussioncomment-1977448),
+- Use an alternate async state lib such as SWR,
+- [Have separate entries for async state libraries](https://github.com/tmm/wagmi/discussions/85#discussioncomment-1978779)
 
 # How we teach this
 
-What names and terminology work best for these concepts and why? How is this
-idea best presented? As a continuation of existing React patterns?
-
-Would the acceptance of this proposal mean the React documentation must be
-re-organized or altered? Does it change how React is taught to new developers
-at any level?
-
-How should this feature be taught to existing React developers?
-
-# Unresolved questions
-
-Optional, but suggested for first drafts. What parts of the design are still
-TBD?
+Will need to add the new API to the docs, and think about the structure of the docs if we want to separate hooks by "action hooks" & "read hooks".
--- a/rfcs/001-deterministic-states-and-caching.md
+++ b/rfcs/001-deterministic-states-and-caching.md
@@ -1,0 +1,189 @@
+- Start Date: 04/02/2022
+- RFC PR:
+
+# Summary
+
+This RFC proposes the amendment of the hooks API within `wagmi` to include **finite & deterministic states**, as well as enabling the ability to **cache** appropriate hooks & data within the library.
+
+This proposal heavily integrates with [`react-query`](https://react-query.tanstack.com/) internally whilst still providing a minimal API surface at the face of `wagmi`.
+
+# Basic example
+
+## Example 1 (finite states)
+
+A very high-level example of the proposal could use the `useSignMessage` hook for signing a message with the connected account.
+
+Currently, the API looks as such:
+
+```tsx
+import { useSignMessage } from 'wagmi'
+
+const App = () => {
+  const [{ data, error, loading }, signMessage] = useSignMessage({
+    message: 'gm wagmi frens',
+  })
+
+  if (!data)
+    return (
+      <button disabled={loading} onClick={async () => await signMessage()}>
+        Sign message
+      </button>
+    )
+
+  return (
+    <div>
+      {data && <div>Signature: {data}</div>}
+      {error && <div>Error signing message</div>}
+    </div>
+  )
+}
+```
+
+This proposal adds a set of states (`isIdle`, `isLoading`, `isSuccess` & `isError`) that represents a finite state machine (only one flag can be truthy at the one time), which makes the following render code easier to read & follow:
+
+```tsx
+import { useSignMessage } from 'wagmi'
+
+const App = () => {
+  const [{ data, error, isIdle, isLoading, isSuccess, isError }, signMessage] = useSignMessage({
+    message: 'gm wagmi frens',
+  })
+
+  return (
+    <div>
+      {isIdle && (
+        <button disabled={loading} onClick={async () => await signMessage()}>
+          Sign message
+        </button>
+      )}
+      {isLoading && <div>Signing...</div>}
+      {isSuccess && <div>Signature: {data}</div>}
+      {isError && <div>Error signing message</div>}
+    </div>
+  )
+}
+```
+
+## Example 2 (caching)
+
+Some hooks in this library could be cacheable, such as `useAccount`, `useBalance`, `useNetwork`, etc. 
+
+There are no examples for caching, as it will be performed under-the-hood. 
+
+However, users' should be able to do either of the following:
+
+- change the cache expiry (time the data remains in the cache, cached results are shown to the user straight away),
+
+```tsx
+const [...] = useAccount({ cacheTime: 60000 });
+```
+
+- change the stale expiry (time when the data is considered stale, and then should perform a background refresh),
+
+```tsx
+const [...] = useAccount({ staleTime: 10000 });
+```
+
+- disable the cache completely.
+
+```tsx
+const [...] = useAccount({ cacheTime: 0 });
+```
+
+# Motivation
+
+https://github.com/tmm/wagmi/discussions/85
+
+TODO
+
+# Detailed design
+
+For the purpose of clarity, I think we should introduce the terminology of "read hooks" and "action hooks" or something similar.
+
+This will allow the consumer to know which hooks will invoke on mount (read hooks), and the ones whos actions won't be invoked on mount (action hooks).
+
+I feel like the naming of read hooks such as `useAccount`, `useBalance`, `useNetwork` are fine, however I reckon it would be nice to be more explicit with action hook naming to include intending verb (i.e. `useTransactionSend` instead of `useTransaction`). This also aligns with existing hooks such as `useContractWrite` and `useEnsLookup`.
+
+### Read hooks
+
+The following hooks will see a common set of changes to their APIs:
+
+#### Arguments
+
+The following set of **optional** arguments will be added to the read hooks:
+
+- `cacheTime: number`: The time (in ms) that the data remains in the cache, cached data is displayed initially to the user while a background fetch is being performed.
+- `staleTime: number`: The time (in ms) after which the data is considered stale, and a background refetch is needed.
+- `onSuccess: (data: TData) => void`: A callback function to be invoked when the data is successfully fetched.
+- `onError: (error: TError) => void`: A callback function to be invoked when an error has been thrown.
+- `onSettled: (data?: TData, error?: TError) => void`: A callback function to be invoked when the data has either been fetched successfully or not (kinda like `finally`).
+- `suspense: boolean`: Enable the hook to use suspense
+
+#### Return values
+
+The `loading` flag will be removed in favour of the following set of deterministic states:
+
+- `isLoading`: A flag to indicate that the data is loading for the first time
+- `isReloading`: A flag to indicate that the data is loading in the background, and showing cached success/error data in the meantime (meaining this flag can be truthy when isSuccess/isError is truthy).
+- `isSuccess`: A flag to indicate the data has been fetched successfully.
+- `isError`: A flag to indicate that an error was thrown upon fetching the account
+
+#### `useAccount` 
+
+The new `useAccount` API will look like such:
+
+```tsx
+const [{
+  data,         
+  error,        
+  isLoading,    
+  isReloading,  
+  isSuccess,    
+  isError       
+}, disconnect] = useAccount({
+  fetchEns?: boolean,
+  cacheTime?: number
+})
+```
+
+TODO
+
+### Action hooks
+
+# Drawbacks
+
+Why should we *not* do this? Please consider:
+
+- implementation cost, both in term of code size and complexity
+- whether the proposed feature can be implemented in user space
+- the impact on teaching people React
+- integration of this feature with other existing and planned features
+- cost of migrating existing React applications (is it a breaking change?)
+
+There are tradeoffs to choosing any path. Attempt to identify them here.
+
+# Alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Adoption strategy
+
+If we implement this proposal, how will existing React developers adopt it? Is
+this a breaking change? Can we write a codemod? Should we coordinate with
+other projects or libraries?
+
+# How we teach this
+
+What names and terminology work best for these concepts and why? How is this
+idea best presented? As a continuation of existing React patterns?
+
+Would the acceptance of this proposal mean the React documentation must be
+re-organized or altered? Does it change how React is taught to new developers
+at any level?
+
+How should this feature be taught to existing React developers?
+
+# Unresolved questions
+
+Optional, but suggested for first drafts. What parts of the design are still
+TBD?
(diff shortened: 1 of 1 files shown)


sammdec:
This looks great to me! Amazing work!

shunkakinoki:
Thank you for your rfc!

As someone who is using `swr`, it would be great if we can consider both caching approaches as listed here

- [Have separate entries for async state libraries](https://github.com/tmm/wagmi/discussions/85#discussioncomment-1978779)

gosseti:
This proposal looks fantastic, great work!

rfcs/001-deterministic-states-and-caching.md
```diff
@@ -0,0 +1,716 @@
...
+
+## A note on hook categorization
+
+For the purpose of clarity, I think we should introduce the terminology of "read hooks" and "action hooks" or something similar.
+
+This will allow the consumer to know which hooks will invoke on mount (read hooks), and the ones whos actions won't be invoked on mount (action hooks).
+
+## A note on hook naming
+
+I feel like the naming of read hooks such as `useAccount`, `useBalance`, `useNetwork` are fine, however I reckon it would be nice to be more explicit with action hook naming to include intending verb (i.e. `useTransactionSend` instead of `useTransaction`). This also aligns with existing hooks such as `useContractWrite` and `useEnsLookup`.
```
tmm, line 117: Agree about the naming change for `useTransaction` (and action hooks in general)

rfcs/001-deterministic-states-and-caching.md
```diff
@@ -0,0 +1,716 @@
...
+### Return values
+
+The `loading` flag will be removed in favour of the following set of deterministic states:
+
+- `isIdle`: A flag to indicate that the promise has not yet been invoked, and data has not been loaded yet.
+- `isLoading`: A flag to indicate that the data is loading for the first time
+- `isReloading`: A flag to indicate that the data is loading in the background, and showing cached success/error data in the meantime (meaining this flag can be truthy when isSuccess/isError is truthy).
+- `isSuccess`: A flag to indicate the data has been fetched successfully.
+- `isError`: A flag to indicate that an error was thrown upon fetching the account
+
```
tmm, line 201: Can we also expose `status` too? Think it's a useful prop for folks that don't like booleans.

tmm (commented):
Overall looks great! Some details to chat about.

### Consistent Defaults


Better to have opinionated defaults for specific hooks (i.e. `useEnsAvatar` default enabled, `useFeeData` default disabled, etc.) versus the same defaults?

Think it makes sense that “we” (the library authors) use our industry knowledge to help assist folks around cache duration/location, etc., but would it be too [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) if defaults aren't consistent?

### Testing


How do we want to handle tests for the increased surface API? Do we test config arguments that are React Query config? Probably doesn’t make sense to test all the options that are just passed through to React Query, but maybe some.

### Dependency Type


Should React Query be a core or peer dependency? ethers.js is a peer dependency. Could see a case for either approach.

### Additional Configuration


Moving forward, how do we decide what React Query features make it into wagmi or not? Should we support additional hook-level configuration (i.e. `retry`, `refetchOnWindowFocus`, etc)?

Do we want to expose query keys (or functions for getting array keys)? Definitely more pro, but could help folks tune things to their liking

jxom:
> ### Consistent Defaults
> Better to have opinionated defaults for specific hooks (i.e. `useEnsAvatar` default enabled, `useFeeData` default disabled, etc.) versus the same defaults?
> 
> Think it makes sense that “we” (the library authors) use our industry knowledge to help assist folks around cache duration/location, etc., but would it be too [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) if defaults aren't consistent?

Yeah absolutely. I guess the objective of this PR is to essentially make the UX of wallet & contract interaction better, and to achieve this, we need to enforce opinionated defaults. However, it could be a little confusing from a DX perspective if we have inconsistent defaults (some hooks are cached, while some are not), I guess there are a couple of alternatives at the top of my mind:

1. Disable the cache by default, and make it opt-in on a per-hook basis (a user can set a `cacheEnabled` flag or something to enable the cache),
1. Just leave it as is, per RFC spec, to push & assist developers with opinionated defaults, but clearly state in the docs on a per-hook basis which ones are cached, and which ones are not.

> ### Testing
> How do we want to handle tests for the increased surface API? Do we test config arguments that are React Query config? Probably doesn’t make sense to test all the options that are just passed through to React Query, but maybe some.

Yeah true, maybe we could just test the more "critical" functionality like the new state vars & ensuring caching works as intended?

> ### Dependency Type
> Should React Query be a core or peer dependency? ethers.js is a peer dependency. Could see a case for either approach.

If `ethers` is a peer dep, then I think it might make sense for `react-query` to also be a peer dep. Though, my only concern, is that there could be users who consume an alternative async state library (such as SWR), so it might be weird having both deps in their project. In that case, it might make sense to have it as a core dependency to hide the `react-query` surface entirely from them. 

> ### Additional Configuration
> Moving forward, how do we decide what React Query features make it into wagmi or not? Should we support additional hook-level configuration (i.e. `retry`, `refetchOnWindowFocus`, etc)?

Yeah, definitely see an advantage in exposing more React Query features, but didn't want the RFC to feel too overwhelming. Maybe we could extend the config arg to accept RQ config that make sense for this library?

> Do we want to expose query keys (or functions for getting array keys)? Definitely more pro, but could help folks tune things to their liking

Yeah, totally, I think it would be beneficial if consumers want to programmatically interact with the `react-query` query client. However, I feel like we should blanket the query client with our own API, as the RQ terminology might not align with this library, might be overwhelming for users to learn RQ, and we might want to constrain the API. For example, we might want to have query client functionality more specific to our "read" hooks:

- `queryClient.getQueryData` -> `wagmiClient.getData(useAccount)`, `wagmiClient.getData(useBalance, { addressOrName: '...' })`, etc
- `queryClient.prefetchQuery` -> `wagmiClient.prefetch(useAccount)`, `wagmiClient.prefetch(useBalance, { addressOrName: '...' })`, etc
- `queryClient.refetchQueries` -> `wagmiClient.refetch()`, `wagmiClient.refetch(useAccount)`, `wagmiClient.refetch(useBalance, { addressOrName: '...' })`, etc
- etc etc for other `queryClient` stuff

Not sure if that should be in-scope for an MVP of this? Or should I extend this RFC to touch the surface of this?

jxom:
> @jxom
> 
> Thank you for your rfc!
> 
> As someone who is using `swr`, it would be great if we can consider both `react-query` & `swr` caching approaches as listed here, having different package entrypoints and being peer dependency agnostic
> 
> One of the many strong points about this library is that its only dependency is `ethers`, and this leads to a very fast and smooth, while being composible lib
> 
> ref:
> 
> * [Have separate entries for async state libraries](https://github.com/tmm/wagmi/discussions/85#discussioncomment-1978779)

Yeah, the alternative of having separate entrypoints has definitely been a valid consideration! But I think the downsides of this approach outweighed the upsides. We need to keep in mind that most users' aren't even using an async state lib anyway, so having the option of picking between using `react-query` or `swr` could be a bit overwhelming. IMO, I think it would be less overwhelming to the end user (developer) if they don't have to make this choice, and it's already built in. Plus, having it built in means we can change up the API terminology a bit to make more sense for `wagmi` (so the consumer doesn't have to worry about learning an additional RQ or SWR API) & also enforce a constrained/minimal API surface.

rfcs/001-deterministic-states-and-caching.md
jxom: Yes, great idea.

sammdec:
My 2 cents re: Consistent Defaults
* I would rather go the more opinionated route but we are providing a set of consistent configurations for users who want to change, so while the defaults are not consistent the api to change them is.

tmm:
Going to give the weekend for folks to share additional thoughts :)

In the meantime, I will start another discussion thread around implementation (who, what, how, etc).

sammdec:
@tmm I can give some hours to the project to help out with implementation, doc updates etc. Wont be available next week but I can start from the week after.

jxom:
> Going to give the weekend for folks to share additional thoughts :)
> 
> In the meantime, we can start chatting about implementation.* We'll need to update the hooks, tests, docs, and examples. I'm down to work on this and would be great if there were one or two others that were also interested. Could make sense for folks to work on specific hooks and update them across surfaces. As part of this work, I'm also down to expand the core team and give folks access to the repo.
> 
> Anyone interested in being involved in implementation? What do we think about assigning hooks to work on versus doing things more horizontally across surfaces?
> 
> *We probably also want to figure out the best way to message changes, but can do that later.

Yeah, I'm defs down to help out with this.

shunkakinoki:
I'm also def down to help! 

Also would love for the opportunity to chat and contribute regarding this.

shunkakinoki:
>  That said, keeping things lightweight resonates with me (why I suggested entrypoints) and I've also been in situations before where I was frustrated with decisions maintainers made for OSS projects.

Speaking of, i do still stand by the above

As of the moment (2/10/2022), npm downloads of [react query](https://www.npmjs.com/package/react-query) vs [swr](https://www.npmjs.com/package/swr) ration is roughly about 2:1,
1,120,187 vs 591,196
However, the package size diff is 2.25 MB vs 228 kb (potentially a bottleneck)

As a swr enthusiast, it would be great if we can move forward `wagmi` to having multiple entrypoints, despite its technical challenges - love to be of help too.

sammdec:
One idea that came up in conversation earlier was what if wagmi “core” was just some basic functions that wasn’t really tied to hooks or a particular data mgmt library, this would mean it could be used it any context. Then you could build some wrapper that would essentially be a thin abstraction around react-query or swr that simply consumed the core library as async functions.
That way we wouldn’t have to worry about creating some complex adapter system and instead could more easily test the core functions and treat them as black boxes that can then simply be wrapped by a consumption layer that handles caching etc. 

The one thing I havnt figured out is how you would share state as the library currently uses a context

jxom:
Yeah, just to give a bit of background, I guess I proposed `react-query` as it objectively meets the criteria of what we want from a DX perspective in `wagmi`, albeit the larger bundle size. The main criterion being deterministic states (react-query has first class support for this). Along with other implicit criteria being: trivial programmatic retrieval and mutation of the cache, render batching, garbage collection & partial query matching.

However, thinking about it, I agree there should be an option to still consume this library at a complete minimal surface, as there could well be minimalistic consumers. I really think @sammdec has a valid point that we could extract the core functionality out of the hooks, and consumers can consume that in whatever async state lib they want (swr, react async, react loads, etc). Which is similar to having multiple entrypoints, but a little bit more explicit. But if we were to move this to be more agnostic, and moved out of hooks into functions, I think we would need to put a bit of thought into how we deal with reactions & events (such as switching wallets, chains, etc). 

Could be something like so:

```tsx
import { fetchAccount, watchAccount } from 'wagmi/core';
import useSWR from 'swr';

function Example() {
  const { data, mutate } = useSWR('account', fetchAccount);
  
  useEffect(() => {
    // If an event is emitted to change the account, then broadcast a revalidation.
    const unwatch = watchAccount(mutate);
    return () => unwatch();
  }, []);
  
  ...
}
```

sammdec:
What about using something like zustand to contain the state and events. https://github.com/pmndrs/zustand#using-zustand-without-react
Therefore you can still have nice abstracted state without having to get into sticky use effect hooks etc 
It is also a very tiny library so would add minimal code

bpierre:
> I really think @sammdec has a valid point that we could extract the core functionality out of the hooks, and consumers can consume that in whatever async state lib they want (swr, react async, react loads, etc). Which is similar to having multiple entrypoints, but a little bit more explicit. But if we were to move this to be more agnostic, and moved out of hooks into functions, I think we would need to put a bit of thought into how we deal with reactions & events (such as switching wallets, chains, etc).

I completely agree with the idea of extracting the core functionality out of the hooks, this is the main idea behind this proposal I made the other day: https://github.com/tmm/wagmi/discussions/140 (alongside replacing the error prone hybrid functions like `getBalance()` by a `refetch()` function).

### On the modules & exports strategy

I think having a core version of the library (`wagmi/core`) is a great idea. Using a subdirectory might be problematic to support both ESM and CJS, since we can only use the package.json fields for the main module (`wagmi`). That would remove the possibility to support Node with CJS imports, which I think would be great to have for a core version of the library.

I think wagmi could be opinionated about having first class support for React, and keep providing the React + core exports of the library under the main module (`wagmi`). We could then use an npm scope to export the core version as a separate package, e.g. `@wagmi/core`, and other versoins of the library like `@wagmi/vue`. That way, we could use package.json files for the entry points of each version, and support both ESM + CJS among other things.

Note: if we take care of having the library fully ready for tree shaking (I can help with this), and for consumers using a bundler with tree shaking enabled, then using the core exports from `wagmi` should produce the same result in the final bundle than using `@wagmi/core`, which is a nice thing to have. The documentation could recommend using `@wagmi/core` to Node users (using CJS imports) and to users that don’t know if their bundler supports tree shaking.

### On exposing React Query features

Otherwise I agree with @jxom, I think it makes sense to use React Query especially since the size won’t be impacted much by it (vs. swr) in the current state of things. But do we want to expose specific React Query features, or only a subset of its result object? e.g. [the proposed API for storage persistence](https://github.com/jxom/wagmi-1/blob/ad88f59d4a3fed30e5f2b7a5d8c58cc0132dd62d/rfcs/001-deterministic-states-and-caching.md#storage-persistence) is coming straight from React Query, which means that any change in React Query would require wagmi to follow and break its own API.

Another possibility would be to expose these features but express them in a more “wagmi way”:

```tsx
import { Provider } from 'wagmi'

const App = () => (
  <Provider persist> // uses localStorage.window if `true`
    …
  </Provider>
)
```

Which should be easy to keep maintaining even if React Query completely changes the way persistence is expressed in their API.

#### Full access to React Query features?

What would you guys think about (optionally) letting users gain as much access as possible to react-query? Users with advanced cases could define their own `QueryClient`, and the hooks could take a `reactQuery` parameter that would let users access set any [`useQuery()` option](https://react-query.tanstack.com/reference/useQuery), taking priority over the ones we would expose by default:

```tsx
const [{ data, error, loading }] = useBalance({
  addressOrName: name || address,
  formatUnits: 'gwei',
  enabled: Boolean(name || address),
  reactQuery: {
    refetchOnReconnect: false, // advanced option requiring to pass `reactQuery` 
    enabled: false, // takes priority over `enabled` above
  }
})
```

And we could do the same in the returned values:

```tsx
const [{ data, error, loading, reactQueryResult }] = useBalance({
  addressOrName: address,
})

// reactQueryResult would contain the full object as returned by `useQuery()`
```

I just wanted to put this here to discuss this possibility, but I am not sure how I feel about it personally. It would certainly give users great flexibility over the react-query behavior, but it would make parts of the wagmi API dependent on the evolution of the React Query API. It would also make the implementation less flexible, e.g. moving from `useQuery` to `useQueries()` internally to optionally support multiple results (see [the example](https://github.com/tmm/wagmi/discussions/140) under “Less flexibility when using the hook” in my proposal) would have to be made with great care to break as little as possible the API for users. I think this might cause more issues than it would solve, and we can add it at any point in the future if it becomes needed, so it’s probably not worth adding from the start.

I feel a bit differently about letting users define their own `QueryClient`, because it wouldn’t really get in the way and clearly feel like an advanced part of the API:

```tsx
import { Provider as Wagmi, QueryClient } from 'wagmi'

const queryClient = new QueryClient({ /* … */ })

const App = () => (
  <Wagmi queryClient={queryClient}>
    …
  </Wagmi>
)
```

…but an issue is that it makes little sense to expose `QueryClient` without the rest of the React Query library. And providing the entire React Query exports under `wagmi`, even if it doesn’t bring any conflicts, would feel weird for users. “What version does wagmi use internally?“ “Can they use the React Query exported by wagmi for their app too, or do they also need to install React Query seperately?” Which brings me to the next question :point_down: 

#### React Query as a peer dependency?

By requiring users to install react-query themselves, it would have the following advantages:

- Remove `13kb` from the final bundle if the consuming app also uses React Query.
- Make it explicit that wagmi is based on React Query. It would then split the API in two clear parts: wagmi, and React Query. For example, users would import `createPersistor()` or `QueryClient` from `react-query`, not `wagmi` (this would still be for advanced cases only, normal use cases should never have to import from `react-query`). If the API of React Query changes, it would not impact the `wagmi` API.
- Let users upgrade React Query to the latest version (as long as it stays in the range supported by wagmi).

But it would also bring some issues:

- `pnpm add wagmi ethers` is better than `pnpm add wagmi react-query ethers` :wink: 
- Consumer apps using React Query and a bundler without deduping would still have 13kb + 13kb.
- Same for apps using another version of React Query.
- We’ll have to support a range of React Query versions to support, which might introduces subtle issues.

So here too, my preference approach would be to not do this for now, and keep React Query as an internal dependency, while keeping it as a possibility for later (after the refactoring).

jnsdls:
thank you for this great proposal @jxom, we recently started integrating wagmi to handle wallet connections & signer / provider management into all of our projects @ thirdweb, we also are heavy users of react-query so this would be the perfect solution for us. (when I first saw wagmi the api interface reminded me a lot of react-query, to the point that I was actually surprised it was *not* built on top of it.)

needless to say we'd be excited to lend a hand helping to implement these changes, contributing fixes and improvements & long-term maintenance since we would likely have ended up forking wagmi eventually instead.

so yay for react-query 👍

### couple of loose (weekend) thoughts (coming from a perspective of our own use-case)
- as a heavy react-query user already the ideas of exposing an as-react-query-as-possible interface would be a huge selling point
- can we make the `queryClient` interoperable with an existing one? I don't necessarily want to define 2 query clients (one for app-code, one for wagmi), either explicitly or implicitly (aka: I would like to just pass the one I have already into the wagmi provider)
- regarding the peerdependency question, this is an interesting one and I think there are solid arguments for both approaches. since wagmi already has a peerdep of ethers I think it would be *ok* to add to that. (especially since it'll mean we can avoid the mismatching dependency problem for use-cases where the parent-app already makes use of react-query)

thanks again for the work on this so far, also was helpful to read the discussion above, we'll be watching this closely to see when and where and how we can help out!

jxom:
Response to @bpierre:

> ### On the npm modules & exports strategy

Yep, agree with this. I think a set of scoped packages will be nice. So essentially, we could have the following:
- `@wagmi/core`*: core unopinionated & framework/library agnostic functions
- `wagmi`*: React bindings for `@wagmi/core` w/ opinionated async fn management (react-query) 
- `@wagmi/swr`**: React bindings for `@wagmi/core` w/ opinionated async fn management (swr)
- `@wagmi/vue`: Vue bindings for `@wagmi/core`
- `@wagmi/svelte`: Svelte bindings for `@wagmi/core`

\* = build this now
** = also build this now if @shunkakinoki is free & available

> ### On the npm modules & exports strategy

Yeah, all this sounds good! Just to clarify, the [proposed API for storage persistance](https://github.com/jxom/wagmi-1/blob/ad88f59d4a3fed30e5f2b7a5d8c58cc0132dd62d/rfcs/001-deterministic-states-and-caching.md#storage-persistence) is already "wagmi simplified" (I don't expose `persistQueryClient` to the consumer), however, I wanted the consumer to have the ability to chose between whether to use the `localStorage` API or something completely different (sessionStorage, cookies, etc). But maybe we could have it more simplified, and by default it persists to local storage if `persist` is present, otherwise, you can override `persist` with a persistor?

```tsx
// By default, `persist` uses the local storage persistor

const App = () => (
  <Provider persist>
    <YourRoutes />
  </Provider>
)

///////////////////////////////////////////////////

// Override `persist` with a custom persistor (session storage)

import { Provider, createPersistor } from 'wagmi';

const sessionStoragePersistor = createPersistor({ storage: window.sessionStorage })

const App = () => (
  <Provider persist={sessionStoragePersistor}>
    <YourRoutes />
  </Provider>
)
```

> #### Full access to React Query?

Sounds good to me, an escape-hatch to allow a user to specify any React Query argument or return value sounds good. I guess we have to keep in mind that if we expose a `reactQueryResult` variable in the return values, this may introduce some performance issues (`useQuery` triggering re-renders when unused `reactQueryResult` return values are changed). Something to keep in mind.

> #### React Query as a peer dependency?

Yeah agree with all the trade-offs there. I am definitely for having React Query being an internal dependency – main reason is because it won't put such a reliance burden on this lib (if React Query upgrades to v4 - which we may not support straight away - then consumers will be waiting for us to migrate to RQ v4 before they can upgrade RQ to v4).

shunkakinoki:
@jxom 
Completely stand by the above! 
I'm down to help build the `wagmi/swr` lib - thank you for your thoughts here.

sammdec:
@jxom Is it worth spinning up a new branch that creates the basic folder, build structure for wagmi/core, wagmi/react, hopefully that should lower the setup complexity around contributing to this new setup

tmm:
Quick thoughts (more coming soon): `@wagmi/core` already exists! It's just called `wagmi-private`. We should rename (and deprecate) on npm. In addition, we can start abstracting logic from the hooks into the core package. Maybe into an `actions` directory that has a similar sub-directory setup as `hooks` (accounts, contracts, ens, etc.)?

o-az:
I've been following this discussion for the past few days, looks very exciting. There hasn't been any update in the past 5 days, just making sure if this is still happening?

I'm available and would be more than happy to help buidl! Cheers.

tmm:
The `@wagmi` scope on npm was taken so I renamed `wagmi-private` to `wagmi-core` in https://github.com/tmm/wagmi/pull/195. I also started extracting logic from the hooks in https://github.com/tmm/wagmi/pull/197 (h/t @bpierre for #140). Once all the hooks are extracted, it should make it really easy to start on adding react-query (or swr). If anyone wants to help migrate logic over to core "actions," go ahead!

This will also make it much easier for someone to add `wagmi-vue`, `wagmi-svelte`, etc.

jxom:
Going to merge this in for now. We currently have an [active development branch](https://github.com/tmm/wagmi/tree/next) for the new `@wagmi/core` package that will serve as a framework agnostic wagmi client (it is undocumented – however, feel free to [check out the example](https://github.com/tmm/wagmi/tree/next/examples/using-core)).

Right now, we are working on migrating the `wagmi` (react) package to [integrate with React Query](https://github.com/tmm/wagmi/pull/247) as per this proposal. 

If anyone wants to help with a [swr](https://swr.vercel.app/) adaptor, or even a svelte or vue client for wagmi, feel free to reach out! With the new foundation of `@wagmi/core`, building these framework specific clients should be trivial.

shunkakinoki:
Thank you and hats off to @jxom and @tmm for leading the effort into this.
You guys are just amazing...

Hope to be able to work on `wagmi/swr` in april, kind of caught off this month

PR #162, opened 2022-02-13
feat: ini ens resolveName

shunkakinoki:
Added a `resolveName` hook to resolve ens to address

ref: https://docs.ens.domains/dapp-developer-guide/resolving-names

### DIFF
--- a/packages/react/src/hooks/accounts/useBalance.ts
+++ b/packages/react/src/hooks/accounts/useBalance.ts
@@ -126,15 +126,9 @@ export const useBalance = ({
   /* eslint-disable react-hooks/exhaustive-deps */
   React.useEffect(() => {
     if (skip || !addressOrName) return
-
-    let didCancel = false
-    if (didCancel) return
     getBalance({ addressOrName, formatUnits, token })
-
-    return () => {
-      didCancel = true
-    }
-  }, [addressOrName, cacheBuster, skip, token])
+    return cancelQuery
+  }, [addressOrName, cacheBuster, cancelQuery, skip, token])
   /* eslint-enable react-hooks/exhaustive-deps */
 
   /* eslint-disable react-hooks/exhaustive-deps */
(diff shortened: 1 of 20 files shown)


bpierre:
Nice! :raised_hands: I didn’t realize `useResolver()` was not doing the actual resolving.

There are a few issues with this hook structure that got addressed by this PR: https://github.com/tmm/wagmi/pull/135

In summary:

- `didCancel` in the `useEffect()` doesn’t do anything (the condition will always be false since it happens right after the declaration) and can be safely removed.
- `resolveName` is not captured by the `useEffect()` dependencies, which means that the `useEffect()` wouldn’t react to a change of provider (returned by `useProvider()` and used by `resolveName()`).
- `setState()` is being called after an async operation ([here](https://github.com/tmm/wagmi/pull/162/files#diff-e5166895abc4256ecd8161f4ad8aebcfa8cc351015a90783a3b0b47e33057c39R36) and [here](https://github.com/tmm/wagmi/pull/162/files#diff-e5166895abc4256ecd8161f4ad8aebcfa8cc351015a90783a3b0b47e33057c39R40)) and will trigger warnings if the component has been rerendered / unmounted in the meantime. A `didCancel` flag (in a ref, outside of the `useCallback()`) would prevent this to happen.

packages/react/src/hooks/accounts/useBalance.ts
```diff
@@ -126,15 +126,9 @@ export const useBalance = ({
...
-
-    let didCancel = false
-    if (didCancel) return
     getBalance({ addressOrName, formatUnits, token })
-
-    return () => {
-      didCancel = true
-    }
-  }, [addressOrName, cacheBuster, skip, token])
+    return cancelQuery
```
tmm, line 130: Some carryover from #135

tmm:
thanks @shunkakinoki! was meaning to add this one.

@bpierre made the updates from your comment.

> resolveName is not captured by the useEffect() dependencies, which means that the useEffect() wouldn’t react to a change of provider (returned by useProvider() and used by resolveName()).

Used `cacheBuster` instead since it's further upsteam dep of `provider`. Also, update a few other hooks to use it. Another reason I didn't add actions to the `useEffect` deps was because I noticed a lot of unnecessary requests triggered when testing.

PR #197, opened 2022-02-20
refactor: create actions

tmm:
Starting to separate hook logic into core actions.

### Related
- https://github.com/tmm/wagmi/discussions/140
- #153

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -23,6 +23,7 @@ module.exports = {
     {
       ...config,
       displayName: 'core',
+      setupFilesAfterEnv: ['<rootDir>/packages/core/test/setup.ts'],
       testEnvironment: 'node',
       testRegex: 'packages/core/.*\\.test\\.ts$',
     },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -54,6 +54,7 @@
   "devDependencies": {
     "@walletconnect/ethereum-provider": "1.7.1",
     "ethers": "^5.5.1",
+    "wagmi-testing": "0.1.11",
     "walletlink": "^2.2.8"
   }
 }
--- a/packages/core/src/actions/accounts/balance.test.ts
+++ b/packages/core/src/actions/accounts/balance.test.ts
@@ -1,0 +1,62 @@
+import { contracts, wallets } from 'wagmi-testing'
+
+import { getProvider } from '../../../test'
+import { balanceAction } from './balance'
+
+const provider = getProvider()
+
+describe('balanceAction', () => {
+  it('fetches eth balance', async () => {
+    const balance = await balanceAction({
+      config: {
+        addressOrName: wallets.ethers1.address,
+      },
+      provider,
+    })
+    expect(balance).toMatchInlineSnapshot(`
+      {
+        "decimals": 18,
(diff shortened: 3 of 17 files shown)


tmm:
@bpierre curious what you think of this. anything you would change or do differently?

bpierre:
> @bpierre curious what you think of this. anything you would change or do differently?

Sorry for replying so late. It’s really cool, I like how we can now test these functions independently!

A few thoughts:

I was wondering if it could be a better idea to pick names like `getBalance()` rather than `balanceAction()` for the wagmi-core functions, and also have the config params directly rather than nested in `{ config: { … } }`, to make wagmi-core feel more like an independant library.

But thinking more about it, these functions won’t be easy to use anyway, since they don’t carry any context and it (chain, provider) needs to be setup, so it’s probably not going to be an issue. And we could always add a class in the future, to keep the context around:

```tsx
import { Wagmi } from 'wagmi-core'

// The Wagmi constructor would take the same options than the Provider
const wagmi = new Wagmi({ connectors: /* … */, autoConnect: true })

// Note 1: The Wagmi method names could follow the hooks, without the use- prefix.
// Note 2: The context is saved by Wagmi, no need to pass it to methods.
const balance = await wagmi.balance({ addressOrName: '0x…', token: '0x…' })
```

<hr />

About the erronous state issue described in https://github.com/tmm/wagmi/discussions/140, I don’t really see a way to fix this without breaking compatibility:

- If we export a `balanceAction()` wrapper (with the connection context) rather than [`getBalance()`](https://github.com/tmm/wagmi/blob/9a21e7e374f236114052e91454c5071d669c0982/packages/react/src/hooks/accounts/useBalance.ts#L53) to make it side effect free, it will break for consumers that were using it as a `refetch()` function.
- And if we make it a pure `refetch()` function, it will break for consumers that were using it as a data fetching function.

But with the React Query refactor, [`getBalance()` is now a `refetch()` function accepting a config](https://github.com/tmm/wagmi/blob/tmm/rq/packages/react/src/hooks/accounts/useBalance.ts#L85). From what I read (I haven’t tried to run this branch yet), we can still override the config passed to the hook frome here, but it doesn’t return anymore. Also, having `watch: true` now respects the latest config passed (rather than reverting to the hook params on the next update).

@tmm @jxom I was wondering if we shouldn’t simply remove `getBalance()` from the returned object in the React Query refactor, since we are breaking the compatibility anyway?

Migrating to the new version of wagmi would require consumers to either:

- Call `refetch()` if they want to force-refresh.
- Update the hook params if they want to update their hook configuration.
- Use `balanceAction()` from wagmi-core if they were using the promise returned by `getBalance()`.

jxom:
Yo @bpierre,

Yeah, I added getter fns (e.g. `getBalance()`) in addition to `refetch` just to try and keep the ability to imperatively invoke the hook functions, however thinking about it, it definitely would break compatibility with the current API, and might be a bit confusing since the config you pass to the getter fns also modify the cache. I reckon it might be a lot clearer to just remove this (as you suggested), and for the consumer to keep track of the hook parameters. I think this approach is more declarative, and would allow the consumer to think about side-effects more directly.

tmm:
Good points! Let's remove this from the new version.

PR #202, opened 2022-02-21
refactor: create blockNumberAction

margalit:
Move blockNumberAction into dedicated function.

Related
https://github.com/tmm/wagmi/discussions/140
https://github.com/tmm/wagmi/pull/153


packages/core/src/actions/network-status/blockNumber.ts
```diff
@@ -0,0 +1,12 @@
+import { WithProvider } from '../../types'
+
+export type BlockNumberActionArgs = WithProvider<unknown>
+
+export type BlockNumberActionResult = number
```
margalit, line 5: @tmm Do you think we should wrap this `number` in a `BlockNumber` type?

packages/core/src/actions/network-status/blockNumber.ts
```diff
@@ -0,0 +1,12 @@
+import { WithProvider } from '../../types'
+
+export type BlockNumberActionArgs = WithProvider<unknown>
```
margalit, line 3: `unknown` here works but is a bit gross. Open to alternatives since ts complains on `{}`.

margalit (commented):
@tmm I've had a crack at refactoring the easiest action out. Let me know if this is what you had in mind.

tmm:
Really appreciate this work @margalit! Going to roll into #203

PR #215, opened 2022-02-25
feat:support EIP-712 sign typed data

SunskyXH:
### Motivation
Sign plain text message may confuse user, especially sign some text looks like OpenSea's list message. So more devs are going to use typed data in favor of plain text(eg. OpenSea's new list contract).
### Changes
- support EIP-712: Ethereum typed structured data hashing and signing
- use `signer._signTypedData` to implementation this feature
- 🔗 ethers.js signTypedData doc link: https://docs.ethers.io/v5/api/signer/#Signer-signTypedData

Your ENS/address: sunskyxh.eth

### DIFF
--- a/.changeset/brave-mugs-deny.md
+++ b/.changeset/brave-mugs-deny.md
@@ -1,0 +1,7 @@
+---
+'docs': minor
+'wagmi': minor
+'wagmi-testing': minor
+---
+
+Support EIP-712 sign typed data.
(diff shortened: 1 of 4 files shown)


kesar:
Could this get reviewed / merged? 🙏🏻 

We are using wagmi and implemented a very similar code for signing, it would be great if this would be part of the lib 👍🏻

tmm:
Closed and opened #250 to add updates

PR #264, opened 2022-03-20
[bug] useAccount "loading" should reflect ENS lookup from first render

davisshaver:
Your ENS/address: davisshaver.eth

See https://github.com/tmm/wagmi/issues/263

### DIFF
--- a/packages/react/src/hooks/ens/useEnsLookup.test.ts
+++ b/packages/react/src/hooks/ens/useEnsLookup.test.ts
@@ -86,6 +86,17 @@ describe('useEnsLookup', () => {
     `)
   })
 
+  it('not loading from first render', async () => {
+    const { result } = renderHook(() => useEnsLookup({ skip: true }))
+    expect(result.current[0]).toMatchInlineSnapshot(`
+      {
+        "data": undefined,
+        "error": undefined,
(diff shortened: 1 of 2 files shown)


davisshaver:
Here's a CodeSandbox link showing the extra cycles of `loading` state with ENS lookup: https://codesandbox.io/s/dazzling-wright-4405hb?file=/src/App.tsx:0-32

In terms of use case for this change, I have an API request that is triggered by connection event. I would like to include ENS name with this API request. At present I have needed to introduce additional state variables (`hasLoadedFirstTime`, `isLoadingSecondTime`, `hasLoadedSecondTime`) to ensure that ENS lookup has been completed before API request is made.

tmm:
Thanks for spending time doing this.

In the next wagmi version, `fetchEns` is [removed from `useAccount`](https://next.wagmi.sh/docs/migrating-to-030#useaccount=), in favor of keeping it as lightweight as possible. [`useEnsName`](https://next.wagmi.sh/docs/hooks/useEnsName) and [`useEnsAvatar`](https://next.wagmi.sh/docs/hooks/useEnsAvatar) should be used instead to take advantage of caching, request deduplication, etc.

PR #279, opened 2022-03-24
feat: MetaMask connector

adbario:
Fixes https://github.com/tmm/wagmi/issues/260

Adds a MetaMask specific connector by extending the existing `InjectedConnector`.

### Why do we need this?

When the user has both MetaMask and Coinbase Wallet browser extensions installed, injected connector fires them both. This new provider can be used to target only MetaMask.

### How to use it?

All you need to do is change the `InjectedConnector` to `MetaMaskConnector`:

```javascript
import { InjectedConnector } from "wagmi/connectors/injected";

const connectors = () => {
  return [
    new InjectedConnector({
      chains,
      options: { shimDisconnect: true },
    }),
  ];
};

// Becomes to...

import { MetaMaskConnector } from "wagmi/connectors/metaMask";

const connectors = () => {
  return [
    new MetaMaskConnector({
      chains,
      options: { shimDisconnect: true },
    }),
  ];
};
```


SonnyCastro:
This is great! It's funny because I was working on a PR for this connector as well. I did something similar within the getProvider() function. It's cool to see your files changed and compare and see where I missed things.

Were you able to test this and if so how? 

Hope this gets merged would be cool!

Thank you @adbario

tmm:
@adbario thanks for putting this together. we probably also want to override `ready` and check for `isMetaMask`.

adbario:
> @adbario thanks for putting this together. we probably also want to override `ready` and check for `isMetaMask`.

Happy to look into this soon 🚀

adbario:
@tmm `ready` property is implemented now and merged `main` to solve the conflicts as type declarations were moved to another file on some other PR.

adbario:
Just a quick note here regarding Brave Wallet; they do extremely good work trying to make sure it looks like a MetaMask inside the `ethereum` object, but have already found a few differences that can be used to detect this properly.

Also, the Brave Wallet detection in https://github.com/tmm/wagmi/pull/320 doesn't work currently at all, as that property is not in use and they set `isMetaMask` to `true`.

Will come back soon with some options.

adbario:
MetaMask detection is updated now and works well now to distinguish it from the new Brave Wallet.

Brave Wallet is trying to mimic MetaMask extremely well, they even set the `isMetaMask` property to `true`.

I used the following methods to further check whether the provider really is MetaMask:
- `ethereum._events.connect` event only exists on MetaMask
- `ethereum._state` object only exists on MetaMask

It's good to also understand that the methods these wallets implement might change at any time and Brave will possibly try to make it even more difficult to differentiate them.

tmm:
Closing in favor of #391

PR #282, opened 2022-03-25
Add provider dynamic import example

albertputrapurnama:
Your ENS/address: apurn.eth

I moved `Provider` away from `_app.tsx` to demonstrate how to dynamically load `wagmi`'s Provider.

This serves as an answer for https://github.com/tmm/wagmi/discussions/32.

## Improvements

Before:
Screen Shot 2022-03-25 at 12 39 26 AM
First Load JS >300kb across all pages

After:
Screen Shot 2022-03-25 at 12 44 17 AM
First load JS on heaviest page (the one where we import `wagmi` without any lazy loading) = 200kb
Shared JS (~70kb) because we dynamically loaded `wagmi` AND `ethers`

### DIFF
--- a/examples/next/src/components/Provider.tsx
+++ b/examples/next/src/components/Provider.tsx
@@ -1,0 +1,78 @@
+// Imports
+import React from 'react'
+import { providers } from 'ethers'
+import {
+  Connector,
+  Provider as WagmiProvider,
+  chain,
+  defaultChains,
+} from 'wagmi'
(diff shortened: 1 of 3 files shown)


tmm:
Thanks for doing this! Let's keep the example in GitHub Discussions for now.

seeARMS:
@LaCocoRoco, it seems your solution doesn't actually cause the WagmiProvider to be dynamically imported. Perhaps it's due to the reasons mentioned in #32.

PR #286, opened 2022-03-28
Update useWaitForTransaction.mdx

Hussainzz:
Your ENS/address:


Hussainzz:
i was figuring out the way to wait for transaction manually. as there was no example in the wagmi doc i looked into the repo code and also in the etherjs doc, that we can pass wait/hash while calling the wait() manually. so thought of adding this snippet so new members can understand it easily

tmm:
Thanks for this! I'll include something similar in the new docs launching soon.

PR #310, opened 2022-04-03
update provider example with Alchemy provider

thatguyintech:
Your ENS/address: thatguyintech.eth


tmm:
@thatguyintech thanks! Let's add these type of changes to the [`next`](https://github.com/tmm/wagmi/tree/next) branch.

PR #312, opened 2022-04-04
Remove disconnect from useAccount docs and refer to useDisconnect hook

albertputrapurnama:
Your ENS/address: apurn.eth


tmm:
Great minds think alike. I updated this at the same time as you :) https://github.com/tmm/wagmi/pull/313

Will add the note to the migration guide.

PR #314, opened 2022-04-04
add more example rpc url chains

thatguyintech:
Your ENS/address: thatguyintech.eth

### DIFF
--- a/docs/pages/docs/constants/rpcs.en-US.mdx
+++ b/docs/pages/docs/constants/rpcs.en-US.mdx
@@ -2,13 +2,15 @@
 
 ## alchemyRpcUrls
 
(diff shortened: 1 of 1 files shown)


docs/pages/docs/constants/rpcs.en-US.mdx
```diff
@@ -2,13 +2,15 @@
 
 ## alchemyRpcUrls
 
-RPC URLs supported by [Alchemy](https://www.alchemy.com).
+RPC URLs supported by [Alchemy](https://alchemy.com/?a=wagmi).
```
thatguyintech, line 5: hope this is ok -- useful for us to track how many people are visiting from wagmi docs :)

PR #319, opened 2022-04-05
fix: fix exposing global types

cawabunga:
## Issue
Let's say I need to override `ethereum` object in window. I need to create `global.d.ts` and put something like:

```ts
export declare global {
  interface Window {
    ethereum: any,
  }
}
```

Currently it is impossible, because types are not declared properly and they override my overriding. 
I created repository with an example https://github.com/cawabunga/wagmi-window-error-example

## References
- Typescript Doc https://www.typescriptlang.org/docs/handbook/declaration-merging.html#global-augmentation
- Example https://github.com/cawabunga/wagmi-window-error-example

## Implementation
The PR doesn't break backward capability. All types are exposed to global scope (I think by mistake), so PR keeps this behaviour as well. The PR just allows to override exposed interfaces.

### DIFF
--- a/packages/core/src/types/declarations.d.ts
+++ b/packages/core/src/types/declarations.d.ts
@@ -1,52 +1,54 @@
-type AddEthereumChainParameter = {
-  chainId: string // A 0x-prefixed hexadecimal string
-  chainName: string
-  nativeCurrency?: {
-    name: string
-    symbol: string // 2-6 characters long
-    decimals: 18
+export declare global {
+  type AddEthereumChainParameter = {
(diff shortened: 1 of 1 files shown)


tmm:
@cawabunga thanks for doing this!

can you move `declarations.d.ts`'s content to `packages/core/src/types/index.ts` instead and drop the `export`? Will help simplify things.

tmm:
Fixed in https://github.com/tmm/wagmi/commit/468035ad8e70ab8761667be22b79a8ad7210d0a4

PR #324, opened 2022-04-06
docs: add Brave Wallet to list of connectors

josheleonard:
Added Brave Wallet to the features/wallet connectors section of the repo's Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ React Hooks library for Ethereum, built on [ethers.js](https://github.com/ethers
 ## Features
 
 - 🚀 20+ hooks for working with wallets, ENS, contracts, transactions, signing, etc.
-- 💼 Built-in wallet connectors for MetaMask, WalletConnect, and Coinbase Wallet
+- 💼 Built-in wallet connectors for MetaMask, WalletConnect, Coinbase Wallet, and Brave Wallet
 - 🌀 Auto-refresh data on wallet and network changes
 - 🦄 TypeScript ready
 - 💨 Zero-dependencies (besides ethers.js peer dependency)


tmm:
Thanks for checking this out! Let's keep this list to MetaMask, WalletConnect, and Coinbase Wallet for now. These are the most recognizable.

Also, MM (and Brave Wallet) aren't really distinct connectors in wagmi*. They are all part of the injected connector. There are lots of other wallets (Frame, Tally, etc.) captured by the injected connector too.

---

\* Might [add a built-in MM connector](https://github.com/tmm/wagmi/pull/279) soon. Could also do this for others where it makes sense.

bbondy:
Sad to see that this list is being limited to incumbents only. I'd prefer to see the other more open source options like Brave Wallet and Tally included in the list.

tmm:
We can and should list other options at some point. I'm working on the next version of wagmi right now and don't have time to make this decision.

@bbondy What criteria would you use to determine what gets added to the list?

bbondy:
Thanks for responding @tmm. I'd probably say to just keep it in sync with this list:
https://github.com/tmm/wagmi/blob/main/packages/core/src/types/index.ts#L60

I'm sure if you consider re-opening with a review- comment, then @josheleonard would expand the list to also include Tally and Frame.

josheleonard:
Updated my forked branch with the other missing wallets 
https://github.com/josheleonard/wagmi/blob/docs-include-brave-wallet-in-features-list/README.md

tmm:
I updated the [README](https://github.com/tmm/wagmi/tree/next#features=) and [docs](https://next.wagmi.sh/#features=) for the next version of wagmi to reflect the following:

> Built-in wallet connectors for Injected (e.g. MetaMask), WalletConnect, and Coinbase Wallet

I also added the following note to the [`InjectedConnector`](https://next.wagmi.sh/docs/connectors/injected#usage=) docs:

> Injected wallets can set up custom names mappings in wagmi. You can see the full list and add to it here. By default, "Injected" is the name for unmapped wallets.

By default, all injected wallets (should) work without any custom code. It's not scalable to list out every custom injected name mapping. WalletConnect is very similar in the way that it supports many wallets. We don't list out each wallet in that case either.

As much as I think MM needs to improve across many fronts, it has 30M+ users and developers look for it when deciding what libraries to use. I appreciate the thoughts and we can re-evaluate this decision in the future.

PR #328, opened 2022-04-09
feat: pass custom chain id

tmm:
Allows for passing custom chain id to hooks and actions to override root ethers provider.

For example, when you set up a chain-aware ethers provider when creating a client:

```ts
import { providers } from 'ethers'
import { createClient } from 'wagmi'

const client = createClient({
  provider(config) {
    return new providers.AlchemyProvider(config.chainId, 'Your Alchemy ID')
  },
})
```

The `chainId` of that provider will change when the user switches chains. This isn't always desired. Maybe you only want to fetch an account's balance for a specific network. For example, Ethereum Mainnet (`chainId` `1`):

```ts
import { useBalance } from 'wagmi'

function App() {
  const balance = useBalance({
    addressOrName: '0xA0Cf798816D4b9b9866b5330EEa46a18382f251e',
    chainId: 1,
  })
}
```

This PR adds `chainId` as a configuration option to React hooks and  Vanilla JS actions.


packages/react/src/hooks/utils/useChainId.ts
```diff
@@ -1,6 +1,10 @@
 import { useProvider } from '../providers'
 
-export function useChainId() {
-  const provider = useProvider()
-  return provider?.network?.chainId
+export type UseChainIdArgs = {
+  chainId?: number
+}
+
+export function useChainId({ chainId }: UseChainIdArgs = {}) {
```
jxom, line 7: Can the returned chain id be different to the passed `chainId`?

tmm: It can, but only if the provider does not support the chain. [Added a test.](https://github.com/tmm/wagmi/pull/330/files)

PR #336, opened 2022-04-13
feat: add gnosis connector

AmagiDDmxh:
Your ENS/address: amagii.eth

Add gnosis connector, it might help with people who wanna use gnosis connector.

References:
- https://github.com/gnosis/safe-apps-sdk/blob/master/packages/safe-apps-web3modal/src/modal.ts
- https://github.com/gnosis/safe-apps-sdk/blob/master/packages/safe-apps-web3-react/src/connector.ts

### DIFF
--- a/docs/pages/docs/connectors/gnosis.mdx
+++ b/docs/pages/docs/connectors/gnosis.mdx
@@ -1,0 +1,44 @@
+# GnosisConnector (Gnosis, etc.)
+
+The **GnosisConnector** supports Gnosis Dapp connection, that runs in the Gnosis context.
+
+```tsx
+import { GnosisConnector } from 'wagmi/connectors/gnosis'
+```
+
+## Usage
+
+```tsx
+import { GnosisConnector } from 'wagmi/connectors/gnosis'
+
+const connector = new GnosisConnector()
+```
+
+## Configuration
+
+### chains (optional)
+
+Chains supported by app. Defaults to `defaultChains`.
+
+```tsx highlight='5'
+import { defaultChains, defaultL2Chains } from 'wagmi'
+import { GnosisConnector } from 'wagmi/connectors/gnosis'
+
+const connector = new GnosisConnector({
+  chains: [...defaultChains, ...defaultL2Chains],
+})
+```
+
+### options
+
+Options for configuring the connector.
+
+```tsx highlight='4-6'
+import { GnosisConnector } from 'wagmi/connectors/gnosis'
+
+const connector = new GnosisConnector({
+  options: {
+    shimDisconnect: true,
+  },
+})
+```
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -49,6 +49,8 @@
   },
   "dependencies": {
     "@ethersproject/providers": "^5.5.1",
+    "@gnosis.pm/safe-apps-provider": "^0.11.0",
+    "@gnosis.pm/safe-apps-sdk": "^7.3.0",
     "eventemitter3": "^4.0.7"
   },
   "devDependencies": {
--- a/packages/core/src/connectors/gnosis.ts
+++ b/packages/core/src/connectors/gnosis.ts
@@ -13,9 +13,13 @@ import { ConnectorNotFoundError } from '../errors'
 
 const __IS_SERVER__ = typeof window === 'undefined'
 
+type SafeConnectOptions = SafeOpts & {
+  doNotAutoConnect?: boolean
+}
+
 export class GnosisConnector extends Connector<
   SafeAppProvider,
-  SafeOpts | undefined
+  SafeConnectOptions | undefined
 > {
   readonly id = 'gnosis'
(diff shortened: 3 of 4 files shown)


blokboy:
hey @AmagiDDmxh does this require you to open the app from within gnosis safe still?

AmagiDDmxh:
> hey @AmagiDDmxh does this require you to open the app from within gnosis safe still?

no, u can use it within a no safe app env

blokboy:
i think it would be a nice touch. would you like for me to make a PR against your branch?

AmagiDDmxh:
sure it'd be super nice

packages/core/src/connectors/gnosis.ts
```diff
@@ -19,19 +19,26 @@ export class GnosisConnector extends Connector<
...
-      this.#sdk = new SafeAppsSDK(config.options)
-      // Auto connect on safe environment
-      this.connect()
-    }
+    this.#isSafeApp()
+      .then((isSafeApp) => {
+        if (!__IS_SERVER__ && isSafeApp) {
+          this.ready = true
+          this.#sdk = new SafeAppsSDK(config.options)
+          // Auto connect on safe environment
```
blokboy, line 35: Do you think it would be okay if I just add a check here within options for the provided flag `autoConnect` and if the user has flagged for it then execute else just return the sdk?

blokboy: or perhaps `doNotAutoConnect` would be a better naming convention and don't run `connect` when the flag is provided

AmagiDDmxh: I think `doNotAutoConnect` is better

AmagiDDmxh: 😉 add `doNotAutoConnect`, plz check

blokboy (approved):

packages/core/src/connectors/gnosis.ts
blokboy: Looks great :)

packages/core/package.json
```diff
@@ -49,6 +49,8 @@
   },
   "dependencies": {
     "@ethersproject/providers": "^5.5.1",
+    "@gnosis.pm/safe-apps-provider": "^0.11.0",
+    "@gnosis.pm/safe-apps-sdk": "^7.3.0",
```
kesar, line 53: I think new providers should be independent packages so we dont have third party dependencies for the core of this lib. 

If we allow gnosis, we should allow 100 more

AmagiDDmxh: sure. the possible solution could be expose needed connector api to allow third party to integrate they own, but this is quite lacking right now.

or set these pkgs as peer deps.

bigint:
Love to see this 🤗

packages/core/package.json
janaSunrise: I was going through this, and I might have a workaround for this. Users won't probably even need to many connectors, but since this is a monorepo, what if we published other connectors under `@wagmi/connector-<service>`, such as `@wagmi/connector-gnosis` or something as such.

That would be totally opt-in and you install what you need.

jcheese1:
would love to see this as well <3

tmm:
Thanks for working on this @AmagiDDmxh! What's the best way to go about testing this? (e.g. Do I need to download an app, install a browser extension, create a Gnosis Safe, etc.)

jcheese1:
@tmm To test, create a gnosis safe, and create a demo webapp with a `manifest.json` file in the public folder like so:
```json
  "short_name": "Bridgeworld",
  "name": "Bridgeworld is a game of strategic commerce, trade and domination which sits at the center of the Treasure metaverse",
  "description": "Bridgeworld by TreasureDAO",
  "iconPath": "/img/logo.svg",
  "icons": [
    {
      "src": "favicon.ico",
      "sizes": "64x64 32x32 24x24 16x16",
      "type": "image/x-icon"
    }
  ],
  "start_url": ".",
  "display": "standalone",
  "theme_color": "#bf0026",
  "background_color": "#bf0026"
}
```

And then in gnosis, go to Apps -> Add Custom App and link your demo app there

bigint:
Excited for this, we want it badly 🙏🏻

docs/pages/docs/connectors/gnosis.mdx
```diff
@@ -0,0 +1,44 @@
+# GnosisConnector (Gnosis, etc.)
```
mikhailxyz, line 1: Could this be changed to `Safe`? The Safe project has spin out of gnosis recently. Also, there are other Gnosis' projects such as Gnosis Chain, etc

mikhailxyz:
@kesar fair enough, we'll discuss that

tmm:
@kesar good point! @mikhailxyz i think this would be ideal if you could swing it. happy to discuss more as well about trade-offs, etc.

PR #377, opened 2022-04-28
refactor: serialize

tmm:
## Description

More stable, faster `serialize`

## Additional Information

Your ENS/address: awkweb.eth


packages/react/src/index.ts
```diff
@@ -39,6 +39,8 @@ export {
   useWebSocketProvider,
 } from './hooks'
 
+export { deserialize, serialize } from './utils'
```
tmm, line 42: expose for folks that want to create their own persister

PR #387, opened 2022-04-30
Feature/ankr rpcs

codingwithmanny:
## Description

Added Ankr rpc provider support and documentation alphabetically.

## Additional Information

Your ENS/address:

codingwithmanny.eth


codingwithmanny:
@tmm Did you want to wait for the other testnets to be added to ethers to revise this to include the testnets or fair to proceed with these ones and add the additional testnets in a separate PR?

Ethersjs commit with Ankr Providers
https://github.com/ethers-io/ethers.js/commit/d9f45b3b9db92c72f9c606bab8315d0eb02fec70

tmm:
@codingwithmanny can wait if cool. @jxom is looking into a streamlined provider and chains API too that might inform some of this.

codingwithmanny:
@tmm would this replace both infura and alchemy providers and potentially support quicknode all in one fell swoop :) ?

If you think it's going to take a while, would be ok if we merged it in the meantime, I have a use-case for it where I want to work with Ankr's premium API which apparently has a load balancer that uses all providers, if one fails, it goes to the next.

codingwithmanny:
Good news! Testnets were merged in ethers.js 5.6.5
https://github.com/ethers-io/ethers.js/issues/2949#issuecomment-1115749557

codingwithmanny:
@tmm this the direction you were talking about? If so 🔥 
https://github.com/tmm/wagmi/pull/408

jxom:
> @tmm this the direction you were talking about? If so 🔥 #408

you are correct!

PR #391, opened 2022-05-01
feat: MetaMaskConnector

tmm:
## Description

- Adds a MetaMask-specific connector (s/o @adbario for kicking off #279)
- Adds optional `name` configuration option to `InjectedConnector`
  ```ts
  new InjectedConnector({ options: { name: 'Injected' }})
  ```
-  Decouple `CoinbaseWalletConnector` `ready` from `InjectedConnector`

## Additional Information

- [x] Update docs
- [x] Update examples

Your ENS/address: awkweb.eth

### DIFF
--- a/packages/core/src/connectors/coinbaseWallet.ts
+++ b/packages/core/src/connectors/coinbaseWallet.ts
@@ -67,7 +67,6 @@ export class CoinbaseWalletConnector extends Connector<
         ),
       }
     } catch (error) {
-      console.log(error)
       if (
         /(user closed modal|accounts received is empty)/i.test(
           (<ProviderRpcError>error).message,
--- a/packages/core/src/connectors/coinbaseWallet.ts
+++ b/packages/core/src/connectors/coinbaseWallet.ts
@@ -37,8 +37,7 @@ export class CoinbaseWalletConnector extends Connector<
 > {
   readonly id = 'coinbaseWallet'
   readonly name = 'Coinbase Wallet'
-  readonly ready =
-    typeof window !== 'undefined' && !window.ethereum?.isCoinbaseWallet
+  readonly ready = true
 
   #client?: CoinbaseWalletSDK
   #provider?: CoinbaseWalletProvider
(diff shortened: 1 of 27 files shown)


packages/core/src/connectors/coinbaseWallet.ts
```diff
@@ -37,8 +37,7 @@ export class CoinbaseWalletConnector extends Connector<
 > {
   readonly id = 'coinbaseWallet'
   readonly name = 'Coinbase Wallet'
-  readonly ready =
-    typeof window !== 'undefined' && !window.ethereum?.isCoinbaseWallet
+  readonly ready = true
```
tmm, line 40: `CoinbaseWalletConnector` was coupled to `InjectedConnector` so that only one would be active at a time. `CoinbaseWalletConnector` should always work either via QR code modal and/or the Coinbase Wallet browser extension so it can always be set to `true`.

packages/core/src/connectors/coinbaseWallet.ts
```diff
@@ -68,7 +67,12 @@ export class CoinbaseWalletConnector extends Connector<
         ),
       }
     } catch (error) {
-      if (/user closed modal/i.test((<ProviderRpcError>error).message))
+      console.log(error)
```
markdalgleish, line 70: Just double checking - is this console log meant to be here?

packages/core/src/connectors/metaMask.ts
```diff
@@ -0,0 +1,49 @@
...
+  async getProvider() {
+    if (typeof window !== 'undefined')
+      this.#provider = this.#findProvider(window.ethereum)
+    return this.#provider
+  }
+
+  #getReady(ethereum?: Ethereum) {
+    const isMetaMask = !!ethereum?.isMetaMask
+    // Brave Wallet tries to make itself look like MetaMask
+    // Agressively try matching against these properties unique to MetaMask
```
markdalgleish, line 39: 🤯

packages/core/src/connectors/coinbaseWallet.ts
tmm: nope! good catch

packages/core/src/connectors/metaMask.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (isMetaMask && !!ethereum._events?.connect && !!ethereum._state)
```
markdalgleish, line 40: This is still needed even though `ethereum.isBraveWallet` is present? (Based on the following lines)

tmm: Nope, should be good without it. Removed!

jxom (approved):

PR #392, opened 2022-05-01
fix: memoize `createClient` in Remix example

jxom:
## Description

Memoizing the wagmi client creation so we don't clear out the wagmi state on every render.


examples/remix/app/root.tsx
```diff
@@ -35,45 +36,51 @@ const isChainSupported = (chainId?: number) =>
...
+                },
+              },
+            }),
+          ]
+        },
+        provider({ chainId }) {
+          return new providers.AlchemyProvider(
+            isChainSupported(chainId) ? chainId : defaultChain.id,
+            alchemyId,
+          )
```
aramalipoor, line 73: Do you think it's a good idea to fallback to Web3 provider? Otherwise if Infura/Alchemy does not support the chain it'll just fail with an exception.

Just as PoC (pretty bad error handling 😄 ). This can help with smoothly supporting custom chains (e.g. Moonbeam):
```ts
try {
  return new providers.InfuraProvider(config.chainId, infuraId);
} catch (e) {
  try {
    return providers.getDefaultProvider(config.chainId);
  } catch (e) {
    try {
      return new providers.Web3Provider(
        window.ethereum as any,
        Number(config.chainId)
      );
    } catch (e) {
      return providers.getDefaultProvider();
    }
  }
}
```

jxom: Yeah, it is a good idea! We are actually planning to add a first class RPC Provider API very shortly (which will have fallback support for: if a provider goes down, or if a chain is unsupported by a provider).

We will update the examples accordingly once we add this.

PR #404, opened 2022-05-03
Store/expose caught errors before rethrowing

holic:
## Description

Store/expose caught errors before rethrowing custom wagmi-specific errors.

Motivation: I wanted to dive deeper into why Rainbow was throwing a "switch chain error" when calling switch chain and realized I didn't have access to the originally thrown error, which might have more hints as to what's going on.

I initially extended ~all of the custom wagmi errors, but found only a subset were used in try/catch blocks, so ended up only wrapping those. There are a couple places in mocks where we pass an `undefined` argument, which is a little gross, but the argument type safety seemed nice to have in all other spots.

## Additional Information

Your ENS/address:
frolic.eth

### DIFF
--- a/packages/core/src/actions/accounts/signMessage.ts
+++ b/packages/core/src/actions/accounts/signMessage.ts
@@ -21,7 +21,7 @@ export async function signMessage(
   } catch (error_) {
     let error: Error = <Error>error_
     if ((<ProviderRpcError>error_).code === 4001)
-      error = new UserRejectedRequestError()
+      error = new UserRejectedRequestError(error_)
     throw error
   }
 }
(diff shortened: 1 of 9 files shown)


holic:
For my specific use case, this was an issue switching to Rinkeby on Rainbow for iOS. The underlying error is `Chain currently not supported`. I wonder if wagmi should detect/throw `SwitchChainNotSupportedError` instead?

tmm:
Great idea! (Amazing this is #404.) Traveling and will take a look soon.

holic:
🙏  much nicer, thanks for making this pretty

PR #408, opened 2022-05-04
feat: `configureChains` & API Providers

jxom:
## Description

This PR introduces the `configureChains` function and the concept of "API Providers".

The `configureChains` function allows you to configure your chains with API providers such as: Alchemy, Infura, or something else. This means you don't need to worry about defining RPC URLs and chain configuration in your [connector](https://wagmi-mu88njirb-zoo.vercel.app/docs/connectors/walletconnect#options) or [provider](https://wagmi-mu88njirb-zoo.vercel.app/docs/client#provider-optional). This is managed internally by wagmi.

This PR also deprecates the use of providing `connectors` with a function to `createClient`. Now, the `connectors` option on `createClient` should be a static array of connectors. This fixes the issue where [switching chains will make the active connector stale](https://github.com/tmm/wagmi/issues/365).

Check out the docs for `configureChains` here: https://wagmi-lcp3docvd-zoo.vercel.app/docs/api-providers/configuring-chains


packages/core/src/chains/configureChains.test.ts
```diff
@@ -0,0 +1,336 @@
...
+          You may need to add \`fallbackProvider\` to \`configureChains\` so that the chain can fall back to the public RPC URL."
+        `)
+      })
+    })
+
+    describe('jsonRpc', () => {
+      beforeAll(() => {
+        ;[chain.mainnet, chain.polygon, chain.optimism, chain.arbitrum].forEach(
+          (chain) => {
+            nock(`https://${chain.id}.example.com`).get('/').reply(200, {})
```
tmm, line 207: can we either use `msw` here or convert `useEnsAvatar` and `fetchEnsAvatar` to `nock`?

packages/core/src/client.ts
```diff
@@ -278,20 +279,16 @@ export class Client<
...
-    if (subscribeConnectors || subscribeProvider || subscribeWebSocketProvider)
+    if (subscribeProvider || subscribeWebSocketProvider)
       this.store.subscribe(
         ({ data }) => data?.chain?.id,
         (chainId) => {
           this.setState((x) => ({
             ...x,
-            connectors: subscribeConnectors
-              ? connectors({ chainId })
-              : x.connectors,
```
tmm, line 294: 🔥🔥🔥

packages/core/src/connectors/walletConnect.ts
```diff
@@ -89,10 +89,20 @@ export class WalletConnectConnector extends Connector<
...
+            }),
+            {},
+          )
+        : undefined
+
       const WalletConnectProvider = (
         await import('@walletconnect/ethereum-provider')
       ).default
-      this.#provider = new WalletConnectProvider(this.options)
+      this.#provider = new WalletConnectProvider({ ...this.options, rpc })
```
tmm, line 105: How would this work if someone set up their connector like this?

```ts
new WalletConnectConnector({
  options: {
    rpc: { 1: 'https://eth-mainnet.mycoolrpc.com/v2' }
  },
})
```

Seems like the defined `rpc` would get overridden because `infuraId` isn't defined.

packages/core/src/apiProviders/jsonRpc.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (rpcUrl === '') return null
+    return {
+      chain: {
+        ...chain,
+        rpcUrls: {
+          ...chain.rpcUrls,
+          default: rpcUrl,
+        },
+      },
+      provider: () => new providers.StaticJsonRpcProvider(rpcUrl, chain.id),
```
tmm, line 24: I'm all for using `StaticJsonRpcProvider` (cuts down on unnecessary calls), but it might be confusing that `jsonRpcProvider` doesn't use [`JsonRpcProvider`](https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider). Could add a `static` param that is defaulted to `true` or something.

tmm (commented):
Looking solid! LMK if you want more detail about any of the comments.

With this change, I wonder if we should make `createClient({ provider })` required and drop the `getDefaultProvider` default? Seems like most pass their own provider to `createClient` anyway (this new API will make it even easier!) and the default `getDefaultProvider` just uses the default ethers provider keys. `getDefaultProvider` also has a runtime cost of importing a lot of providers most will never use.

packages/core/src/chains/configureChains.ts
```diff
@@ -0,0 +1,70 @@
...
+        ].join('\n'),
+      )
+    }
+  }
+
+  return {
+    chains,
+    provider: ({ chainId }: { chainId?: number }) => {
+      return providers[
+        chainId && chains.some((x) => x.id === chainId)
```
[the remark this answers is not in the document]
jxom: Yep, pretty sure it does.

packages/core/src/apiProviders/jsonRpc.ts
jxom: maybe i'll rename this to `staticJsonRpcProvider`, might make types easier.

packages/core/src/actions/providers/watchWebSocketProvider.ts
```diff
@@ -8,11 +8,11 @@ import {
 } from './getWebSocketProvider'
 
 export type WatchWebSocketProviderCallback<
-  TWebSocketProvider extends providers.WebSocketProvider = providers.WebSocketProvider,
+  TWebSocketProvider extends providers.BaseProvider = providers.WebSocketProvider,
 > = (webSocketProvider: GetWebSocketProviderResult<TWebSocketProvider>) => void
 
 export function watchWebSocketProvider<
-  TWebSocketProvider extends providers.WebSocketProvider = providers.WebSocketProvider,
+  TWebSocketProvider extends providers.BaseProvider = providers.WebSocketProvider,
```
tmm, line 15: Why the switch from `WebSocketProvider` to `BaseProvider`?

packages/core/src/types/index.ts
```diff
@@ -7,7 +7,8 @@ import {
 
 export type Chain = {
   id: number
-  name: AddEthereumChainParameter['chainName']
+  name: string
+  displayName: AddEthereumChainParameter['chainName']
```
tmm, line 11: I'm good with this, since it [aligns with ethers](https://github.com/ethers-io/ethers.js/blob/948f77050dae884fe88932fd88af75560aac9d78/packages/networks/src.ts/index.ts#L132=), but it is kind of a small breaking change. Maybe we can keep `name` as the "display" name and find a better term for the ethers network name?

tmm: A comment saying what the difference is between the "display" name and ethers network name would be good too.

packages/core/src/constants/index.ts
```diff
@@ -10,8 +10,14 @@ export {
   defaultChains,
   defaultL2Chains,
 } from './chains'
+export type { ChainName } from './chains'
```
tmm, line 13: Too bad ethers network names have dashes in them, otherwise `ChainName` could of been the same.

packages/core/src/connectors/walletConnect.ts
```diff
@@ -90,10 +90,23 @@ export class WalletConnectConnector extends Connector<
...
     if (!this.#provider || create) {
+      const rpc = !this.options?.infuraId
+        ? this.chains.reduce(
+            (rpc, chain) => ({
+              ...rpc,
+              [chain.id]: chain.rpcUrls.default,
+            }),
+            {},
+          )
+        : undefined
```
tmm, line 101: should this be `{}` so `...rpc` below works?

packages/core/src/client.ts
```diff
@@ -8,7 +8,7 @@ import { warn } from './utils'
 
 export type ClientConfig<
   TProvider extends providers.BaseProvider = providers.BaseProvider,
-  TWebSocketProvider extends providers.WebSocketProvider = providers.WebSocketProvider,
+  TWebSocketProvider extends providers.BaseProvider = providers.WebSocketProvider,
```
tmm, line 11: Why did this change from `WebSocketProvider` to `BaseProvider`?

tmm: Chatted about this offline

holic:
🎉🙏

PR #436, opened 2022-05-09
refactor: MetaMask and Brave Wallet collision

tmm:
## Description

- Fixes https://github.com/tmm/wagmi/issues/434
- Adds better typing to `window.ethereum` wrapped RPC API

## Additional Information

Your ENS/address: awkweb.eth

### DIFF
--- a/packages/core/src/connectors/injected.ts
+++ b/packages/core/src/connectors/injected.ts
@@ -102,7 +102,7 @@ export class InjectedConnector extends Connector<
   async getAccount() {
     const provider = await this.getProvider()
     if (!provider) throw new ConnectorNotFoundError()
-    const accounts = await provider.request<string[]>({
+    const accounts = await provider.request({
       method: 'eth_requestAccounts',
     })
     // return checksum address
@@ -113,7 +113,7 @@ export class InjectedConnector extends Connector<
     const provider = await this.getProvider()
     if (!provider) throw new ConnectorNotFoundError()
     return await provider
-      .request<string>({ method: 'eth_chainId' })
+      .request({ method: 'eth_chainId' })
       .then(normalizeChainId)
   }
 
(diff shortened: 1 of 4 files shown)


bbondy:
@tmm why do you have this code? It seems like it's causing a lot of webcompatability issues that don't need to be there. For example we would work with https://foundation.app/feed otherwise, but this PR causes it to not be able to connect with Brave Wallet.

tmm:
@bbondy Hi there! Looks like they are using the [`MetaMaskConnector`](https://wagmi.sh/docs/connectors/injected) instead of the [`InjectedConnector`](https://wagmi.sh/docs/connectors/metaMask). The `MetaMaskConnector` only works with MetaMask, while the `InjectedConnector` works with any injected wallet (MetaMask, Brave Wallet, Coinbase Wallet, Tally Ho, etc.). You can see the difference between the two in the [Connect Wallet Example](https://wagmi.sh/examples/connect-wallet).

> why do you have this code?

Sometimes developers want to target specific wallets to do things, like show icons or custom interface. With injected wallets, it's hard to isolate a specific one from the group – e.g. maybe the injected wallet is unknown to the app developer or there are multiple attached to the window (e.g. Coinbase Wallet's [`ethereum.providers`](https://github.com/NoahZinsmeister/web3-react/issues/300#issuecomment-992931744) approach). This code exists to give folks the option to just target MM if they want. In fact, anyone can create a [custom connector](https://wagmi.sh/examples/custom-connector).

Happy to chat more if you have additional questions.

bbondy:
thanks @tmm

bbondy:
If we're API compatible with MetaMask though, what is the benefit of blocking Brave Wallet from the MetaMask option?

bbondy:
To add a bit more depth to my previous message.

Brave Wallet is a wallet like other wallets, but it is also an implementation of the MetaMask API.  That is why we set `isMetaMask` to true. 

There are a lot of Dapps out there using the MetaMask connector only. We find it is important to provide not only an alternate wallet, but  an alternative implementation to the MetaMask API.  This allows for Dapp usage that is built for MetaMask but for people that don't want to be locked into MetaMask's control.  

I have a suggestion that might be best. 

Allow wallets that are compatible to MetaMask with to set isMetaMask to true. And by default allow the MetaMask connector to work with it.  But also expose an option that a Dapp website can use if it truly wants a more strict check.  And don't enable that option by default.

bbondy:
@tmm could you help? ^

PR #442, opened 2022-05-10
use write function like this

paulmars:
args needs to be explicitly blank or explicitly set.

## Description

## Additional Information

- [ ] I read the [contributing docs](/tmm/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

### DIFF
--- a/docs/pages/docs/hooks/useContractWrite.en-US.mdx
+++ b/docs/pages/docs/hooks/useContractWrite.en-US.mdx
@@ -79,7 +79,7 @@ Arguments to pass to function call. Accepts `any | any[]`.
 import { useContractWrite } from 'wagmi'
 
 function App() {
-  const contractWrite = useContractWrite(
+  const {write} = useContractWrite(
     {
       addressOrName: '0xecb504d39723b0be0e3a9aa33d646642d1051ee1',
       contractInterface: wagmigotchiABI,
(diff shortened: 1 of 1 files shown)


paulmars:
i understand your comment to mean that args are optional so the docs should show them as optional. here's some more clear notes about what I was trying to change.

image

in the above image you can see "contractWrite" but "write" is not defined, but "write" is called. write as a function needs to be defined.

image

this image also shows "write" being passed in "onClick". i've found for react, onClick is passed a click event to whatever method is used. so the first "arg" for "write" will be a click event or tap event. which is not the desired value to pass forward to the contract.

you can see in the diff, i wrap "write" with an lambda so that the arg is explicitly blank, vs accidently a click event. i also make write explicitly declared / available in the pull request.

---- 

on a more stylistic personal choice, i'd recommend that the docs provide examples of args for people to see. like one with a uint256, string, and array of something. because we're dealing with JS which is often coerced in to what an API for ETH expects, explicit examples are helpful to me and i assume many other people.  

let me know if anything isn't clear. thanks for all your work, and the site looks great.

paulmars:
no pressure, but when i reply to this is it reopened, or ignored by the process? just not sure if i should make a new PR or not

PR #453, opened 2022-05-13
Rename "API Providers" to "Providers"

jxom:
## Description

Renaming "API Providers" to just "Providers" to be more aligned with ethers terminology. Also replaced references to the React `Provider` with `WagmiProvider` to avoid confusion and ambiguity.


.config/hardhat.config.ts
```diff
@@ -4,7 +4,7 @@ const config: HardhatUserConfig = {
   networks: {
     hardhat: {
       forking: {
-        url: `https://eth-mainnet.alchemyapi.io/v2/${process.env.ALCHEMY_ID}`,
+        url: `https://eth-mainnet.alchemyapi.io/v2/PjT72qifrAFZ4WV_drrd30N5onftY5VA`,
```
tmm, line 7: Hardcoded

docs/components/core/Providers.tsx
```diff
@@ -1,13 +1,18 @@
...
+  defaultChains,
+} from 'wagmi'
 
 import { CoinbaseWalletConnector } from 'wagmi/connectors/coinbaseWallet'
 import { InjectedConnector } from 'wagmi/connectors/injected'
 import { MetaMaskConnector } from 'wagmi/connectors/metaMask'
 import { WalletConnectConnector } from 'wagmi/connectors/walletConnect'
 
-import { alchemyProvider } from '../../../packages/core/apiProviders/alchemy'
+import { alchemyProvider } from '../../../packages/core/providers/alchemy'
```
tmm, line 15: Can we import this from the wagmi package entrypoint?

tmm (commented):
Solid. Can we add a redirect for the Provider docs page too? Can set it up in `_middleware.ts` (there's another redirect already set up there as an example).

.config/hardhat.config.ts
jxom: shite

docs/components/core/Providers.tsx
jxom: hmm, tried to do this, but it was yelling at me

jxom: fixed now

codingwithmanny:
So thankful that we're renaming it to `WagmiProvider`, I usually ended up doing this as default.

codingwithmanny:
So thankful that we're renaming it to `WagmiProvider`, I usually ended up doing this as default.

PR #468, opened 2022-05-16
refactor: rename exports

tmm:
## Description

- Remove duplicate exports with different names and the same functionality to simplify public API
- Use new `Provider` and `WebSocketProvider` types instead of `providers.BaseProvider` and `providers.WebSocketProvider`

## Additional Information

Your ENS/address: awkweb.eth


packages/react/src/index.ts
```diff
@@ -1,17 +1,10 @@
 export {
   createClient,
-  createClient as createWagmiClient,
-  Provider,
-  Provider as WagmiProvider,
+  ClientProvider,
```
jxom, line 3: Chatted offline, but we are going to rename this to `WagmiConfig`

packages/react/src/index.ts
```diff
@@ -1,17 +1,10 @@
...
+  ClientProvider as WagmiProvider,
```
jxom, line 4: awesome! maybe we can deprecate usage of (and not document) `WagmiProvider` too?

PR #469, opened 2022-05-16
add magic connector

nasdf:
## Description

This PR adds a Magic Connector to enable simple Web3 on-boarding.
I've added docs as well as a simple demo to the developer playground so you can see how it functions.

There are some quirks when using the Magic connector because you have to gather user input before calling connect.
I came up with a simple work around to keep the connector APIs unchanged while allowing Magic to get inputs.

(This is documented in the docs as well)

Before connecting you must select one of the available [auth modules](https://magic.link/docs/api-reference/client-side-sdks/web#auth-module).

```ts
connector.loginWithMagicLink({ showUI, email, redirectURI })

connector.loginWithEmailOTP({ email })

connector.loginWithSMS({ phoneNumber })

connector.loginWithCredential(credentialOrQueryString?)
```

## Additional Information

Your ENS/address:
nasdf.eth


tmm:
Thanks for spending the time to put this together! Some thoughts:

Let's create a [new discussion thread](https://github.com/tmm/wagmi/discussions/new) for this to gauge interest and chat more about it.

It's not always possible to add new built-in connectors to wagmi as they increase the surface area for the core team. Ideally @MagicLabs would have an official connector they manage (either as a separate package or in the core repo with their commitment to maintain). @EveripediaNetwork recently open sourced a [`MagicConnector` package](https://github.com/EveripediaNetwork/wagmi-magic-connector) as well (cc @Royal-lobster @kesar).

Finally, if this was to land in wagmi, we would want to follow the exact same Connector API. Using the same API makes the developer experience much easier to reason about.

nasdf:
Thanks for the heads up! I think I'll move the connector to an external package for now.

If possible can there be a section of the readme dedicated to community created plugins?

tmm:
You can [share here](https://github.com/tmm/wagmi/discussions/201) for now. Going to add more structure around this soon.

PR #480, opened 2022-05-20
Re-add Gnosis Chain support

PJColombo:
## Description

This PR re-adds the missing gnosis chain configuration that was included in version [0.1.20](https://github.com/tmm/wagmi/pull/326) which got lost at some point as it doesn't exists anymore.
 
## Additional Information

### DIFF
--- a/packages/core/src/constants/chains.ts
+++ b/packages/core/src/constants/chains.ts
@@ -12,6 +12,7 @@ export type ChainName =
   | 'mainnet'
   | 'optimism'
   | 'optimismKovan'
+  | 'gnosis'
   | 'polygon'
   | 'polygonMumbai'
   | 'rinkeby'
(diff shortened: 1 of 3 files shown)


packages/core/src/constants/chains.ts
```diff
@@ -148,6 +150,17 @@ export const chain: Record<ChainName, Chain> = {
     },
     testnet: true,
   },
+  gnosis: {
+    id: chainId.gnosis,
+    name: 'Gnosis Chain',
+    network: 'gnosis-chain',
```
sembrestels, line 156: It should be just `gnosis`.

tmm:
Thanks for doing this! Going to hold off for now though. I'm working on a way to add a bunch of different chains to the core that is tree-shake-able and easy to maintain.

In the meantime, you can create and use chains in your own project by conforming to the [`Chain` type](https://github.com/tmm/wagmi/blob/main/packages/core/src/types/index.ts#L8-L23=). Related discussion https://github.com/tmm/wagmi/discussions/370#discussioncomment-2650012

PR #526, opened 2022-05-30
feat: `chainId` config for useConnect

jxom:
## Description

This PR adds the ability for consumers to programmatically connect wallets to a given chain ID. I have added a `chainId` config option to `useConnect` as well as the `connect` & `connectAsync` functions that the hook returns.

### Behaviour

Note that there are some behavioural workarounds that needed to be implemented for connectors that don't support native "connect to chain" functionality such as: Injected & Coinbase connectors. The current behaviour is as follows:

- **WalletConnect**: prompt to connect wallet directly to chain via [WalletConnectProvider](https://github.com/tmm/wagmi/pull/526/files#diff-c81c047e9261792bce06826648ac51336629e453cd6f7e501ec54c3fbd7907d8R121)
- **Injected**, **MetaMask** & **Coinbase**: prompt to connect wallet, and then another prompt to switch network

### Flows

**InjectedConnector**, **MetaMaskConnector** & **CoinbaseWalletConnector**:
- If a consumer doesn't provide `chainId` to `useConnect()` or `connect()` , and
  - If the user is currently on a supported chain, it will connect to the chain
  - If the user is currently on an unsupported chain, it will switch chain to the first chain in chains
- If a consumer provides `chainId` to `useConnect()` or `connect()`, and
  - If the user is currently on a supported chain, it will switch chain to the given chainId
  - If the user is currently on an unsupported chain, it will switch chain to the given chainId
  - If the consumer provides an unsupported `chainId`, it'll still connect to it

**WalletConnectConnector**:

- If a consumer doesn't provide `chainId` to `useConnect()` or `connect()`, and
  - If the user hasn't connected to the site before, it will use the first chain in chains
  - If the user has connected to the site before, it will use the previous chain
- If a consumer provides `chainId` to `useConnect()` or `connect()`, and
  - The app supports the `chainId` it'll connect to it
  - The app does not support the `chainId` it will default to whatever the app defaults to.


docs/pages/docs/connectors/coinbaseWallet.en-US.mdx
```diff
@@ -38,6 +38,8 @@ const connector = new CoinbaseWalletConnector({
 })
 ```
 
+Note: Upon connection, the connector will attempt to connect to the first chain in the array unless otherwise specified by a [`chainId` config on useConnect](/docs/hooks/useConnect#chainid-optional).
```
markdalgleish, line 41: These docs need to be updated to mention that it attempts to reconnect to the last used chain.

packages/core/src/connectors/coinbaseWallet.ts
```diff
@@ -57,8 +57,17 @@ export class CoinbaseWalletConnector extends Connector<
...
       const account = getAddress(accounts[0])
-      const id = await this.getChainId()
-      const unsupported = this.isChainUnsupported(id)
+
+      let id = await this.getChainId()
+      const unsupported = this.isChainUnsupported(chainId ?? id)
+      if (chainId && id !== chainId) {
+        const chain = await this.switchChain(chainId)
+        id = chain.id
+      } else if (unsupported) {
```
markdalgleish, line 66: It looks like `unsupported` will still be returned to the consumers as `true` once this branch is finished even though it's switched to a supported chain?

jxom: ah yep, fixing.

packages/core/src/connectors/metaMask.ts
```diff
@@ -28,6 +29,18 @@ export class MetaMaskConnector extends InjectedConnector {
...
+    // upon switching chains. This workaround ensures that the
+    // user currently isn't in the process of switching chains.
+    const onDisconnect = this.onDisconnect
+    super.onDisconnect = () => {
+      if (this.#switchingChains) {
+        this.#switchingChains = false
+        return
+      }
+      onDisconnect()
+    }
```
tmm, line 43: Think we should expose this behind a `shim` flag? This a more robust version of something I did before in https://github.com/tmm/wagmi/pull/190 (removed since I thought MM fixed the issue), but there could be unintended side effects. Maybe we default the flag to `true` since it seems like something most will want.

Let's link to this issue too: https://github.com/MetaMask/metamask-extension/issues/13375#issuecomment-1087023057

docs/pages/docs/connectors/walletConnect.en-US.mdx
```diff
@@ -1,4 +1,4 @@
-# WalletConnect
+'# WalletConnect
```
markdalgleish, line 1: Typo?

jxom: wtf how did that get in there

packages/react/src/hooks/accounts/useConnect.ts
```diff
@@ -70,14 +73,38 @@ export function useConnect({
...
+          chainId,
+          connector: connector_ ?? connector,
+        })
+      }
+      const args = connectorOrArgs
+      return mutate(<ConnectArgs>{
+        chainId: args?.chainId ?? chainId,
+        connector: args?.connector ?? connector,
+      })
+    },
```
tmm, line 89: Think it's easier to construct these args and pass to `mutate` once or construct independently and call twice?

packages/core/src/client.ts
```diff
@@ -204,11 +207,15 @@ export class Client<
...
   async destroy() {
     if (this.connector) await this.connector.disconnect?.()
+    this.#autoConnecting = false
     this.clearState()
     this.store.destroy()
   }
 
   async autoConnect() {
+    if (this.#autoConnecting) return
+    this.#autoConnecting = true
```
tmm, line 217: Is `#autoConnecting` for making sure this function doesn't run multiple times simultaneously?

Can we change `#autoConnecting` to `#isAutoConnecting` too.

tmm: Random aside about `autoConnect`: Was thinking we could remove the `React.useEffect` in `WagmiConfig` and call the `autoConnect` function inside the client (e.g. constructor) instead. This way `createClient({ autoConnect: true })` in the Vanilla JS version would work without calling separately.

Haven't spent the time yet to figure out the best way to do this reliably across different rendering environments though (React versus Vanilla, client versus SSR).

packages/core/src/client.ts
```diff
@@ -224,12 +231,17 @@ export class Client<
       const isAuthorized = await connector.isAuthorized()
       if (!isAuthorized) continue
 
-      const data = await connector.connect()
+      let data: ConnectorData
+      try {
+        data = await connector.connect()
+      } catch {
+        // Swallow connect errors
```
tmm, line 238: Was this throwing before in some situations?

packages/core/src/connectors/walletConnect.ts
```diff
@@ -98,9 +110,13 @@ export class WalletConnectConnector extends Connector<
     return chainId
   }
 
-  async getProvider(create?: boolean) {
+  async getProvider(
+    create?: boolean,
+    {
+      chainId = this.options?.chainId || this.chains[0].id,
+    }: { chainId?: number } = {},
+  ) {
```
tmm, line 118: Maybe we should change `getProvider`'s interface from:

```ts
abstract getProvider(create?: boolean): Promise<Provider>
```

To an object:

```ts
abstract getProvider(config: { chainId?: number; create?: boolean } = {}): Promise<Provider>
```

Makes things more flexible in the future if we need to add on more params (maybe for [multi-chain support](https://github.com/tmm/wagmi/discussions/396) eventually).

packages/core/src/connectors/walletConnect.ts
```diff
@@ -11,6 +11,7 @@ import {
 import { Chain } from '../types'
 import { normalizeChainId } from '../utils'
 import { Connector } from './base'
+import { getClient } from '../client'
```
tmm, line 14:
```suggestion
import { getClient } from '../client'
import { Chain } from '../types'
import { normalizeChainId } from '../utils'
import { Connector } from './base'
```

tmm: ^ just reordering

tmm (commented):
Some comments, but overall solid. Do we want to make the switching chains on connect part optional? At least for connectors that don't supporting connecting directly to a chain. 

Getting shown two prompts back-to-back could feel like a lot, especially if there's no need to be on a "supported chain" immediately (e.g. not signing/creating transactions immediately).

packages/core/src/connectors/metaMask.ts
jxom: Sounds good!

packages/core/src/client.ts
jxom: Yeah, it was throwing unexpectedly when attempting to switch chains... Not sure if we need to throw errors in `autoConnect` if we are trying to eagerly connect?

packages/core/src/connectors/walletConnect.ts
jxom: Sounds good!

docs/pages/docs/connectors/injected.en-US.mdx
```diff
@@ -25,6 +25,20 @@ const connector = new InjectedConnector()
 
 ## Configuration
 
+### allowConnectToUnsupportedChain
```
tmm, line 28: Maybe we should call this `switchChainOnConnect` instead to more accurately represent what this does?

`allowConnectToUnsupportedChain` implies that someone can't switch to unsupported chains, but most wallet apps will still allow this through their interface.

packages/core/src/client.ts
tmm: Can we disable chain switching during auto connect? Getting a wallet prompt when the page is loading seems like an anti-pattern.

packages/core/src/connectors/base.ts
```diff
@@ -70,9 +69,13 @@ export abstract class Connector<
   protected abstract onDisconnect(error: Error): void
 
   protected getBlockExplorerUrls(chain: Chain) {
-    const blockExplorer = chain.blockExplorers?.default
-    if (Array.isArray(blockExplorer)) return blockExplorer.map((x) => x.url)
```
tmm, line 68: this didn't work before 😛

PR #552, opened 2022-06-04
fix(react): out-of-sync persisted contract results

### LINKED ISSUE #460: [bug] useContractRead not returning key/value pairs after second page refresh

### Package Version

0.3.5

### Current Behavior

Hey all,

I was using `useContractWrite` to write something to my local Hardhat, and then refreshed the page which triggers `useContractRead` and noticed that first refresh I get the following (note the key/value pairs):

Screen Shot 2022-05-14 at 12 52 01 PM

And then I refresh the page again, and I get this (no key/value pairs) in the console:

Screen Shot 2022-05-14 at 12 52 13 PM

Is there something obvious I'm missing as to why the key/value pairs disappear after a couple refreshes when I'm returning an array of structs? Or is this a bug?

Also not sure if this is an `ethers` issue or if it's a `wagmi` issue.

### Expected Behavior

The expected behavior is for it to return the struct with key/value pairs like so:

Screen Shot 2022-05-14 at 12 52 01 PM

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

jxom:
Fixes #460

## Description

There is an interesting bug where persisted `useContractRead` struct results would be out of sync with the ethers contract `Result`. This is because the ethers `Result` currently doesn't serialize properly with our `serialize` function. I first attempted to try fix this in our `serialize`/`deserialize` functions to try get it to persist into our storage, however, after doing that, I then realised that React Query also doesn't cater for this data structure in their data comparison functions. 

In this proposed fix, I ended up just checking to see if the result returned from RQ is in an out-of-sync state, and if it is, it will use the contract interface to parse it back into an ethers contract `Result`.


packages/react/src/hooks/contracts/useContractRead.ts
```diff
@@ -127,7 +127,7 @@ export function useContractRead(
     watch,
   ])
 
-  return useQuery(queryKey_, queryFn, {
+  const contractReadQuery = useQuery(queryKey_, queryFn, {
```
tmm, line 130: Would react-query's `select` config option work instead of capturing the whole result? That way, the return value still takes advantage of the tracked behavior.

```
select: (data: TData) => unknown
Optional
This option can be used to transform or select a part of the data returned by the query function.
```

packages/react/src/hooks/utils/index.ts
```diff
@@ -1,3 +1,4 @@
+export { parseContractResult } from './parseContractResult'
```
tmm, line 1: Let's move this function to `src/utils` rather than `src/hooks/utils`.

packages/react/src/hooks/utils/parseContractResult.test.ts
```diff
@@ -0,0 +1,47 @@
...
+const contractInterface = [
+  {
+    inputs: [
+      {
+        internalType: 'uint256',
+        name: '',
+        type: 'uint256',
+      },
+    ],
+    name: 'gms',
```
tmm, line 14: gm

packages/react/src/hooks/utils/parseContractResult.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        type: 'address',
+      },
+    ],
+    stateMutability: 'view',
+    type: 'function',
+  },
+]
+
+describe('parseContractResult', () => {
+  it('should parse the data to an ethers Result if there are no named keys', () => {
```
tmm, line 33: Can we add one more test for a plan result that skips this logic? Could use mloot or wagmigotchi.

tmm (commented):
Nice job hunting this down.

packages/react/src/hooks/utils/index.ts
jxom: Oops, didn't realise!

packages/react/src/hooks/contracts/useContractRead.ts
jxom: Forwarding from convo: I tried using `select`, but it doesn't work b/c of the data comparison bug i found in RQ lol

tmm: Lol figures. Maybe worthwhile opening up something over there and we try to find another workaround. Losing the tracked result isn't ideal to fix this case. What do you think?

jxom: yeah sounds good. 

https://github.com/tannerlinsley/react-query/pull/3669

tmm (approved):
Looks great! Nice job on getting RQ update in.

0xfffxfff:
@jxom Since updating to wagmi@0.4.10 I'm getting errors with *all* my instances of useContractRead:

```
next-dev.js?3515:25 TypeError: Cannot add property null, object is not extensible
    at Function.assign (<anonymous>)
    at parseContractResult (wagmi.esm.js?20d3:73:1)
    at Object.select (wagmi.esm.js?20d3:1275:1)
    at QueryObserver.createResult (queryObserver.mjs?eb18:273:1)
    at QueryObserver.getOptimisticResult (queryObserver.mjs?eb18:92:1)
    at useQuery (wagmi.esm.js?20d3:365:5)
    at useContractRead (wagmi.esm.js?20d3:1272:1)
```
For now I'm reverting to 0.4.7 our previous version. Any idea what this could be caused by? 


```bash
npm view wagmi dependencies
{
  '@coinbase/wallet-sdk': '^3.2.0',
  '@wagmi/core': '^0.3.7',
  '@walletconnect/ethereum-provider': '^1.7.8',
  'react-query': '^4.0.0-beta.23',
  'use-sync-external-store': '^1.1.0'
}
```

jxom:
@0xfffxfff would you be able to provide me with a minimal reproducible example via a CodeSandbox or something?
Thanks for spotting this too!

0xfffxfff:
@jxom We're in the middle of a release right now. I can try spinning one up on Monday.

0xfffxfff:
Just to give you an idea where this happens:
```ts
const balanceOf = useKubboPartsRead('balanceOf', { args: account?.data?.address, enabled: Boolean(account?.data?.address), watch: true })
```
where `useKubboPartsRead` is a hook that wraps `useContractRead`
```ts
export const useKubboPartsRead = (
  functionName: string, config?: Parameters<typeof useContractRead>[2]
) => {
  const network = useNetwork()
  const contract = Contracts.KubboParts[network?.activeChain?.id || chain.mainnet.id];
  return useContractRead(contract, functionName, config)
}
```
Entirely possible that my use of the hook is causing this.

jxom:
Thanks! I'll do some digging.

0xfffxfff:
Thanks @jxom! This fixed it.

> I'm assuming the return type of this method was a tuple?
 It was an array of uint256. Funnily enough I had to change a useEffect dependency from this:
 ```ts
   }, [
    tokensOf.data,
    ...
  ]);
 ```
 to this:
  ```ts
   }, [
    tokensOf.data?.reduce((c,i) => c+i?.toString()
    ...
  ]);
 ```
I'm guessing the reason is that it's no longer receiving the same array but rather a newly constructed one on each call? In any case, with this in place everything's now properly chugging along 👍

PR #564, opened 2022-06-07
Add Foundry To Chains Export

carllippert:
Foundry is becoming more popular. This makes it easy for all.

## Description

Make Foundry as easy as hardhat

## Additional Information

Your ENS/address:

carllippert.eth

### DIFF
--- a/packages/core/src/constants/chains.ts
+++ b/packages/core/src/constants/chains.ts
@@ -31,6 +31,7 @@ export const chainId = {
   arbitrumRinkeby: 421_611,
   localhost: 1_337,
   hardhat: 31_337,
+  foundry: 31337
 } as const
 
 export const chain: Record<ChainName, Chain> = {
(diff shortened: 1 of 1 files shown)


packages/core/src/constants/chains.ts
```diff
@@ -31,6 +31,7 @@ export const chainId = {
   arbitrumRinkeby: 421_611,
   localhost: 1_337,
   hardhat: 31_337,
+  foundry: 31337
```
jxom, line 34:
```suggestion
  foundry: 31_337
```

tmm (changes requested):
We should export this from the tree shakable chains entrypoint in each package.

Also, should we call this Foundry Anvil?

carllippert:
Was thinking the same about Anvil naming! Someone from that team probably has an opinion but don't know them.

tmm:
Closing in favor of https://github.com/tmm/wagmi/pull/570

PR #572, opened 2022-06-10
[FIX] Remove ReloadOnDisconnect for CoinbaseWallet Connector

jeongmincho:
## Description

@tmm hi! we noticed you already implemented our latest update on the SDK for disabling reload on disconnect to the library. however, the changes on our extension haven't been release to the public yet, so you'll notice the integration is breaking on the docs (and probably everywhere with the new update). we would like to revert this change for the time being until our products get shipped with the functionality.

Screen Shot 2022-06-10 at 12 37 40 PM
Screen Shot 2022-06-10 at 12 38 19 PM

## Additional Information

- [ ] I read the [contributing docs](/tmm/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

### DIFF
--- a/packages/core/src/connectors/coinbaseWallet.ts
+++ b/packages/core/src/connectors/coinbaseWallet.ts
@@ -46,7 +46,6 @@ export class CoinbaseWalletConnector extends Connector<
     super({
       chains,
       options: {
-        reloadOnDisconnect: false,
         ...options,
       },
     })


jeongmincho:
closing this out after confirming our latest extension version indeed does have the necessary changes. thanks!

PR #573, opened 2022-06-11
fix: eslint config

jxom:
## Description

Turns out that running `pnpm eslint` wasn't actually running eslint as we were not specifying a target directory. This PR adds that in, and also fixes eslint config issues arising from including a target directory.


.eslintrc
```diff
@@ -15,8 +15,7 @@
     "eslint:recommended",
     "plugin:eslint-comments/recommended",
     "plugin:@typescript-eslint/recommended",
-    "plugin:import/errors",
-    "plugin:import/warnings",
+    "plugin:import/recommended",
```
jxom, line 18: this is a combination of `import/errors` and `import/warnings`

.eslintrc
```diff
@@ -56,7 +56,7 @@
   },
   "overrides": [
     {
-      "files": ["*.mdx"],
+      "files": "**/*.mdx/**",
```
jxom, line 59: `*.mdx` wasn't working well with TS. this seemed to fix it.

.node-version
```diff
@@ -1 +1 @@
-14.19
+16
```
tmm, line 1: bumping this since it's what ci and vercel both run on

PR #611, opened 2022-06-26
feat: add chainId config to useContractWrite and useSendTransaction

tmm:
## Description

Adds `chainId` configuration parameter to `useContractWrite` and `useSendTransaction`. If `chainId` is provided, the connector will validate that `chainId` is the active chain before sending a transaction (and switch to `chainId` if necessary).

## Additional Information

Your ENS/address: awkweb.eth


jxom (approved):

holic:
😍

examples/_dev/src/components/Account.tsx
```diff
@@ -18,8 +18,9 @@ export const Account = () => {
   })
   const ensAvatar = useEnsAvatar({
     addressOrName: account?.address,
+    chainId: 1,
```
holic, line 21: Should this just be the default for ENS queries? I feel like you more often want the ENS data on mainnet. It's rare that I see someone with an e.g. rinkeby ENS set up properly.

packages/core/src/actions/accounts/switchNetwork.ts
```diff
@@ -11,10 +14,12 @@ export type SwitchNetworkResult = Chain
 export async function switchNetwork({
   chainId,
 }: SwitchNetworkArgs): Promise<SwitchNetworkResult> {
-  const client = getClient()
-  const connector = client.connector
+  const { connector } = getClient()
+  if (!connector) throw new ConnectorNotFoundError()
   if (!connector?.switchChain)
```
holic, line 19: nit: don't need the optional chaining here anymore (`?.`)

examples/_dev/src/components/Account.tsx
tmm: Could make sense. RainbowKit might be heading [in that direction](https://github.com/rainbow-me/rainbowkit/pull/496). Should think about if there are any arguments against this and address in a separate PR.

PR #646, opened 2022-07-01
Use correct chainId in Foundry chain definition

lochie:
## Description

Small fix to update the `id` definition for the chain `foundry` as it was using `chainId.hardhat` instead of `chainId.foundry`.

Possibly worth mentioning that the values are the same for `hardhat` and `foundry` in the [`chainId` constant](https://github.com/tmm/wagmi/blob/e4bd79fe8da24e80100515c516a358ccdd71509a/packages/core/src/constants/chains.ts#L18-L19), not sure if these IDs are actually supposed to be the same. 🤷‍♀️
```
hardhat: 31_337,
foundry: 31_337,
```

## Additional Information

Your ENS/address: `lochie.eth`

### DIFF
--- a/packages/core/src/constants/chains.ts
+++ b/packages/core/src/constants/chains.ts
@@ -309,7 +309,7 @@ export const hardhat: Chain = {
 }
 
 export const foundry: Chain = {
-  id: chainId.hardhat,
+  id: chainId.foundry,
   name: 'Foundry',
   network: 'foundry',
   rpcUrls: {


tmm:
Yup, both are the same. Foundry was trying to be compatible for folks switching from Hardhat if I remember correctly:
- [Hardhat](https://github.com/NomicFoundation/hardhat/blob/ad343502b26654991c5a40d891a975ea4218b089/packages/hardhat-core/src/internal/core/config/default-config.ts#L44=)
- [Foundry](https://github.com/foundry-rs/foundry/blob/0c2ceccf866a356ac6dcaa3446cd56a56cde4095/common/src/lib.rs#L11=)

PR #647, opened 2022-07-01
Update useNetwork.en-US.mdx

jelilat:
## Description

According to the documentation, `chain` is one of `useNetwork()`'s return values. However, I realised that it doesn't return `chain`, but returns `activeChain` instead. Therefore, using what's currently in the documentation will result in an error.

So, I updated the documentation to fix this.

### Documentation

Screenshot 2022-07-01 at 15 36 35

### Actual Return Value

Screenshot 2022-07-01 at 15 39 15

## Additional Information

Your ENS/address: tjelailah.eth

### DIFF
--- a/docs/pages/docs/hooks/useNetwork.en-US.mdx
+++ b/docs/pages/docs/hooks/useNetwork.en-US.mdx
@@ -17,11 +17,11 @@ import { useNetwork } from 'wagmi'
 import { useNetwork } from 'wagmi'
 
 function App() {
-  const { chain, chains } = useNetwork()
+  const { activeChain, chains } = useNetwork()
 
   return (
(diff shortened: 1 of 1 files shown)


jelilat:
cc @tmm

tmm:
What version of wagmi are you targeting?

`useNetwork`'s API [changed slightly in `0.5.x`.](https://wagmi.sh/docs/migration-guide#usenetwork=)

jelilat:
Oh. I'm using 0.4.12

Now I get it now. Thanks

tmm:
@jelilat [`0.4.x` docs](https://0.4.x.wagmi.sh) if you need them :)

PR #654, opened 2022-07-03
fix: fix multicall returning nullish data for all calls unexpectedly

### LINKED ISSUE #645: [bug] useContractReads returning [null] or undefined data unexpectedly

### Package Version

latest

### Current Behavior

There was a previous issue which was closed but had reported the same issue. Maybe there is some human error here but I'm struggling to find it. I've added a repro to illustrate this one with the same data.

### Expected Behavior

Work

### Steps To Reproduce

repro attached

https://codesandbox.io/s/pensive-danny-fzwj5g?file=/src/App.js

### Anything else?

We are migrating a large amount of our codebase from redux multicall to wagmi multicall, but have run into a number of issues so far so we are halting and will help illustrate them until there is more stability.

An example of this hook migration from redux multicall to wagmi multicall.

Screenshot_272

jxom:
Fixes #645

## Description

When a contract method returned zero data (`0x`) for any call in a multicall, this would cause the function to throw an error, resulting in the multicall returning nullish data unexpectedly. This PR guards for the case where a call could return zero data, and warns the user that the function does not exist on the targeted contract.

### DIFF
--- a/packages/core/src/actions/contracts/multicall.ts
+++ b/packages/core/src/actions/contracts/multicall.ts
@@ -7,6 +7,7 @@ import { multicallInterface } from '../../constants'
 import { ReadContractConfig } from './readContract'
 import {
   ChainDoesNotSupportMulticallError,
+  ContractMethodNoDataError,
   ProviderChainsNotFound,
 } from '../../errors'
 
@@ -79,7 +80,7 @@ export async function multicall<Data extends any[] = Result[]>({
         )
       return {
         target: addressOrName,
-        allowFailure: allowFailure,
+        allowFailure,
         callData,
       }
     },
@@ -93,6 +94,18 @@ export async function multicall<Data extends any[] = Result[]>({
     const { addressOrName, contractInterface, functionName } = <
       MulticallContract
     >contracts[i]
+
+    if (returnData === '0x') {
+      const err = new ContractMethodNoDataError({
+        addressOrName,
+        blockExplorer: chain.blockExplorers?.default,
+        functionName,
+      })
+      if (!allowFailure) throw err
+      console.warn(err.message)
+      return undefined
+    }
+
     const contract = getContract({
       addressOrName,
       contractInterface,
--- a/packages/core/src/actions/contracts/multicall.ts
+++ b/packages/core/src/actions/contracts/multicall.ts
@@ -7,7 +7,7 @@ import { multicallInterface } from '../../constants'
 import { ReadContractConfig } from './readContract'
 import {
   ChainDoesNotSupportMulticallError,
-  ContractMethodNoDataError,
+  ContractMethodNoResultError,
   ProviderChainsNotFound,
 } from '../../errors'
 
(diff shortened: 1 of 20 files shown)


packages/core/src/actions/contracts/multicall.ts
```diff
@@ -93,6 +94,18 @@ export async function multicall<Data extends any[] = Result[]>({
...
     >contracts[i]
+
+    if (returnData === '0x') {
+      const err = new ContractMethodNoDataError({
+        addressOrName,
+        blockExplorer: chain.blockExplorers?.default,
+        functionName,
+      })
+      if (!allowFailure) throw err
+      console.warn(err.message)
```
jxom, line 105: thought it would be nice to give feedback to the user as to why this is happening. though understand it could bloat the console especially if these errors are expected (i.e. a derived address that is non-existent). wonder if this is okay? or should it be under a flag? or just remove it altogether?

packages/react/src/hooks/contracts/useContractReads.ts
```diff
@@ -20,29 +20,44 @@ export type UseContractReadsConfig = QueryConfig<ReadContractsResult, Error> &
...
       overrides,
     },
   ] as const
 
+const queryKeyHashFn = ([queryKey_]: ReturnType<typeof queryKey>) => {
+  const { contracts, ...rest } = queryKey_
+  const contracts_ = contracts?.map((contract) => {
+    // Exclude the contract interface from the serialized query key.
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { contractInterface, ...rest } = contract
```
jxom, line 43: `contractInterface` bloated the hashed query key and caused a weird bug where data would persist on the wrong query key.

packages/react/src/hooks/contracts/useContractReads.ts
```diff
@@ -20,29 +20,44 @@ export type UseContractReadsConfig = QueryConfig<ReadContractsResult, Error> &
...
 }: QueryFunctionArgs<typeof queryKey>) => {
   return readContracts({
+    allowFailure,
     contracts,
     overrides,
   })
 }
 
 export function useContractReads({
+  allowFailure = true,
```
jxom, line 60: should also include `allowFailure` as a prop here.

packages/core/src/actions/contracts/multicall.ts
```diff
@@ -93,14 +94,31 @@ export async function multicall<Data extends any[] = Result[]>({
     const { addressOrName, contractInterface, functionName } = <
       MulticallContract
     >contracts[i]
+
+    if (returnData === '0x') {
```
jxom, line 98: this is the guts of this PR

packages/core/src/actions/contracts/multicall.ts
tmm: Good point! There are a few places where we display warning messages already. Might be nice to make this configurable at some point (e.g. you can switch off warnings, only show in dev mode, etc.).

packages/core/src/errors.ts
```diff
@@ -110,6 +111,34 @@ export class ConnectorNotFoundError extends Error {
   message = 'Connector not found'
 }
 
+export class ContractMethodNoResultError extends Error {
```
tmm, line 114: Should we export this from the various `index.ts`s like the other errors?

jxom: Yes!

PR #658, opened 2022-07-04
Prepare Hooks

jxom:
This PR introduces "eager" hooks for "send transaction" & "write contract". 

h/t @markdalgleish for doing all the investigation into this

### Problem

The current implementations of `useSendTransaction` and `useContractWrite` don't work very well when a user is connected via WalletConnect, Safari Extension, or any connector that deeplinks to an app on iOS devices. When a user tries to send a transaction, they will always see an iOS prompt to open the app or be redirected to a universal link fallback page/app store. See the video below when I click "Send Transaction".

Connected via Rainbow:

https://user-images.githubusercontent.com/7336481/177080668-e9d69765-f363-4ebf-aa95-da53f316e7a8.MP4

Connected via MetaMask:

https://user-images.githubusercontent.com/7336481/177080676-32fdfa72-ec01-4d70-bb79-f926e2e067d0.MP4

Turns out, this is not specifically an issue on `wagmi`, but more a side-effect of using an [ethers `signer.sendTransaction`](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L269) inside of an event handler (e.g. "Send Transaction" button `onClick` handler) – the `sendTransaction` function performs too much async work in the event handler (see [here](https://github.com/ethers-io/ethers.js/blob/master/packages/providers/src.ts/json-rpc-provider.ts#L271), [here](https://github.com/ethers-io/ethers.js/blob/608864fc3f00390e1260048a157af00378a98e41/packages/providers/src.ts/json-rpc-provider.ts#L226), and [here](https://github.com/ethers-io/ethers.js/blob/608864fc3f00390e1260048a157af00378a98e41/packages/providers/src.ts/json-rpc-provider.ts#L232)), causing the deeplinking to break as iOS _thinks_ the user did not explicitly initiate the request.

### Solution

Most of the async work in `sendTransaction` comes from calculating the gas limit (`estimateGas`) & resolving an ENS address (if needed). The proposed solution is to build the arguments required for `sendTransaction` eagerly (at mount time), and then pass those arguments in when the user triggers the event handler, so that `sendTransaction` performs minimal/no async work. This can be done via "prepare" hooks: `usePrepareSendTransaction` & `usePrepareContractWrite`. 

### TODOs

- [x] `prepareSendTransaction`
  - [x] changesets
- [x] `prepareContractWrite`
  - [x] changesets
- [x] `writeContract`
  - [x] changesets
- [x] `sendTransaction`
  - [x] changesets
- [x] `usePrepareSendTransaction`
  - [x] docs
  - [x] changesets
- [x] `usePrepareContractWrite`
  - [x] docs
  - [x] changesets
- [x] `useSendTransaction`
  - [x] docs 
- [x] `useContractWrite`
  - [x] docs
- [x] examples
- [x] docs: page on TTOW (time to open wallet) and it's side effects
- [ ] 0.6 blog post: "introducing prepare mode" section or something


docs/components/docs/ValidationComparisonTable.tsx
```diff
@@ -0,0 +1,47 @@
...
+                The user will see feedback after they click "Mint".
+              </label>
+            </div>
+          </td>
+          <td>
+            <div style={{ maxWidth: 300 }}>
+              <video
+                id="eager"
+                controls
+                src="/docs/lazy-mint.mp4"
```
markdalgleish, line 34: This should be the eager vid.

docs/pages/docs/hooks/useContractWrite.en-US.mdx
```diff
@@ -39,24 +57,47 @@ function App() {
...
 ```
 
 ## Configuration
 
+### mode
+
+- `dangerouslyUnprepared`: Allow to pass through an adhoc unprepared config. Note: This has [UX pitfalls](/docs/prepare-hooks/intro#ux-pitfalls-without-prepare-hooks), it is highly recommended to not use this and instead prepare the config upfront using the `usePrepareContractWrite` hook.
+- `prepared`: The config has been prepared with parameters required for performing a contract write via the [`usePrepareContractWrite` hook](/docs/prepare-hooks/usePrepareContractWrite)
+
+> This is automatically populated when using [`usePrepareContractWrite` hook](/docs/prepare-hooks/usePrepareContractWrite).
```
markdalgleish, line 74: IMO these notes should be immediately under the headings.

docs/pages/docs/prepare-hooks/intro.en-US.mdx
```diff
@@ -0,0 +1,82 @@
...
+description: 'What are Prepare Hooks?'
+---
+
+import { ValidationComparisonTable } from '../../../components/docs/ValidationComparisonTable'
+
+# Prepare Hooks
+
+In previous versions of wagmi, "transactional" hooks such as `useSendTransaction` and `useContractWrite` performed potentially long-running asynchronous work. These hooks internally invoke the [ethers `signer.sendTransaction` function](https://docs.ethers.io/v5/api/signer/#Signer-sendTransaction) that fetches the gas estimate & block number, and resolves an ENS name (if needed) before the transaction gets sent. On a slow network, these asynchronous tasks can take a while.
+
+Performing asynchronous tasks in an event handler (like a button) or in a non-initiated user event (like a page load) is not ideal as it can lead to a variety of connected user-experience (UX) related pitfalls ([listed below](#ux-pitfalls-without-prepare-hooks)).
```
markdalgleish, line 12:
```suggestion
Performing asynchronous tasks in an event handler (like a button) or in an event that wasn't initiated by the user (like a page load) is not ideal as it can lead to a variety of connected user-experience (UX) related pitfalls ([listed below](#ux-pitfalls-without-prepare-hooks)).
```

docs/pages/docs/prepare-hooks/usePrepareSendTransaction.en-US.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+Eagerly fetches the parameters required for sending a transaction such as the gas estimate and resolving an ENS address (if required).
+
+```tsx
+import { usePrepareSendTransaction } from 'wagmi'
+```
+
+## Usage
+
+`usePrepareSendTransaction` gives back a "prepared request" to be sent through to `useSendTransaction`.
```
markdalgleish, line 18: Should this be "prepared config"?
```suggestion
`usePrepareSendTransaction` gives back a "prepared config" to be sent through to `useSendTransaction`.
```

docs/pages/examples/send-transaction.en-US.mdx
```diff
@@ -0,0 +1,283 @@
...
+title: 'Send Transaction'
+description: 'Learn how to send transactions with wagmi.'
+---
+
+import Callout from 'nextra-theme-docs/callout'
+import { SendTransaction } from '../../components/examples'
+
+# Send Transaction
+
+The following example teaches you how to send transactions in wagmi. The example below builds on the [Connect Wallet Example](/examples/connect-wallet) and uses the [`usePrepareSendTransaction`](/docs/prepare-hooks/usePrepareSendTransaction) & [`useSendTransaction`](/docs/hooks/useSendTransaction) hooks. Try it out before moving on.
```
markdalgleish, line 11:
```suggestion
The following example teaches you how to send transactions in wagmi. The example below builds on the [Connect Wallet Example](/examples/connect-wallet) and uses the [`usePrepareSendTransaction`](/docs/prepare-hooks/usePrepareSendTransaction), [`useSendTransaction`](/docs/hooks/useSendTransaction) & [`useWaitForTransaction`](/docs/hooks/useWaitForTransaction) hooks. Try it out before moving on.
```

packages/core/src/actions/transactions/sendTransaction.ts
```diff
@@ -1,54 +1,111 @@
...
+    // when using it in a click handler (iOS deep linking issues,
+    // delay to open wallet, etc).
+
+    const uncheckedSigner = signer.connectUnchecked()
+    const { hash, wait } = await uncheckedSigner.sendTransaction(request)
+
+    /********************************************************/
+    /** END: WalletConnect mobile deep link cautious code.  */
+    /** Go nuts!                                            */
+    /********************************************************/
```
markdalgleish, line 106: Nice comments in this file 👍

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export const queryKey = (
+  {
+    args,
+    addressOrName,
+    contractInterface,
+    functionName,
+    overrides,
+  }: UsePrepareContractWriteArgs,
+  { chainId, signer }: { chainId?: number; signer?: FetchSignerResult },
```
tmm, line 30: Is the first object the public values and second object the internal values?

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
```diff
@@ -0,0 +1,136 @@
...
+      addressOrName,
+      args,
+      contractInterface,
+      overrides,
+      functionName,
+      request: undefined,
+      mode: 'prepared',
+      ...prepareContractWriteQuery.data,
+    } as PrepareWriteContractResult,
+  })
```
tmm, line 135: How does this affect the tracked behavior of `useQuery`?

packages/react/src/hooks/transactions/useSendTransaction.ts
```diff
@@ -45,25 +114,43 @@ export function useSendTransaction({
...
+    (args?: UseSendTransactionMutationArgs) =>
+      mutateAsync({
+        chainId,
+        mode,
+        request: args?.dangerouslySetRequest ?? request,
+      } as SendTransactionArgs),
+    [chainId, mode, mutateAsync, request],
   )
 
+  useAccount({ onDisconnect: () => reset() })
```
tmm, line 152: Comment

tmm (commented):
Looking solid!
- Should we default "transaction" hooks to a specific `mode`?
- Do we want to type "prepared" address values (e.g. `to`) as `0x${string}`?
- Since we are removing automatic chain switching for "transaction" hooks, can we also remove from [`signTypedData`](https://github.com/tmm/wagmi/blob/jxom/eager/packages/core/src/actions/accounts/signTypedData.ts#L50=)?

packages/core/src/actions/contracts/writeContract.ts
```diff
@@ -1,82 +1,101 @@
...
+   * it is highly recommended to not use this and instead prepare the request upfront
+   * using the {@link prepareWriteContract} function.
+   *
+   * `prepared`: The request has been prepared with parameters required for sending a transaction
+   * via the {@link prepareWriteContract} function
+   * */
+  mode: 'prepared'
+  /** The prepared request. */
+  request: PopulatedTransaction & {
+    to: NonNullable<PopulatedTransaction['to']>
```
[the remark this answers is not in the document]
jxom: Sounds good!

packages/core/src/actions/transactions/prepareSendTransaction.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export async function prepareSendTransaction({
+  chainId,
+  request,
+  signerOrProvider = getProvider(),
+}: PrepareSendTransactionArgs): Promise<PrepareSendTransactionResult> {
+  const [to, gasLimit] = await Promise.all([
+    isAddress(request.to)
+      ? Promise.resolve(request.to)
+      : fetchEnsAddress({ name: request.to }),
```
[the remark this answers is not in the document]
jxom: Defs! Maybe we can put this in a separate PR from this one?

packages/core/src/actions/transactions/sendTransaction.ts
```diff
@@ -1,54 +1,111 @@
...
-export type SendTransactionResult = providers.TransactionResponse
+export type SendTransactionArgs = {
+  /** Chain ID used to validate if the signer is connected to the target chain */
+  chainId?: number
+} & (SendTransactionPreparedRequest | SendTransactionUnpreparedRequest)
+
+export type SendTransactionResult = {
+  hash: providers.TransactionResponse['hash']
+  wait: providers.TransactionResponse['wait']
+}
```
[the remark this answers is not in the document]
jxom: The original `signer.sendTransaction` made an additional request to the RPC to fetch the transaction data after the transaction was sent. Since this RPC request can be slow & not all consumers need the whole object, I thought it would be better to only return what is given from the unchecked `signer.sendTransaction`, and then if the consumer needs more properties from the transaction, then they can call `useWaitForTransaction` with `confirmations: 0`. [See here](https://github.com/tmm/wagmi/pull/658/files/dc3c388d31f9ceb830f77f9cd7e5e274dc96ed2e#diff-93ee9dd40c1bffb44416caccad93c8bb95d55c09df0156aa2a41855c140be6c7R1)

packages/react/src/hooks/contracts/useContractWrite.ts
```diff
@@ -88,49 +171,59 @@ export function useContractWrite({
...
       functionName,
+      mode,
       mutateAsync,
       overrides,
-      signerOrProvider,
+      request,
     ],
   )
 
+  useAccount({ onDisconnect: () => reset() })
```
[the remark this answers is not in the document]
jxom: Yeah, thought we need to invalidate (clear) the returned `data` if there is no longer a connected account (signer).

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
jxom: Yeah!

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
jxom: Still works for the stuff in `prepareContractWriteQuery`, however, we lose it for `config`. I figured most usages of `usePrepareContractWrite` will have `config` in the return value, so not sure if it was worth the effort to tweak this into our custom `useQuery`.

jxom:
> Should we default "transaction" hooks to a specific mode?

Thought about this. Figured it might be clearer if it were explicitly provided to avoid unintended usage. Example: if it were default to `dangerouslyUnprepared`, would this make consumers think that this usage is "okay" ? On the otherhand, if it were default to `prepared`, consumers could unintendedly "prepare" their own request without realising that the prepare hooks/actions are meant to do this.

> Do we want to type "prepared" address values (e.g. to) as 0x${string}?

Yes! Sounds great.

> Since we are removing automatic chain switching for "transaction" hooks, can we also remove from [signTypedData](https://github.com/tmm/wagmi/blob/jxom/eager/packages/core/src/actions/accounts/signTypedData.ts#L50=)?

Yep makes sense.

packages/core/src/actions/transactions/prepareSendTransaction.ts
```diff
@@ -0,0 +1,61 @@
...
+    request.gasLimit
+      ? Promise.resolve(request.gasLimit)
+      : signerOrProvider.estimateGas(request),
+  ])
+
+  return {
+    ...(chainId ? { chainId } : {}),
+    request: { ...request, gasLimit, to: to as string },
```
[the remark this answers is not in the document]
jxom: Yeah I reckon we should throw here too.

tmm:
Looks good! Can you check out `MockConnector` in `packages/react/test/utils.ts` real quick? Something with the types is not jiving.

danielduan:
Hey @jxom, I'm wondering if `useSignMessage` needs a `prepareSignMessage` as well. I'm running into issues deeplinking to Rainbow and Metamask Mobile and back when asking for a wallet signature. Thanks!

jxom:
@danielduan – I tried to reproduce a deeplink redirect using `useSignMessage`, but no success. Do you mind if you could provide a reproducible example in a CodeSandbox?

danielduan:
@jxom is there a codesandbox i can fork with the latest version set up?

the function we're using is `signMessageAsync` and it seems to do similar things `sendTransaction` but without the gas fee calculation. thanks!

EvStorM:
When I call signMessageAsync immediately after a successful login, a similar situation occurred. Note that this happened on an H5 page using a mobile browser.

PR #659, opened 2022-07-04
fix: `useContractRead` & `useContractReads` data instability

### LINKED ISSUE #648: [bug] useContractReads data instability

### Package Version

latest

### Current Behavior

useContractReads returns unstable data, equality is always lost despite underlying data being exactly the same

### Expected Behavior

In cases where the underlying data has not changed, for equality to be preserved. 

### Steps To Reproduce

repro below

https://codesandbox.io/s/small-smoke-2i7b0q?file=/src/App.js

jxom:
## Description

Fixes #648.


packages/react/src/hooks/contracts/useContractRead.ts
```diff
@@ -121,40 +121,12 @@ export function useContractRead({
...
-    chainId,
-    client,
-    contractInterface,
-    enabled,
-    functionName,
-    overrides,
-    queryKey_,
-    watch,
-  ])
+  useInvalidateOnBlock({ enabled: watch && !cacheOnBlock, queryKey: queryKey_ })
```
jxom, line 124: The old implementation was also refetching on `provider` changes, but we are already checking that in our query key (via chain id). I have refactored this to "invalidate on new block" if `watch & !cacheOnBlock` is truthy.

matthewlilley:
I'm not super familair with changesets yet, I've been meaning to dig into it a bit more for our own use, but I've seen when changesets are added that pre-release versions are immediately published. I wonder if you'd be able to set that up for wagmi. Would be super useful.

tmm:
Great idea! We can set up [Snapshot Releases](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md) to do this. There's some [work in progress](https://github.com/changesets/changesets/issues/855) that will make this even easier to set up. Looks like it's moving quickly so can prepare a PR for setting this up.

packages/react/src/hooks/utils/useInvalidateOnBlock.ts
```diff
@@ -0,0 +1,30 @@
...
+  React.useEffect(() => {
+    if (!enabled) return
+    if (blockNumber === startBlock.current) return
+    if (!startBlock.current) {
+      startBlock.current = blockNumber
+      return
+    }
+
+    queryClient.invalidateQueries(queryKey)
+  }, [blockNumber, enabled, queryClient, queryKey])
```
tmm, line 29: If `useBlockNumber`s implementation was updated slightly, we could use the `onSuccess` callback instead of a separate effect here.

packages/react/src/hooks/utils/useInvalidateOnBlock.ts
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react'
+import { QueryKey, useQueryClient } from 'react-query'
+
+import { useBlockNumber } from '../network-status'
+
+export function useInvalidateOnBlock({
```
tmm, line 6: Think it's useful to have a basic test that checks the query cache between new blocks?

Not easy to do with our current testing setup, but maybe at least we should add a `test.todo`.

tmm (commented):
Want to move the `deepEqual` implementation into `@wagmi/core`? Seems useful to have an ethers-compatible implementation live there (e.g. for dealing with things like `Result`). Also, should we expose it from `wagmi` too - now that it's a default for `isDataEqual`?

packages/react/src/hooks/utils/useInvalidateOnBlock.ts
jxom: Good idea! Updated, and now the implementation is more basic 👌

packages/react/src/hooks/network-status/useBlockNumber.ts
```diff
@@ -11,6 +11,8 @@ import { useProvider, useWebSocketProvider } from '../providers'
 import { useChainId, useQuery } from '../utils'
 
 type UseBlockNumberArgs = Partial<FetchBlockNumberArgs> & {
+  /** Function fires when a new block is created */
+  onBlock?: (blockNumber: number) => void
```
tmm, line 15: pretty awesome

tmm (approved):

PR #663, opened 2022-07-05
docs: fix WalletConnect deep linking on iOS in SIWE example

markdalgleish:
## Description

For the same reasoning behind #658, the SIWE example currently breaks deep linking for WalletConnect users on iOS. This is because the API call to fetch the nonce happens before the message signing process, causing the browser to perform a regular page navigation rather than following the deep link since it's too far removed from the original user action.

The fix for this is to pre-fetch the nonce before allowing the user to sign in so that the message signing happens immediately on click. I've updated the working demo and the example code, doing my best to try and keep the code as similar to the original example.

Note that I've also opted to remove the documentation for combining the "Connect Wallet" and "Sign In" steps since, in light of the deep linking issues we've found, this should be regarded as an anti-pattern. Instead I've replaced it with a warning that these steps need to remain separate if you want to ensure that WalletConnect on iOS works correctly.

## Additional Information

Your ENS/address: markdalgleish.eth


jxom (approved):
👑

tmm (approved):
ty ser

PR #677, opened 2022-07-06
fix: parse tuples correctly in `parseContractResult`

jxom:
We weren't parsing complex tuples correctly. Now we are :D


tmm (approved):
No need to do this here, but maybe we should move `parseContractResult` to `@wagmi/core` too!

tmm (approved):

PR #721, opened 2022-07-17
refactor: clean up

tmm:
## Description

- Stay connected to existing connector when connect action fails
- Make chains arrays tree shakeable
- Make injected connector name override work with SSR
- Fix client auto connect loading status
- Add name to fetch token and refactor to use multicall

## Additional Information

Your ENS/address: awkweb.eth

### DIFF
--- a/packages/core/src/actions/tokens/fetchToken.ts
+++ b/packages/core/src/actions/tokens/fetchToken.ts
@@ -16,6 +16,7 @@ export type FetchTokenArgs = {
 export type FetchTokenResult = {
   address: string
   decimals: number
+  name: string
   symbol: string
   totalSupply: {
     formatted: string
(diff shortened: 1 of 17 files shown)


packages/core/src/actions/tokens/fetchToken.ts
```diff
@@ -28,21 +29,31 @@ export async function fetchToken({
...
-    contract.decimals(),
-    contract.totalSupply(),
-  ])
-  const token = {
+  const erc20Config = {
+    addressOrName: address,
+    contractInterface: erc20ABI,
+    chainId,
+  }
+  const [decimals, name, symbol, totalSupply] = await readContracts<
```
tmm, line 37: switch to multicall

packages/core/src/index.ts
```diff
@@ -91,7 +91,11 @@ export { createClient, Client } from './client'
 export type { ClientConfig } from './client'
 
 export { Connector, InjectedConnector } from './connectors'
-export type { ConnectorData, ConnectorEvents } from './connectors'
+export type {
+  ConnectorData,
+  ConnectorEvents,
+  InjectedConnectorOptions,
```
tmm, line 97: Makes it easier for folks to build their own injected connector

packages/core/src/types/index.ts
```diff
@@ -65,139 +65,139 @@ export type WebSocketProvider = providers.WebSocketProvider & {
...
+   * EIP-3326: Wallet Switch Ethereum Chain RPC Method
+   * https://eips.ethereum.org/EIPS/eip-3326
+   */
+  request(args: {
+    method: 'wallet_switchEthereumChain'
+    params: [{ chainId: string }]
+  }): Promise<null>
+}
+
+declare global {
```
tmm, line 200: this is the only thing that needs to be global here

PR #722, opened 2022-07-18
fix: polling interval for multiple providers

### LINKED ISSUE #650: [bug] pollingInterval not working on jsonRpcProvider

### Package Version

0.5.5

### Current Behavior

`pollingInterval` on `jsonRpcProvider` is not being respected. Seems that no matter what i set, it always uses 4000.

### Expected Behavior

Whatever number is passed in for `pollingInterval` should be used.

### Steps To Reproduce

Code:
```
import { jsonRpcProvider } from 'wagmi/providers/jsonRpc'

jsonRpcProvider({
    rpc: (chain) => {
      return {
        http: 'some rpc url',
      }
    },
    pollingInterval: 10000,
    static: false,
  }),
```

_No response_

### Anything else?

It works fine for `alchemyProvider`, not sure why it works there but not here. Or if I'm doing something wrong.

My `ethers` version is 5.6.9

jxom:
Fixes #650


packages/core/src/utils/configureChains.ts
```diff
@@ -26,7 +29,12 @@ export function configureChains<
 >(
   defaultChains: TChain[],
   providers: ChainProviderFn<TProvider, TWebSocketProvider, TChain>[],
-  { minQuorum = 1, targetQuorum = 1, stallTimeout }: ConfigureChainsConfig = {},
+  {
+    minQuorum = 1,
+    pollingInterval = 4000,
```
tmm, line 34: 😁

```suggestion
    pollingInterval = 4_000,
```

tmm (approved):

PR #734, opened 2022-07-19
fix: make chain aware providers great again

### LINKED ISSUE #732: [bug] useBalance doesn't update if the input chainId changes

### Package Version

0.5.10

### Current Behavior

There's some issues with using multiple useBalance with different chainId, it's not invalidating in expected way, and some odd behavior overall.

### Expected Behavior

Show correct balances for chain

### Steps To Reproduce

see repro

https://codesandbox.io/s/competent-sinoussi-q0ry6b?file=/src/App.js

### Anything else?

_No response_

jxom:
Fixes #732

- Fixes an issue where the `useProvider` hook would not update if `chainId` changes
- Refactor `useProvider` to use `useSyncExternalStoreWithSelector` to better support React 18
- Added test case to validate fixed behavior

### DIFF
--- a/.changeset/gentle-squids-switch.md
+++ b/.changeset/gentle-squids-switch.md
@@ -1,0 +1,5 @@
+---
+'wagmi': patch
+---
+
+Fix issue where `useProvider` would not update when `chainId` config changes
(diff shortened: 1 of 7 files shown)


tmm (commented):
Should we do this for `useWebSocketProvider` too?

jxom:
> Should we do this for `useWebSocketProvider` too?

Good point!

PR #751, opened 2022-07-22
test: switch to anvil

tmm:
## Description

Switch from Hardhat to Anvil

## Additional Information

Your ENS/address: awkweb.eth


packages/core/src/actions/transactions/prepareSendTransaction.ts
```diff
@@ -43,7 +43,7 @@ export type PrepareSendTransactionResult = {
 export async function prepareSendTransaction({
   chainId,
   request,
-  signerOrProvider = getProvider(),
+  signerOrProvider = getProvider({ chainId }),
```
tmm, line 46: think this was missing before

PR #752, opened 2022-07-23
test: switch to vitest

tmm:
## Description

Switch to [Vitest](https://vitest.dev/)

## Additional Information

Your ENS/address: awkweb.eth


.github/workflows/main.yml
```diff
@@ -162,6 +162,16 @@ jobs:
...
         run: pnpm test:coverage
         env:
           REACT_VERSION: ${{ matrix.react }}
+      - name: Test
+        if: matrix.react == '17'
+        run: |
+          cd packages/react
+          pnpm add -D react@17.0.2 react-dom@17.0.2
+          cd ../..
+          pnpm test:run react
```
tmm, line 175: Vitest doesn't mock globals in the same way as Jest. So instead of using mocking React in a setup file, we'll install the target React 17 version here.

packages/react/src/hooks/utils/useSyncExternalStoreWithTracked.test.ts
```diff
@@ -37,23 +38,26 @@ function useExternalStore(
...
-      return <div>{baz}</div>
-    }
+    renderHook(() => {
+      const { gm } = useExternalStore(externalStore, (state) => {
+        renders.push(state)
+      })
 
-    render(wrapper({ children: <App /> }), document.createElement('div'))
+      return gm
+    })
```
tmm, line 60: Switch from rendering React node to hook abstraction

PR #759, opened 2022-07-25
feat: useTransaction

tmm:
## Description

Adds `useTransaction` hook

## Additional Information

Your ENS/address: awkweb.eth


packages/core/src/actions/transactions/fetchTransaction.ts
```diff
@@ -0,0 +1,32 @@
...
+import { getProvider } from '../providers'
+
+export type FetchTransactionArgs = {
+  /** Chain ID used to validate if the signer is connected to the target chain */
+  chainId?: number
+  /** Transaction hash */
+  hash: Hash
+}
+
+export type FetchTransactionResult = providers.TransactionResponse
```
tmm, line 13: Returns unmodified [`TransactionResponse`](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionResponse) here, but maybe we start constructing our own objects instead of returning ethers responses directly?

[`TransactionReceipt`](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionReceipt) also has some useful properties in it (e.g. `logs`).

packages/core/src/types/index.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from '../constants'
 
 export type Address = `0x${string}`
+export type Hash = `0x${string}`
```
tmm, line 11: `Address` and `Hash` are the same. Maybe this is fine for naming.

jxom (approved):

PR #764, opened 2022-07-26
feat: prefer "reckless" > "dangerous"

jxom:
## Description

For Prepare Hooks, the term "dangerous" sounds scary, but the dangerous usage isn't really "dangerous" at all. The term "reckless" fits better.

don't be so reckless.

https://www.youtube.com/watch?v=JIrUqsB-0vw


markdalgleish:
API design is my passion.

tmm (approved):
Good idea

PR #767, opened 2022-07-26
add Optimism Goerli chain

klyap:
## Description

Adding Optimism Goerli as a supported chain

## Additional Information

Your ENS/address: kerls.eth

### DIFF
--- a/packages/core/src/constants/blockExplorers.ts
+++ b/packages/core/src/constants/blockExplorers.ts
@@ -12,6 +12,7 @@ type EtherscanChains = Extract<
   | 'kovan'
   | 'optimism'
   | 'optimismKovan'
+  | 'optimismGoerli'
   | 'polygon'
   | 'polygonMumbai'
   | 'arbitrum'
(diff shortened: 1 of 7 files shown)


packages/core/src/constants/blockExplorers.ts
```diff
@@ -46,6 +47,10 @@ export const etherscanBlockExplorers: Record<EtherscanChains, BlockExplorer> = {
     name: 'Etherscan',
     url: 'https://kovan-optimistic.etherscan.io',
   },
+  optimismGoerli: {
+    name: 'Etherscan',
+    url: 'https://l2-explorer.surge.sh/',
```
tmm, line 52: This isn't an official Etherscan block explorer so let's remove this from here. Can inline this URL on the `blockExplorers.default` field for now.

tmm (changes requested):
Small change, then good

packages/core/src/constants/blockExplorers.ts
klyap: Should I just use 'https://goerli-optimistic.etherscan.io' in the `url` field here instead? Or pass an empty string?

Also, what do you mean by inlining in `blockExplorers.default`? Like inline comment?

tmm: Etherscan doesn't have a block explorer out yet for Goerli Optimism so no need to add to the Etherscan block explorer object.

tmm (approved):

PR #828, opened 2022-08-10
WIP: support dynamically resolving initial chain ID

markdalgleish:
## Description

This is still a WIP, but I wanted to open a PR early to get feedback on it before I polish too much.

This expands the type for the `chainId` option on the `connect` function to accept the following function type:
```ts
(context: { walletChainId?: number; chains: Array<Chain> }) =>
  number | undefined
```

The main use case for this (which I'll be using it for) is to avoid switching chains while connecting if the wallet is already on a supported chain:

```tsx
connect({
  connector,
  chainId: ({ walletChainId, chains }) =>
    chains.find((chain) => chain.id === walletChainId)?.id ?? chains[0]?.id,
});
```

In the case of WalletConnect, `walletChainId` will be undefined, but for MetaMask it will be the ID that it's already connected to.

I'm keen to get your thoughts on the idea and implementation. You could potentially provider a higher level API than this but I wanted to avoid baking any opinions in — hence why I inverted control of the logic to the consumer.

## Additional Information

Your ENS/address: markdalgleish.eth

### DIFF
--- a/examples/_dev/src/components/Connect.tsx
+++ b/examples/_dev/src/components/Connect.tsx
@@ -8,15 +8,33 @@ export const Connect = () => {
   const { connector, isReconnecting } = useAccount()
   const { connect, connectors, isLoading, error, pendingConnector } =
     useConnect()
+  const [ensureSupportedChain, setEnsureSupportedChain] = React.useState(false)
 
   return (
     <div>
+      <label style={{ display: 'flex', marginBottom: 4 }}>
+        <input
+          type="checkbox"
+          checked={ensureSupportedChain}
+          onChange={(e) => setEnsureSupportedChain(e.currentTarget.checked)}
+        />
+        Switch to supported chain on connect
+      </label>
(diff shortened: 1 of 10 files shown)


midnight-madman:
I love this <3 thanks for opening the PR. 
Ran into this problem today and want to avoid forcing users to switch to different chain even though the current one is support too.

jxom:
Looping from internal convo:

We could do this without any wagmi changes.

```tsx
const { chains, getChainId } = connector;
const walletChainId = await getChainId()
const chainId = chains.find((chain) => chain.id === walletChainId)?.id ??
  chains[0]?.id
connect({
  connector: x,
  chainId,
})
```

Wonder if we can document this in the docs.

markdalgleish:
@jxom thanks for highlighting this API! I didn't know you could do this. This makes it super simple 🎉

Closing the PR since we can handle it on our end now.

PR #875, opened 2022-08-23
chore: update to ethers v5.7

jxom:
## Description

Updating `ethers` dev dependency & example dependencies to v5.7


package.json
```diff
@@ -56,7 +56,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-react-hooks": "^4.5.0",
     "eslint-plugin-testing-library": "^5.5.1",
-    "ethers": "^5.6.5",
+    "ethers": "^5.7.0",
```
tmm, line 59: We should bump `@ethersproject/providers` too.

tmm (approved):

PR #889, opened 2022-08-26
refactor: make multicall & readContracts more robust

jxom:
## Description

There have been complaints of `multicall`/`readContracts` returning `null` in some entries. This is most likely because the contract has reverted, or the input data for the contract function is invalid. This PR aims to make `multicall` & `readContracts` more robust, and also warn users (via `console.warn`) _why_ some entries are `null` (most likely arising from a contract error, or an invalid contract call).


packages/core/src/actions/contracts/multicall.ts
```diff
@@ -70,18 +72,27 @@ export async function multicall<Data extends any[] = Result[]>({
...
       })
       const params = Array.isArray(args) ? args : args ? [args] : []
-      const callData = contract.interface.encodeFunctionData(
-        functionName,
-        params,
-      )
-      if (!contract[functionName])
-        console.warn(
-          `"${functionName}" is not in the interface for contract "${addressOrName}"`,
+      try {
```
jxom, line 75: We weren't guarding against potential errors when encoding function data (e.g. a user may provide invalid `params`, causing `encodeFunctionData` to blow up), and this caused `multicall` to always return `[null]`.

packages/core/src/actions/contracts/multicall.ts
```diff
@@ -70,18 +72,27 @@ export async function multicall<Data extends any[] = Result[]>({
...
+          target: addressOrName,
+          allowFailure,
+          callData,
+        }
+      } catch (err) {
+        if (!allowFailure) throw err
+        return {
+          target: addressOrName,
+          allowFailure,
+          callData: '0x',
```
jxom, line 94: If we catch an error from encoding function data, we will make the calldata for that call nullish.

packages/core/src/actions/contracts/readContracts.ts
```diff
@@ -42,17 +46,17 @@ export async function readContracts<Data extends any[] = Result[]>({
         [chainId]: [...(contracts[chainId] || []), contract],
       }
     }, {})
-    const promises = Object.entries(contractsByChainId).map(
-      ([chainId, contracts]) =>
+    const promises = () =>
```
jxom, line 49: We were originally invoking the promises instantly, instead of deferring it to the `Promise.all`.

examples/_dev/next.config.js
```diff
@@ -2,5 +2,8 @@ const withPreconstruct = require('@preconstruct/next')
 
 /** @type {import('next').NextConfig} */
 module.exports = withPreconstruct({
+  eslint: {
+    ignoreDuringBuilds: true,
+  },
```
tmm, line 7: Was this slowing down the build?

examples/_dev/package.json
```diff
@@ -13,7 +13,7 @@
     "next": "^12.2.0",
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
-    "wagmi": "^0.6.4"
+    "wagmi": "workspace:*"
```
tmm, line 16: Wasn't this working before with `preconstruct dev`?

packages/core/src/errors.ts
```diff
@@ -145,27 +145,112 @@ export class ContractMethodNoResultError extends Error {
...
+  }: {
+    addressOrName: string
+    args: any
+    chainId: number
+    functionName: string
+    errorMessage: string
+  }) {
+    super(
+      [
+        'Contract read threw an error.',
```
tmm, line 238: wdyt about something more generic (e.g. `'Contract method reverted with error'`)? That was this is less tied to read methods (unless we want it to be).

packages/core/src/actions/contracts/readContracts.ts
```diff
@@ -63,18 +67,31 @@ export async function readContracts<Data extends any[] = Result[]>({
...
-      ) as Data
+      return (await Promise.allSettled(promises())).map((result, i) => {
+        if (result.status === 'fulfilled') return result.value
+        const { addressOrName, functionName, chainId, args } = contracts[
+          i
+        ] as ReadContractsContract
+        const error = new ContractRevertedError({
+          addressOrName,
+          functionName,
+          chainId: chainId ?? chain.mainnet.id,
```
tmm, line 87: Let's import `mainnet` here instead of getting it off of `chain` since `chain` pulls in a lot more than necessary.

packages/core/src/actions/contracts/readContracts.test.ts
```diff
@@ -29,12 +29,17 @@ const contracts: ReadContractsConfig['contracts'] = [
   },
 ]
 
+let warnMessages: string[] = []
+const warn = vi
+  .spyOn(console, 'warn')
+  .mockImplementation((message) => warnMessages.push(message))
```
tmm, line 35: pretty awesome tests 🙇‍♂️

tmm (approved):

examples/_dev/next.config.js
jxom: Can't remember. Might revert this.

examples/_dev/package.json
jxom: This was to fix issues with linking `@wagmi/core`, but since we merged the lockfile fix, we don't need this anymore!

packages/core/src/errors.ts
jxom: Sounds good!

PR #909, opened 2022-09-02
fix: add `foundryMainnet` chain for tests

jxom:
## Description

Seems like https://github.com/foundry-rs/foundry/pull/2995 caused our tests to expectedly break. This PR makes our test suite more aligned to mainnet by using the mainnet chain id (`1`) instead of `31_337`.


.changeset/nervous-lizards-protect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@wagmi/core': patch
+'wagmi': patch
+---
+
+Add `foundryMainnet` chain to align tests more to mainnet
```
tmm, line 6: we're not changing anything public so probably don't need this

tmm (approved):

PR #933, opened 2022-09-12
feat: pure esm

jxom:
This PR converts `wagmi` from bundling both CJS + ESM, to bundling pure ESM. 

I have published the packages to NPM as a snapshot release under the tag `esm-only` (`npm i wagmi@esm-only`).

This will probably affect tooling that is sensitive to importing ESM such as Jest and Remix. We might need to create some additional guides on how to use wagmi with these tools as we move to pure ESM.


examples/remix/remix.config.js
```diff
@@ -8,4 +8,5 @@ module.exports = {
   serverBuildDirectory: 'build',
   devServerPort: 8002,
   ignoredRouteFiles: ['.*'],
+  serverDependenciesToBundle: [/^@?wagmi.*/, 'zustand/vanilla'],
```
jxom, line 11: https://remix.run/docs/en/v1/pages/gotchas#importing-esm-packages

patches/@preconstruct__cli@2.2.1.patch
```diff
@@ -0,0 +1,949 @@
+# Source: https://github.com/preconstruct/preconstruct/compare/main...jxom:preconstruct:tmm/esm-support
```
jxom, line 1: Source for the patch.

packages/core/chains/package.json
```diff
@@ -1,4 +1,4 @@
 {
-  "main": "dist/wagmi-core-chains.cjs.js",
-  "module": "dist/wagmi-core-chains.esm.js"
+  "main": "dist/wagmi-core-chains.esm.js",
```
jxom, line 2: Ideally, we don't even need these `package.json`'s, but we may need them for a couple of reasons:

- Some tooling (Jest) don't respect the `exports` field, and
- Setting `moduleResolution` to anything other than `node16` or `nodenext` in `tsconfig.json` breaks type resolution. However, if we do set `moduleResolution` to one of those values, we need to append the `.js` extension to every relative import (yuck). There are long running issues for this [here](https://github.com/microsoft/TypeScript/issues/50152) and [here](https://github.com/microsoft/TypeScript/issues/49083).

packages/core/src/actions/accounts/fetchBalance.ts
```diff
@@ -1,5 +1,5 @@
-import { BigNumber, logger } from 'ethers/lib/ethers'
-import { Logger, formatUnits, isAddress } from 'ethers/lib/utils'
+import { BigNumber, logger } from 'ethers/lib/ethers.js'
```
jxom, line 1: Is there a reason why we aren't using `ethers` here? Should I move to that?

tmm:
opening up PR with different approach shortly

PR #936, opened 2022-09-12
feat: custom logger

jxom:
## Description

This PR adds the ability to pass a custom logger to wagmi. Particularly useful if you wish to suppress console warnings.

Also removed the old `warn` util as it was unused.

[Read the changesets for examples](https://github.com/wagmi-dev/wagmi/pull/936/files#diff-68b30f2d9b5fc06985c3d21ef068b153e0a21de897253ad44ad2df07f022e85cR1)


packages/core/src/utils/logger.ts
```diff
@@ -0,0 +1,5 @@
+import { client } from '../client'
+
+export function logWarning(message: string) {
+  client?.config.logger?.warn?.(message)
+}
```
tmm, line 5: Let's use `getClient` and call it `logWarn`?

```suggestion
import { getClient } from '../client'

export function logWarn(message: string) {
  getClient()?.config.logger?.warn?.(message)
}
```

tmm (approved):

packages/core/src/utils/logger.ts
jxom: sg!

tmm:
Should add `logger` to [`Client` docs](https://wagmi.sh/docs/client) too.

Issue #928, opened 2022-09-09
bug: usePrepareContractWrite chainId validation not working


Fugashu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.6.4

### Current Behavior

I'm using the usePrepareContractWrite function in order to mint a NFT on Ethereum Mainnet.

In my MVP I'm calling the function as described [here](https://wagmi.sh/examples/contract-write):

`const {
    config,
    error: prepareError,
    isError: isPrepareError,
  } = usePrepareContractWrite({
    addressOrName: "0xaf0326d92b97df1221759476b072abfd8084f9be",
    contractInterface: ["function mint()"],
    functionName: "mint",
    chainId: 1,
  });

  const { data, error, isError, write } = useContractWrite(config);
`

When the user switches chains via Metamask extension, the `write` function invokes a transaction of 0 ETH to the specified contract address. The same behaviour occurs in the Docs page, while sometimes the 'Mint' button is disabled, occasionally when hopping chains it still tries to send a transaction when it's on the wrong chain.

### Expected Behavior

An actual error whenever the current user is not on the specified chain `chainId` . The `write` function should not execute an empty transaction.

### Anything else?

![Screenshot_20220909_223718](https://user-images.githubusercontent.com/25103129/189439019-517cf983-9a13-4063-a281-fe3547814bb4.png)

ivelin:
I believe this is related to a previously reported bug which was closed without a fix:
https://github.com/wagmi-dev/wagmi/issues/827

tmm:
@ivelin #827 was closed because it needs a minimal reproducible example (see https://github.com/wagmi-dev/wagmi/issues/827#issuecomment-1210061779). If you believe this issue is related, it could be helpful to provide one.

Fugashu:
```
import * as React from "react";
import {
  usePrepareContractWrite,
  useContractWrite,
  useWaitForTransaction,
} from "wagmi";

export function MintNFT() {
  const { config } = usePrepareContractWrite({
    addressOrName: "0xaf0326d92b97df1221759476b072abfd8084f9be",
    contractInterface: ["function mint()"],
    functionName: "mint",
    chainId: 1,
  });
  const { data, write } = useContractWrite(config);

  const { isLoading, isSuccess } = useWaitForTransaction({
    hash: data?.hash,
  });

  return (
    <div>
      <button disabled={!write || isLoading} onClick={() => write?.()}>
[...]

PR #940, opened 2022-09-14
fix: make chainId validation on Prepare Hooks more robust

jxom:
## Description

This PR aims to solve #928 by making the `chainId` config option on the Prepare Hooks more robust. Previously, the `chainId` validation on the prepare hooks didn't work well, particularly when switching networks. 

I have also slightly modified the behavior of `usePrepareSendTransacton` to only execute when the end user is connected to a wallet, instead of executing regardless of connection. This reaches parity with the behavior of `usePrepareContractWrite`. [More details are in the changesets.](https://github.com/wagmi-dev/wagmi/pull/940/files)

### DIFF
--- a/.changeset/rotten-buttons-protect.md
+++ b/.changeset/rotten-buttons-protect.md
@@ -2,10 +2,6 @@
 'wagmi': minor
 ---
 
----
-'wagmi': minor
----
-
 **Breaking**: `usePrepareSendTransaction` now throws when a `chainId` is specified and the end-user is on a different chain id (the wrong network).
 
 If you wish to defer this check until the click handler is pressed, you can place `chainId` in `useContractWrite` instead:
--- a/.changeset/rotten-buttons-protect.md
+++ b/.changeset/rotten-buttons-protect.md
@@ -1,0 +1,30 @@
+---
+'wagmi': minor
+---
+
+---
+'wagmi': minor
+---
+
+**Breaking**: `usePrepareSendTransaction` now throws when a `chainId` is specified and the end-user is on a different chain id (the wrong network).
+
+If you wish to defer this check until the click handler is pressed, you can place `chainId` in `useContractWrite` instead:
+
+```diff
+import { usePrepareSendTransaction, useContractWrite } from 'wagmi'
+import { optimism } from 'wagmi/chains'
+
+// ...
+
+const { config } = usePrepareSendTransaction({
+  request: {
+    to: 'moxey.eth',
+    value: parseEther('1'),
+  },
+})
+const { sendTransaction } = useSendTransaction({ 
+  ...config,
++ chainId: optimism.id
+})
+
+```
--- a/packages/core/src/actions/contracts/prepareWriteContract.ts
+++ b/packages/core/src/actions/contracts/prepareWriteContract.ts
@@ -65,7 +65,7 @@ export async function prepareWriteContract<
   overrides,
   signer: signer_,
 }: PrepareWriteContractConfig): Promise<PrepareWriteContractResult<TSigner>> {
-  const signer = signer_ ?? (await fetchSigner())
+  const signer = signer_ ?? (await fetchSigner({ chainId }))
   if (!signer) throw new ConnectorNotFoundError()
 
   const { chain: activeChain, chains } = getNetwork()
--- a/packages/core/src/actions/contracts/prepareWriteContract.ts
+++ b/packages/core/src/actions/contracts/prepareWriteContract.ts
@@ -5,12 +5,13 @@ import {
 } from 'ethers/lib/ethers'
 
 import {
+  ChainMismatchError,
   ConnectorNotFoundError,
   ContractMethodDoesNotExistError,
 } from '../../errors'
 import { Address, Signer } from '../../types'
 import { minimizeContractInterface } from '../../utils'
-import { fetchSigner } from '../accounts'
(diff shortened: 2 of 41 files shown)


packages/core/src/actions/contracts/prepareWriteContract.ts
```diff
@@ -64,9 +65,21 @@ export async function prepareWriteContract<
...
+  const activeChainId = activeChain?.id
+  if (chainId && chainId !== activeChainId) {
+    throw new ChainMismatchError({
+      activeChain:
+        chains.find((x) => x.id === activeChainId)?.name ??
+        `Chain ${activeChainId}`,
+      targetChain:
+        chains.find((x) => x.id === chainId)?.name ?? `Chain ${chainId}`,
+    })
+  }
```
tmm, line 81: Since we have this logic in a multiple places, I wonder if it makes sense to extract it out into an `assertActiveChain` utility?

packages/react/src/client.ts
```diff
@@ -57,7 +56,7 @@ export function createClient<
       dehydrateOptions: {
         shouldDehydrateQuery: (query) =>
           query.cacheTime !== 0 &&
-          query.queryHash !== JSON.stringify(signerQueryKey()),
+          (query.queryKey[0] as { persist?: boolean }).persist !== false,
```
tmm, line 59: nice abstraction! a quick comment here that `persist` can be added to query keys to control the behavior could be nice.

packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -7,19 +7,25 @@ import {
...
+import { QueryConfig, QueryFunctionArgs } from '../../types'
 import { useQuery } from '../utils'
 
 export type UseSignerConfig = Omit<
   QueryConfig<FetchSignerResult, Error>,
   'cacheTime' | 'staleTime' | 'enabled'
->
+> & {
+  chainId?: number
+}
```
tmm, line 18: Should this use `FetchSignerArgs` instead?

packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -7,19 +7,25 @@ import {
...
 
-export const queryKey = () => [{ entity: 'signer' }] as const
+export const queryKey = ({ chainId }: Partial<UseSignerConfig>) =>
```
tmm, line 20: Could use `FetchSignerArgs` instead of `UseSignerConfig`

packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -7,19 +7,25 @@ import {
...
+  [{ entity: 'signer', chainId, persist: false }] as const
 
-const queryFn = <TSigner extends Signer>() => fetchSigner<TSigner>()
+const queryFn = <TSigner extends Signer>({
+  queryKey: [{ chainId }],
+}: QueryFunctionArgs<typeof queryKey>) => fetchSigner<TSigner>({ chainId })
 
 export function useSigner<TSigner extends Signer>({
+  chainId,
```
tmm, line 28: Should we use `const chainId = useChainId({ chainId: chainId_ })` instead of passing this `chainId` directly through?

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
```diff
@@ -91,18 +98,21 @@ export function usePrepareContractWrite({
...
     queryKey(
       {
         addressOrName,
         functionName,
+        chainId,
         args,
         overrides,
       },
-      { chainId, signerAddress: signer?._address },
+      { activeChainId, signerAddress: signer?._address },
```
tmm, line 115: What's the difference between `chainId` and `activeChainId`? Could they collapse into one with `const activeChainId = useChainId({ chainId })`?

jxom: We need the `activeChainId` to cache bust.

packages/react/src/hooks/accounts/useSigner.ts
jxom: Yep, sounds good!

packages/react/src/client.ts
jxom: Yep! sg!

packages/core/src/actions/contracts/prepareWriteContract.ts
jxom: Was thinking the same. Can do!

packages/react/src/hooks/accounts/useSigner.ts
jxom: Good catch.

packages/core/src/index.ts
```diff
@@ -167,6 +168,7 @@ export type {
 } from './types'
 
 export {
+  assertActiveChain,
```
tmm, line 171: Maybe we don't expose this publicly?

tmm (approved):

.changeset/rotten-buttons-protect.md
```diff
@@ -0,0 +1,30 @@
+---
+'wagmi': minor
+---
+
+---
+'wagmi': minor
+---
```
tmm, line 7: Looks like this is a dupe

jxom: What the

PR #963, opened 2022-09-18
feat: Isolate wagmi query context from react-query default context

aj-may:
## Description

Isolating the react-query context used by wagmi from the default context used by react-query.

## Additional Information

For applications that are using both the wagmi and react-query libraries, both will typically end up using the same queryClient.  This means that useQuery calls from within my application may inadvertently use the wagmi client, and the defaults it sets (if the wagmiConfig provider is the innermost provider), or wagmi calls to useQuery will end up using my applications queryClient and the defaults I have set.

This actually can cause a bit of a debugging nightmare as the `persister` set in the wagmi queryClient doesn't play nice with SSR frameworks like Next.js.

This lack of isolation was also discussed here: https://github.com/wagmi-dev/wagmi/discussions/661

This is my first contribution to this library so I'm not super familiar with the dev set-up.  I ran the tests before and after my changes and got the same results, but I did have failing tests.

I also want to call out, I did overwrite `options.context` in `useBaseQuery`.  I don't think this should have any negative consequences as I don't see context being passed as an option anywhere `useBaseQuery` was called.  If this assumption is not true please let me know, happy to change the approach.

Your ENS/address: ajmay.eth


aj-may:
@tmm @jxom Thoughts on this PR?  Any feedback?

packages/react/src/hooks/accounts/useConnect.ts
```diff
@@ -45,6 +45,7 @@ export function useConnect({
     onMutate,
     onSettled,
     onSuccess,
+    context: queryClientContext,
```
jxom, line 48: Can we have our own `useMutation` implementation (under `utils/query`) so we aren't having to manually pass down `context` on every usage?

packages/react/src/hooks/utils/query/useBaseQuery.ts
```diff
@@ -28,7 +30,7 @@ export function useBaseQuery<
   >,
   Observer: typeof QueryObserver,
 ) {
-  const queryClient = useQueryClient({ context: options.context })
+  const queryClient = useQueryClient({ context: queryClientContext })
```
jxom, line 33: Can we instead leave this as is (would rather not touch this file as it's forked from RQ), and instead place this inside `utils/query/useQuery` and `utils/query/useInfiniteQuery`?

packages/react/src/hooks/utils/useInvalidateOnBlock.ts
```diff
@@ -9,7 +11,7 @@ export function useInvalidateOnBlock({
   enabled?: boolean
   queryKey: QueryKey
 }) {
-  const queryClient = useQueryClient()
+  const queryClient = useQueryClient({ context: queryClientContext })
```
jxom, line 14: Can we have our own `useQueryClient` in `utils/query` that uses `context` so we don't have to manually pass it on each usage?

jxom (changes requested):
This definitely makes sense to have so we can isolate RQ contexts! Just have a few comments.

aj-may:
@jxom For some context, why is `useQuery`, `useBaseQuery`, etc. forked currently?

jxom:
@aj-may – We needed our own `useBaseQuery` as we wanted to modify the return value API of RQ without losing the tracked values behaviour.

jxom:
If we can also get a changeset happening, that would be amazing! 

`pnpm changeset`, and then make this one a `patch` 🙏

packages/react/src/hooks/utils/query/useMutation.ts
```diff
@@ -0,0 +1,78 @@
...
+    | MutationFunction<TData, TVariables>
+    | UseMutationOptions<TData, TError, TVariables, TContext>,
+  arg2?:
+    | MutationFunction<TData, TVariables>
+    | UseMutationOptions<TData, TError, TVariables, TContext>,
+  arg3?: UseMutationOptions<TData, TError, TVariables, TContext>,
+): UseMutationResult<TData, TError, TVariables, TContext> {
+  const options = parseMutationArgs(arg1, arg2, arg3)
+  return rqUseMutation({ context, ...options })
+}
```
tmm, line 78: We should export `useMutation` from `wagmi` too (just like `useQuery`).

packages/react/src/hooks/accounts/useConnect.ts
```diff
@@ -1,9 +1,9 @@
-import { useMutation } from '@tanstack/react-query'
 import { ConnectArgs, ConnectResult, connect } from '@wagmi/core'
 import * as React from 'react'
 
 import { useClient } from '../../context'
 import { MutationConfig } from '../../types'
+import { useMutation } from '../utils/query'
```
tmm, line 6: Let's export `useMutation` from `src/hooks/utils` so we can import from `../utils` (like we do for `useQuery`).

packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -10,6 +9,7 @@ import * as React from 'react'
 
 import { QueryConfig, QueryFunctionArgs } from '../../types'
 import { useChainId, useQuery } from '../utils'
+import { useQueryClient } from '../utils/query'
```
tmm, line 12: Let's export `useQueryClient` from `src/hooks/utils` too.

packages/react/src/hooks/utils/query/index.ts
```diff
@@ -1,3 +1,5 @@
 export { useBaseQuery } from './useBaseQuery'
 export { useInfiniteQuery } from './useInfiniteQuery'
 export { useQuery } from './useQuery'
+export { useMutation } from './useMutation'
+export { useQueryClient } from './useQueryClient'
```
tmm, line 5: We should have an eslint rule for this, but we try to alphabetize exports when possible.

```suggestion
export { useBaseQuery } from './useBaseQuery'
export { useInfiniteQuery } from './useInfiniteQuery'
export { useMutation } from './useMutation'
export { useQuery } from './useQuery'
export { useQueryClient } from './useQueryClient'
```

packages/react/src/hooks/utils/query/useQueryClient.ts
```diff
@@ -0,0 +1,5 @@
+import { useQueryClient as rqUseQueryClient } from '@tanstack/react-query'
```
tmm, line 1: Kinda prefer aliasing to `RQ_useQueryClient`. @jxom what do you think?

tmm (changes requested):
Some thoughts. Let me know if you have any questions. Appreciate you working on this!

packages/react/src/hooks/utils/query/useMutation.ts
aj-may: I'm not sure I understand this one.  Can you provide some more information on this?

tmm:
```ts
// packages/react/src/index.ts

export {
  ...
  useMutation,
  ...
} from './hooks'
```

aj-may: Ahh! yes.  Makes sense.

aj-may: `useQueryClient` also?  If the rational is for an "escape hatch" then I think that makes sense.

packages/react/src/hooks/utils/query/useQueryClient.ts
aj-may: I don't mind changing it.  Personally I prefer sticking to camelCase and not mixing.

aj-may:
@tmm updated based on comments.  Take another look when you can

packages/react/src/hooks/utils/query/useQueryClient.ts
jxom: Could also do `useQueryClient as useQueryClient_` – that is the pattern we use for some dupe import stuff atm.

jxom (approved):
lgtm! looks like tests aren't passing b/c of this branch being on another remote, however, i triggered this branch on our remote and they are passing! https://github.com/wagmi-dev/wagmi/actions/runs/3125668497

PR #965, opened 2022-09-19
fix: block listeners for L2 chains

jxom:
## Description

This PR fixes 3 issues with block events (via `watch`):

- Throttle the behaviour of the ethers `"block"` event (this gets really aggressive for L2 chains),
- Ensure that the `watch` option respects the given `chainId`, and not the current active chain id.
- `useProvider` & `getProvider` were not returning referentially equal providers, causing multiple event emitters to be instantiated.

### DIFF
--- a/packages/core/src/actions/providers/getProvider.ts
+++ b/packages/core/src/actions/providers/getProvider.ts
@@ -8,18 +8,9 @@ export type GetProviderArgs = {
 
 export type GetProviderResult<TProvider extends Provider = Provider> = TProvider
 
-const providers = new Map()
-
 export function getProvider<TProvider extends Provider = Provider>({
   chainId,
 }: GetProviderArgs = {}): GetProviderResult<TProvider> {
   const client = getClient<TProvider>()
-  if (chainId && typeof client.config.provider === 'function') {
-    if (providers.get(chainId)) return providers.get(chainId)
-
-    const provider = client.config.provider({ chainId })
-    providers.set(chainId, provider)
-    return provider
-  }
-  return client.provider
+  return client.getProvider({ chainId }) || client.provider
 }
--- a/packages/core/src/actions/providers/getProvider.ts
+++ b/packages/core/src/actions/providers/getProvider.ts
@@ -12,5 +12,6 @@ export function getProvider<TProvider extends Provider = Provider>({
   chainId,
 }: GetProviderArgs = {}): GetProviderResult<TProvider> {
   const client = getClient<TProvider>()
-  return client.getProvider({ chainId }) || client.provider
+  if (chainId) return client.getProvider({ chainId }) || client.provider
+  return client.provider
 }
--- a/packages/core/src/actions/providers/getProvider.ts
+++ b/packages/core/src/actions/providers/getProvider.ts
@@ -8,11 +8,18 @@ export type GetProviderArgs = {
 
 export type GetProviderResult<TProvider extends Provider = Provider> = TProvider
 
+const providers = new Map()
+
 export function getProvider<TProvider extends Provider = Provider>({
   chainId,
 }: GetProviderArgs = {}): GetProviderResult<TProvider> {
   const client = getClient<TProvider>()
-  if (chainId && typeof client.config.provider === 'function')
-    return client.config.provider({ chainId })
+  if (chainId && typeof client.config.provider === 'function') {
+    if (providers.get(chainId)) return providers.get(chainId)
+
+    const provider = client.config.provider({ chainId })
+    providers.set(chainId, provider)
+    return provider
+  }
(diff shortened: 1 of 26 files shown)


packages/core/src/index.ts
```diff
@@ -169,6 +169,7 @@ export type {
 
 export {
   configureChains,
+  debounce,
```
tmm, line 172: Wonder if we should have an `@wagmi/core/internal` entrypoint for utilities, like `debounce`? It's useful that we can export and use in framework-specific packages (e.g. React for `wagmi`) in the monorepo, but is it okay if folks start depending on them?

packages/react/src/hooks/network-status/useBlockNumber.ts
```diff
@@ -41,19 +42,22 @@ export function useBlockNumber({
...
+    // We need to debounce the listener as we want to opt-out
+    // of the behavior where ethers emits a "block" event for
+    // every block that was missed in between the `pollingInterval`.
+    const listener = debounce((blockNumber: number) => {
       // Just to be safe in case the provider implementation
       // calls the event callback after .off() has been called
       if (watch) queryClient.setQueryData(queryKey({ chainId }), blockNumber)
       if (onBlock) onBlock(blockNumber)
-    }
+    }, 50)
```
tmm, line 60: Why `50`?

packages/core/src/actions/providers/getProvider.ts
```diff
@@ -8,11 +8,18 @@ export type GetProviderArgs = {
...
   chainId,
 }: GetProviderArgs = {}): GetProviderResult<TProvider> {
   const client = getClient<TProvider>()
-  if (chainId && typeof client.config.provider === 'function')
-    return client.config.provider({ chainId })
+  if (chainId && typeof client.config.provider === 'function') {
+    if (providers.get(chainId)) return providers.get(chainId)
+
+    const provider = client.config.provider({ chainId })
+    providers.set(chainId, provider)
```
tmm, line 21: Caching on `chainId` seems fine for most use cases, but someone could technically change `client.config.provider` through `client.setState` and this map would be stale. Could [pull it into `client`](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/client.ts#L305) and bust whenever `client.config.provider` changes.

packages/react/src/hooks/network-status/useBlockNumber.ts
jxom: We can't really determine a static time to wait without knowing the average block interval for the chain. For chains that increment the block number for every transaction, this could get very costly, so it seems that 50 is on the safe side. Open to other ideas too.

packages/core/src/actions/providers/getProvider.ts
jxom: Good idea!

packages/core/src/index.ts
jxom: Makes sense!

packages/core/src/client.ts
```diff
@@ -260,6 +257,35 @@ export class Client<
     return this.data
   }
 
+  fetchProvider({ bust, chainId }: { bust?: boolean; chainId?: number } = {}) {
+    let provider_ = this.providers.get(chainId ?? -1)
```
jxom, line 261: We can't force a `chainId` to the in-memory map since a consumer may not pass a `chainId` (e.g. `provider: () => getDefaultProvider()`), so we fall back to `-1`.

packages/core/src/client.ts
```diff
@@ -55,6 +55,7 @@ export class Client<
   TWebSocketProvider extends WebSocketProvider = WebSocketProvider,
 > {
   config: Partial<ClientConfig<TProvider, TWebSocketProvider>>
```
tmm, line 57: Thoughts on removing `Partial`? Would allow use to remove the `|| x.provider` condition in a few places since `provider` is a required constructor argument and will always be available in `this.config`.

packages/react/src/hooks/network-status/useBlockNumber.ts
```diff
@@ -41,19 +42,22 @@ export function useBlockNumber({
...
+    // We need to debounce the listener as we want to opt-out
+    // of the behavior where ethers emits a "block" event for
+    // every block that was missed in between the `pollingInterval`.
+    const listener = debounce((blockNumber: number) => {
       // Just to be safe in case the provider implementation
       // calls the event callback after .off() has been called
       if (watch) queryClient.setQueryData(queryKey({ chainId }), blockNumber)
       if (onBlock) onBlock(blockNumber)
-    }
+    }, 1)
```
tmm, line 60: Comment about why `1` would be good.

Issue #966, opened 2022-09-19
bug: CoinbaseWalletConnector resets to mainnet on refresh


markdalgleish:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.6.6

### Current Behavior

After connecting with Coinbase Wallet and switching networks, if I refresh the page the network is reset to mainnet.

If I open the Coinbase Wallet extension, it still thinks I'm connected to the network I originally switched to.

### Expected Behavior

After refreshing the page, I expect to be on the same network.

### Steps To Reproduce

I've shared a link to a repro below.

1) Start dev server.
2) Connect with Coinbase.
3) Switch to Polygon.
4) Refresh the page.

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

https://github.com/markdalgleish/wagmi-coinbase-chain-issue

### Anything else?

I've dug into the logic a bit, and I can manually pass in a `chainId` to the Coinbase connector, but internally it falls back to the first chain if a chain ID wasn't passed in.

Instead, this logic might need to fall back to the chain the Coinbase Wallet extension is already connected to. When using the extension, I found that the `CoinbaseWalletSDK` instance has a `walletExtension` property with a `getChainId` method that returns the expected chain ID on load. If I hard code this value as the 2nd arg to `makeWeb3Provider`, it seems to resolve the issue.

jxom:
Fixed on `0.6.7`

PR #967, opened 2022-09-19
fix: respect connected coinbase wallet extension chain id

jxom:
## Description

This PR fixes an issue (#966) in `CoinbaseWalletConnector` where the browser extension would unintendedly reset the network when the browser is refreshed.

### DIFF
--- a/packages/core/src/connectors/coinbaseWallet.ts
+++ b/packages/core/src/connectors/coinbaseWallet.ts
@@ -134,7 +134,7 @@ export class CoinbaseWalletConnector extends Connector<
       /**
        * Mock implementations to retrieve private `walletExtension` method
        * from the Coinbase Wallet SDK.
-       * */
+       */
       abstract class WalletProvider {
         // https://github.com/coinbase/coinbase-wallet-sdk/blob/b4cca90022ffeb46b7bbaaab9389a33133fe0844/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts#L927-L936
         abstract getChainId(): number
--- a/packages/core/src/connectors/coinbaseWallet.ts
+++ b/packages/core/src/connectors/coinbaseWallet.ts
@@ -118,12 +118,6 @@ export class CoinbaseWalletConnector extends Connector<
 
   async getProvider() {
     if (!this.#provider) {
-      const chain =
-        this.chains.find((chain) => chain.id === this.options.chainId) ||
-        this.chains[0]
-      const chainId = this.options.chainId || chain?.id
-      const jsonRpcUrl = this.options.jsonRpcUrl || chain?.rpcUrls.default
-
       let CoinbaseWalletSDK = (await import('@coinbase/wallet-sdk')).default
       // Workaround for Vite dev import errors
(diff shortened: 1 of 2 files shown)


packages/core/src/connectors/coinbaseWallet.ts
```diff
@@ -135,8 +129,20 @@ export class CoinbaseWalletConnector extends Connector<
         CoinbaseWalletSDK = (<{ default: typeof CoinbaseWalletSDK }>(
           (<unknown>CoinbaseWalletSDK)
         )).default
-
       this.#client = new CoinbaseWalletSDK(this.options)
+
+      // @ts-expect-error – accessing `private` method
+      const walletExtensionChainId = this.#client.walletExtension?.getChainId()
```
tmm, line 135: Will this work with the mobile app? Looks `walletExtension` [only exists in an injected context](https://github.com/coinbase/coinbase-wallet-sdk/blob/78e1fcb95d18fdd6eb4c613738bbe348d10ca59c/packages/wallet-sdk/src/CoinbaseWalletSDK.ts#L234).

tmm, line 135: Could explicitly type this instead of using `@ts-expect-error`. Not sure which one is better though.

```suggestion
      // Mock implementations for typing
      class WalletProvider {
        // https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/src/provider/CoinbaseWalletProvider.ts#L927-L936
        getChainId(): number { return 1 }
      }
      class Client {
        // https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/src/CoinbaseWalletSDK.ts#L233-L235
        get walletExtension(): WalletProvider | undefined { return }
      }
      const walletExtensionChainId = (this.#client as unknown as Client).walletExtension?.getChainId()
```

tmm (commented):
We should check that this works with mobile and follow up with the CBW team. Not ideal to rely on this private method. I wonder if an `eth_chainId` request could work also?

packages/core/src/connectors/coinbaseWallet.ts
jxom: Yep! Works with the mobile app. It will fallback to the previous behavior if the `walletExtension` is not found.

jxom: Probably better than `@ts-expect-error`!

packages/core/src/connectors/coinbaseWallet.ts
```diff
@@ -135,8 +129,33 @@ export class CoinbaseWalletConnector extends Connector<
         CoinbaseWalletSDK = (<{ default: typeof CoinbaseWalletSDK }>(
           (<unknown>CoinbaseWalletSDK)
         )).default
-
       this.#client = new CoinbaseWalletSDK(this.options)
+
+      /**
+       * Mock implementations to retrieve private `walletExtension` method
+       * from the Coinbase Wallet SDK.
+       * */
```
tmm, line 137:
```suggestion
       */
```

tmm (approved):

PR #968, opened 2022-09-19
docs: add example for `usePrepareContractWrite` with dynamic arguments

jxom:
## Description

This PR adds an example of [how to use `useContractWrite` with dynamic arguments](https://wagmi-git-jxom-prepare-hooks-dynamic-example-wagmi-dev.vercel.app/examples/contract-write-dynamic) & also tweaks the tests to use the new `tokenId` arg (and remove `getTotalSupply`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "scripts": {
-    "anvil": "anvil --fork-url $ANVIL_FORK_URL --fork-block-number 15132000",
+    "anvil": "anvil --fork-url $ANVIL_FORK_URL --fork-block-number 15564160",
     "build": "preconstruct build",
     "changeset:release": "pnpm build && changeset publish",
     "changeset:version": "changeset version && pnpm install --lockfile-only",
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "root",
   "scripts": {
-    "anvil": "anvil --fork-url $ANVIL_FORK_URL --fork-block-number 15132000",
+    "anvil": "anvil --fork-url $ANVIL_FORK_URL --fork-block-number 15564160",
```
jxom, line 4: Updating the block number so we can query the new mint example contract.

jxom, line 4: Updating the block number so we can query the new mint example contract.

packages/core/src/actions/accounts/fetchBalance.test.ts
```diff
@@ -19,10 +19,10 @@ describe('fetchBalance', () => {
         ).toMatchInlineSnapshot(`
           {
             "decimals": 18,
-            "formatted": "1.889009973656812885",
+            "formatted": "0.40742480512617271",
```
jxom, line 22: These snapshots are a side-effect from updating the block number.

PR #978, opened 2022-09-21
tests: update test wagmi contract

jxom:
## Description

This PR updates the `wagmiContractConfig` to use the new wagmi mint example contract. Since the new contract supports a `tokenId` argument to the `mint` function, we can remove the `getTotalSupply` test util.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -162,7 +162,7 @@ jobs:
         run: anvil --fork-url $ANVIL_FORK_URL --fork-block-number $ANVIL_BLOCK_NUMBER &
         env:
           ANVIL_FORK_URL: ${{ secrets.ANVIL_FORK_URL }}
-          ANVIL_BLOCK_NUMBER: 15564160
+          ANVIL_BLOCK_NUMBER: 15578840
       - name: Test React 18
         if: matrix.react-version == 18
         run: pnpm test:coverage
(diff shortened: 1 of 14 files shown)


.github/workflows/main.yml
```diff
@@ -162,7 +162,7 @@ jobs:
         run: anvil --fork-url $ANVIL_FORK_URL --fork-block-number $ANVIL_BLOCK_NUMBER &
         env:
           ANVIL_FORK_URL: ${{ secrets.ANVIL_FORK_URL }}
-          ANVIL_BLOCK_NUMBER: 15132000
+          ANVIL_BLOCK_NUMBER: 15564160
```
jxom, line 165: Updating the block number so we can discover the contract.

packages/core/src/actions/accounts/fetchBalance.test.ts
```diff
@@ -19,10 +19,10 @@ describe('fetchBalance', () => {
         ).toMatchInlineSnapshot(`
           {
             "decimals": 18,
-            "formatted": "1.889009973656812885",
+            "formatted": "0.40742480512617271",
```
jxom, line 22: Side effect from updating block number.

packages/react/src/hooks/accounts/useBalance.test.ts
```diff
@@ -19,10 +19,10 @@ describe('useBalance', () => {
       {
         "data": {
           "decimals": 18,
-          "formatted": "1.889009973656812885",
+          "formatted": "0.40742480512617271",
```
jxom, line 22: Side effect from updating block number.

packages/core/test/utils.ts
```diff
@@ -233,3 +215,7 @@ export function getCrowdfundArgs({
   }
   return Object.values(data)
 }
+
+export function getRandomTokenId() {
+  return BigNumber.from(Math.floor(Math.random() * 1000) + 69420)
```
tmm, line 220: don't want to use `bigint` 😛

tmm (approved):

PR #993, opened 2022-09-26
Jxom/support bytes32 erc20

### LINKED ISSUE #939: bug: useBalance fails when token has bytes32 symbol

### Package Version

0.6.6

### Current Behavior

`useBalance` fails on tokens that have their symbol as bytes32, unable to decode the bytes32 result of the function call. For example, `useBalance` with Maker (0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2) returns:
```
Failed to decode contract function result.

Config:
{
  "addressOrName": "0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2",
  "contractInterface": "...",
  "functionName": "symbol",
  "chainId": 1
}

Details: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="symbol()", data="0x4d4b520000000000000000000000000000000000000000000000000000000000", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.7.0)
```

### Expected Behavior

Would like to be able to use the useBalance hook and get a balance value back whether the other values are returned or not. symbol, in this case, doesn't seem like it should block the core function of the hook.

### Steps To Reproduce

```
import { erc20ABI } from 'wagmi';
const chainId = 1;
const tokenAddress = '0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2';
const { data: balance, error } = useBalance({
    addressOrName: accountAddress,
    chainId,
    token: tokenAddress,
  });
```

_No response_

### Anything else?

_No response_

jxom:
## Description

This PR supports ERC20 contracts that represent `string`s as `bytes32`s such as MKR. Fixes #939

### DIFF
--- a/.changeset/short-chefs-think.md
+++ b/.changeset/short-chefs-think.md
@@ -1,0 +1,6 @@
+---
+'@wagmi/core': patch
+'wagmi': patch
+---
+
+Support ERC20 contracts that represent strings as bytes32
--- a/packages/core/src/actions/accounts/fetchBalance.test.ts
+++ b/packages/core/src/actions/accounts/fetchBalance.test.ts
@@ -106,6 +106,25 @@ describe('fetchBalance', () => {
         `)
       })
 
+      it('bytes32 contract', async () => {
+        expect(
+          await fetchBalance({
+            addressOrName: '0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2',
+            token: '0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2',
+          }),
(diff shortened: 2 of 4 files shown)


tmm (commented):
Looks great! Should we add similar logic to [`fetchToken`](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/actions/contracts/fetchToken.ts#L37)?

tmm (approved):

PR #998, opened 2022-09-27
fix: preserve order of `readContracts` contracts

### LINKED ISSUE #999: `useContractReads` result not matching `contracts` input order

jxom:
## Description

This PR fixes an issue where the ordering of the `contracts` passed to `useContractReads` would not be preserved (they were grouped by chain id instead). 

Fixes #999


packages/core/src/actions/contracts/readContracts.ts
```diff
@@ -37,26 +37,36 @@ export async function readContracts<Data extends any[] = Result[]>({
...
+    // We also want to preserve the positions (index) of the `contracts` argument
+    // so we can reorder it later.
     const contractsByChainId = contracts.reduce<{
-      [chainId: number]: ReadContractsConfig['contracts']
-    }>((contracts, contract) => {
+      [chainId: number]: {
+        contract: ReadContractsContract
+        index: number
+      }[]
+    }>((contracts, contract, index) => {
```
tmm, line 50: Assuming yes, but do we need this generic here or can it be inferred from `reduce` return type?

packages/core/src/actions/contracts/readContracts.ts
```diff
@@ -37,26 +37,36 @@ export async function readContracts<Data extends any[] = Result[]>({
...
       const chainId = contract.chainId ?? provider.network.chainId
       return {
         ...contracts,
-        [chainId]: [...(contracts[chainId] || []), contract],
+        [chainId]: [...(contracts[chainId] || []), { contract, index }],
       }
     }, {})
```
tmm, line 56: Could move the `reduce` generic here to type `{}` instead (think that will work).

tmm (approved):

holic:
Any chance of releasing as a patch for 0.6.x? Currently blocked on this and RainbowKit/ConnectKit haven't updated to 0.7.x yet.

Issue #1098, opened 2022-10-16
bug: usePrepareContractWrite error when disabled


soccerarenadev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.7.3

### Current Behavior

Using the usePrepareContractWrite hook, even if the "enabled" parameter is set to false, it returns an error.
In this case, I pass no parameter as  enabled is set to false.

<img width="565" alt="image" src="https://user-images.githubusercontent.com/99416733/196054010-79fb20c4-2a42-406a-80d5-67cbd7c3a80b.png">

### Expected Behavior

Not being an error as is set to disabled.

jxom:
It could be because the result is cached, so it is returning the cached result (error) instead of fetching again. `wagmi` uses React Query internally as our caching layer – you can read on this behavior [here](https://tanstack.com/query/v4/docs/guides/disabling-queries). You may need to opt-out of this behaviour using `internal.isFetchedAfterMount` flag returned from `usePrepareContractWrite`.

Issue #1065, opened 2022-10-12
bug: `usePrepareContractWrite` enabled doesn't work


flux0uz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.6.8

### Current Behavior

`enabled: false` doesn't disable the preparation of the contract write function.

### Expected Behavior

Setting `enabled: false` should disable the preparation of the contract write function.

### Steps To Reproduce

```
const prepareBuy = usePrepareContractWrite({
    addressOrName: tokenAddress,
    contractInterface: ['function buyShare(address from, uint256 amount) public returns (uint256)'],
    functionName: 'buyShare',
    args: [address, parseUnits(totalValue, 6)],
    enabled: false,
  });
  ```

PR #1099, opened 2022-10-16
feat: add `isFetchedAfterMount` to return value

jxom:
## Description

The `isFetchedAfterMount` flag will be useful for folks who do not want to show previously cached data in their UI when `enabled` is set to `false`. Will help a lot with issues like #1098 & #1065.

Note: I chose not to omit this from the `internal` namespace to avoid breaking changes.


tmm (approved):
lgtm! wdyt about adding a comment to the `internal` `isFetchedAfterMount` with a `TODO` scheduled for deletion in next minor version?

PR #1132, opened 2022-10-21
Add erc4626ABI support

toniocodo:
## Description

This PR adds erc4626 ABI for easy integration of the [tokenized vault standard](https://ethereum.org/en/developers/docs/standards/tokens/erc-4626)

Your ENS/address: toniocodo.eth

### DIFF
--- a/.changeset/rich-months-hide.md
+++ b/.changeset/rich-months-hide.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/core': minor
+---
+
+Added erc4626ABI support
--- a/.changeset/rich-months-hide.md
+++ b/.changeset/rich-months-hide.md
@@ -1,5 +1,0 @@
----
-'@wagmi/core': minor
----
-
-Added erc4626ABI support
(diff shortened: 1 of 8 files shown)


.changeset/rich-months-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@wagmi/core': minor
```
tmm, line 2: Let's make this a `patch`

packages/core/src/constants/abis.ts
```diff
@@ -775,3 +775,582 @@ export const multicallABI = [
...
+/**
+ * [ERC-4626 Tokenized Vaults Standard](https://ethereum.org/en/developers/docs/standards/tokens/erc-4626)
+ */
+export const erc4626ABI = [
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: true,
+        internalType: 'address',
```
tmm, line 788: Let's get rid of `internalType` everywhere.

tmm (changes requested):
Left a couple comments. We'll also want to export `erc4626ABI` from `packages/(core|react)/src/index.ts` files and add a patch changeset for `wagmi`.

PR #1176, opened 2022-10-27
fix: move away from public Alchemy RPC URLs

jxom:
## Description

The public Alchemy RPC URL is currently rate limited. Ideally we should move away from Alchemy RPC URLs in the public provider. Also made a note to make `apiKey` required in `alchemyProvider` and `infuraProvider` for `0.8`.


packages/core/src/constants/rpcs.ts
```diff
@@ -89,11 +89,11 @@ type PublicChains = Extract<
...
 export const publicRpcUrls: Record<PublicChains, string> = {
-  mainnet: `${alchemyRpcUrls.mainnet}/${defaultAlchemyApiKey}`,
-  ropsten: `${alchemyRpcUrls.ropsten}/${defaultAlchemyApiKey}`,
-  rinkeby: `${alchemyRpcUrls.rinkeby}/${defaultAlchemyApiKey}`,
-  goerli: `${alchemyRpcUrls.goerli}/${defaultAlchemyApiKey}`,
-  kovan: `${alchemyRpcUrls.kovan}/${defaultAlchemyApiKey}`,
+  mainnet: 'https://cloudflare-eth.com',
+  ropsten: 'https://rpc.ankr.com/eth_ropsten',
+  rinkeby: 'https://rpc.ankr.com/eth_rinkeby',
+  goerli: 'https://rpc.goerli.mudit.blog/',
```
tmm, line 95: let's move this to ankr for now and find a long term replacement later

tmm (approved):

Issue #1127, opened 2022-10-21
bug: useContractRead method with watch affecting other call


yujiym:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.7.5

### Current Behavior

`useContractRead` method with `watch: true` option affecting other calls.

1. `useContractRead` with `watch: false` -> called every new block (should be called once)
2. `useContractRead` with `watch: true` -> called every new block

![Screen Shot 2022-10-21 at 16 38 11](https://user-images.githubusercontent.com/1409395/197140304-4dd057cb-00c3-4be2-9214-94c07bf27717.png)

If both methods are set to `watch: false`, they will only be called once.

### Expected Behavior

1. `useContractRead` with `watch: false` -> called once
2. `useContractRead` with `watch: true` -> called every new block

### Steps To Reproduce

Open codesandbox link and watch browser's console.

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

https://codesandbox.io/p/sandbox/boring-pascal-w4jnlq

### Anything else?

Since `useBlockNumber` has the same behavior, `useBlockNumber`'s state called inside `useContractRead` may be the cause.

This cloud be due to the same cause. https://github.com/wagmi-dev/wagmi/issues/888

jxom:
Can now use the `scopeKey` config option on `useContractRead` to scope the hook to a given key to isolate it from other instances!

rsodre:
Does it mean that if we don't use the `scopeKey`, one single hook using `watch` wil trigger all other hooks to refetch?
And in a group of hooks with `scopeKey`, one update will trigger all the other hooks?

rsodre:
> 

Does it mean that if we don't use the scopeKey, one single hook using watch wil trigger all other hooks to refetch?
And in a group of hooks with scopeKey, one update will trigger all the other hooks?

tmm:
Correct, the fetching will be deduplicated across query keys though.

PR #1206, opened 2022-10-31
feat: add `scopeKey` config option to hooks

jxom:
## Description

This PR adds a new configuration option to our "query" hooks, `scopeKey`, which scope the hook's cache to a given context. This key will be useful for situations like #1127 where consumers may want to have hooks which are more "pure" (ie. the result of a hook shouldn't affect other instances of it). The `scopeKey` essentially gives the consumer access to add on an arbitrary key to the query key.

### DIFF
--- a/packages/react/src/hooks/accounts/useBalance.test.ts
+++ b/packages/react/src/hooks/accounts/useBalance.test.ts
@@ -82,7 +82,7 @@ describe('useBalance', () => {
         `)
       })
 
-      it('contextKey', async () => {
+      it('scopeKey', async () => {
         const { result, waitFor } = renderHook(() => {
           return {
             balance: useBalance({
@@ -94,7 +94,7 @@ describe('useBalance', () => {
             }),
             balanceWithContextKey: useBalance({
               addressOrName: '0xA0Cf798816D4b9b9866b5330EEa46a18382f251e',
-              contextKey: 'wagmi',
+              scopeKey: 'wagmi',
               enabled: false,
             }),
           }
--- a/packages/react/src/hooks/accounts/useBalance.test.ts
+++ b/packages/react/src/hooks/accounts/useBalance.test.ts
@@ -88,11 +88,11 @@ describe('useBalance', () => {
             balance: useBalance({
               addressOrName: '0xA0Cf798816D4b9b9866b5330EEa46a18382f251e',
             }),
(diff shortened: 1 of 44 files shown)


packages/react/src/types/index.ts
```diff
@@ -77,7 +80,10 @@ export type InfiniteQueryConfig<Data, Error> = Pick<
   | 'onError'
   | 'onSettled'
   | 'onSuccess'
->
+> & {
+  /** Scope the cache to a given context. */
+  contextKey?: string
```
tmm, line 85: Wonder if it's better to call this something else so it doesn't get confused with React Context or React Query's `context` option if we choose to add something similar in the future? Maybe `scopeKey` or `cacheKey`?

jxom: Yeah, maybe `scopeKey` is a better fit. I have a feeling people may take `cacheKey` too literally (ie. think it is the _actual_ query key).

packages/react/src/hooks/accounts/useBalance.test.ts
```diff
@@ -82,6 +82,37 @@ describe('useBalance', () => {
...
+        const { result, waitFor } = renderHook(() => {
+          return {
+            balance: useBalance({
+              addressOrName: '0xA0Cf798816D4b9b9866b5330EEa46a18382f251e',
+            }),
+            balanceWithoutContextKey: useBalance({
+              addressOrName: '0xA0Cf798816D4b9b9866b5330EEa46a18382f251e',
+              enabled: false,
+            }),
+            balanceWithContextKey: useBalance({
```
tmm, line 95:
```suggestion
            balanceWithScopeKey: useBalance({
```

packages/react/src/hooks/contracts/useContractRead.ts
```diff
@@ -64,6 +69,8 @@ function queryFn<
   }: QueryFunctionArgs<typeof queryKey>) => {
     if (!abi) throw new Error('abi is required')
     if (!address) throw new Error('address is required')
+    if (!args) throw new Error('args is required')
```
tmm, line 72: `args` isn't always required (e.g. wagmigotchi `getHunger`)

tmm (changes requested):

tmm (approved):

PR #1208, opened 2022-10-31
feat: update WalletConnectConnector for v2

ganchoradkov:
## Description

Add WalletConnect v2 support to the existing WC Connector.

## Additional Information
WalletConnect v1 is to be depreciated on March 1, 2023.
This v2 connector will allow wagmi users to have enough time to prepare and migrate to WalletConnect v2 which offers significantly better developer & user experience

Your ENS/address:


packages/core/package.json
```diff
@@ -84,6 +84,8 @@
   "peerDependencies": {
     "@coinbase/wallet-sdk": ">=3.3.0",
     "@walletconnect/ethereum-provider": ">=1.7.5",
+    "@walletconnect/qrcode-modal": ">=1.8.0",
+    "@walletconnect/universal-provider": ">=2.1.0",
```
tmm, line 88: Added range in

packages/core/src/connectors/walletConnect.ts
```diff
@@ -92,21 +188,31 @@ export class WalletConnectConnector extends Connector<
...
+    }
   }
 
   async getAccount() {
     const provider = await this.getProvider()
-    const accounts = provider.accounts
+    let accounts
+    if (this.version === '1')
+      accounts = (provider as WalletConnectProvider).accounts
+    else accounts = await provider.enable() // TODO: Should we use `.request({ method: 'eth_accounts' })` instead?
```
tmm, line 205: This returns an empty array when calling `isAuthorized` so auto connecting is not working.

packages/core/src/connectors/walletConnect.ts
```diff
@@ -206,4 +325,8 @@ export class WalletConnectConnector extends Connector<
   protected onDisconnect = () => {
     this.emit('disconnect')
   }
+
+  protected onDisplayUri = (uri: string) => {
+    this.emit('message', { type: 'display_uri', data: uri })
+  }
```
tmm, line 331: Decoupled `Modal` from this

packages/core/src/connectors/walletConnect.ts
```diff
@@ -56,6 +73,58 @@ export class WalletConnectConnector extends Connector<
...
+                new Promise((_res, reject) =>
+                  provider.on('display_uri', async (uri: string) => {
+                    const { default: Modal } = await import(
+                      '@walletconnect/qrcode-modal'
+                    )
+                    Modal.open(uri, () => reject(new Error('user rejected')))
+                  }),
+                ),
+              ]
+            : []),
```
tmm, line 118: Need this block so when user closes out of QR Code Modal, "rejected request" error bubbles up to user.

packages/core/src/connectors/walletConnect.ts
```diff
@@ -56,6 +73,58 @@ export class WalletConnectConnector extends Connector<
...
+        ])
+
+        // If execution reaches here, connection was successful and we can close modal.
+        if (this.options.qrcode) {
+          const { default: Modal } = await import('@walletconnect/qrcode-modal')
+          Modal.close()
+        }
```
tmm, line 125: This assumption feels a little shaky. Is this the best way to know if `provider.connect` succeeded?

packages/core/src/connectors/walletConnect.ts
ganchoradkov: Updated the `universal-provider` to `2.1.4` where this issue is resolved. Now returns correct the address from its persistence storage after reload

packages/core/src/connectors/walletConnect.ts
ganchoradkov: I think it should be fine. If the connection request is declined by the user an exception is thrown and is being captured correctly `catch` block

packages/core/src/connectors/walletConnect.ts
```diff
@@ -9,36 +11,49 @@ import type { Chain } from '../types'
...
+  qrcode?: boolean
+} & (
+  | ({
+      version?: '1'
+    } & ConstructorParameters<typeof WalletConnectProvider>[0])
+  | ({
+      /**
+       * Project ID associated with [WalletConnect account](https://cloud.walletconnect.com)
+       */
+      projectId: NonNullable<UniversalProviderOpts['projectId']>
```
seendsouza, line 30: Isn't it possible to have no projectId if using your own relayer for walletconnect v2?

jxom:
Closing in favor of https://github.com/wagmi-dev/references/pull/19

PR #1232, opened 2022-11-04
feat: validate chain configured for connector

### LINKED ISSUE #1222: bug: usePrepareContractWrite fails if chainId is not 1

### Package Version

0.7.9

### Current Behavior

When calling `usePrepareContractWrite` while the user's wallet is connected to a non-1 chainId, the preparation and consecutive `useContractWrite` fails.

### Expected Behavior

Preparation should succeed.

### Steps To Reproduce

1. Open https://codesandbox.io/s/nostalgic-wilson-6td2dy?file=/src/App.js
2. Set your wallet to Polygon (chain ID 137)
3. Click the connect button
4. Watch the interface fail

Note how at no point was the chain ID 1.

https://codesandbox.io/s/nostalgic-wilson-6td2dy?file=/src/App.js

### Anything else?

Thanks for your work on this package!

tmm:
## Description

#940 improved `chainId` validation, but did not address an edge case where `Signer` could still be on the wrong chain because it was not configured at the `Connector`-level (resulting in [low-level ethers error](https://github.com/wagmi-dev/wagmi/issues/1222)).

Fixes #1222 

## Additional Information

Your ENS/address: awkweb.eth

### DIFF
--- a/packages/react/src/hooks/contracts/usePrepareContractWrite.ts
+++ b/packages/react/src/hooks/contracts/usePrepareContractWrite.ts
@@ -2,14 +2,15 @@ import {
   FetchSignerResult,
   PrepareWriteContractConfig,
   PrepareWriteContractResult,
+  Signer,
   prepareWriteContract,
 } from '@wagmi/core'
 import { Abi } from 'abitype'
-import { Signer, providers } from 'ethers'
+import { providers } from 'ethers'
 
 import { QueryConfig, QueryFunctionArgs } from '../../types'
-import { useSigner } from '../accounts'
-import { useChainId, useQuery } from '../utils'
+import { useNetwork, useSigner } from '../accounts'
+import { useQuery } from '../utils'
 
 export type UsePrepareContractWriteConfig<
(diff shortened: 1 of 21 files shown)


packages/react/src/hooks/contracts/usePrepareContractWrite.ts
```diff
@@ -118,14 +119,12 @@ export function usePrepareContractWrite<
     onSuccess,
   }: UsePrepareContractWriteConfig<TAbi, TFunctionName> = {} as any,
 ) {
-  const activeChainId = useChainId()
-  const { data: signer } = useSigner<providers.JsonRpcSigner>({
-    chainId: chainId ?? activeChainId,
-  })
+  const { chain: activeChain } = useNetwork()
```
tmm, line 122: Using chain id from `useNetwork` instead of `useChainId` since `useNetwork` returns a chain for the connected wallet (and signer), while `useChainId` returns a chain id for the `client.provider`.

Most of the time, these are the same, except when a wallet is connected to a chain not configured on the connector.

PR #1289, opened 2022-11-11
feat: throw when waiting on reverted transaction

tmm:
## Description

## Additional Information

- [ ] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

### DIFF
--- a/packages/core/src/actions/transactions/waitForTransaction.ts
+++ b/packages/core/src/actions/transactions/waitForTransaction.ts
@@ -1,4 +1,5 @@
 import type { providers } from 'ethers'
+import { toUtf8String } from 'ethers/lib/utils.js'
 
 import type { Hash } from '../../types'
 import { getProvider } from '../providers'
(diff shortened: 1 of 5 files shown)


packages/core/src/actions/transactions/waitForTransaction.ts
```diff
@@ -11,32 +12,45 @@ export type WaitForTransactionArgs = {
...
+  if (args.hash) {
+    promise = provider.waitForTransaction(args.hash, confirmations, timeout)
+    // eslint-disable-next-line testing-library/await-async-utils
+  } else promise = args.wait(confirmations)
+
+  const receipt = await promise
+  if (receipt.status === 0) {
+    const code = await provider.call(receipt, receipt.blockNumber)
+    const reason = toUtf8String(`0x${code.substring(138)}`)
+    throw new Error(reason)
```
tmm, line 52: In the future, we could try to grab the lower-level error message, etc.

.changeset/poor-guests-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@wagmi/core': patch
+---
+
+Changed `waitForTransaction` behavior to throw an error if the transaction reverted.
```
tmm, line 5: @jxom thinking this should probably be `minor`, but curious what you think about as `patch`?

.changeset/poor-guests-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@wagmi/core': patch
```
tmm, line 2: @jxom thinking this should probably be `minor`, but curious what you think about as `patch`?

jxom: It is technically a breaking change, but I wonder how breaking it is. 

Folks could be doing `isSuccess && data.status === 0`, so maybe we should play it safe and mark it as a `minor`?

jxom (approved):

PR #1294, opened 2022-11-12
fix: return type for prepareContractWrite

tmm:
## Description

Sets `abi` return type value for `usePrepareContractWrite`/`prepareContractWrite` to generic value `TAbi` (when not inferrable as `Abi`) instead of `readonly unknown[]` so that type info is preserved.

## Additional Information

Your ENS/address: awkweb.eth

### DIFF
--- a/packages/core/src/actions/contracts/prepareWriteContract.ts
+++ b/packages/core/src/actions/contracts/prepareWriteContract.ts
@@ -51,9 +51,7 @@ export type PrepareWriteContractResult<
   TAbi = Abi,
   TFunctionName extends string = string,
 > = {
-  abi: TAbi extends Abi
-    ? [ExtractAbiFunction<TAbi, TFunctionName>]
-    : readonly unknown[]
+  abi: TAbi extends Abi ? [ExtractAbiFunction<TAbi, TFunctionName>] : TAbi
   address: string
   chainId?: number
   functionName: TFunctionName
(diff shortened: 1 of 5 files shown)


packages/core/src/actions/contracts/prepareWriteContract.ts
```diff
@@ -129,10 +127,7 @@ export async function prepareWriteContract<
   const minimizedAbi = minimizeContractInterface({
     abi: abi as Abi, // TODO: Remove cast and still support `Narrow<TAbi>`
     functionName,
-  }) as TAbi extends Abi
-    ? [ExtractAbiFunction<TAbi, TFunctionName>]
-    : readonly unknown[]
```
tmm, line 134: By returning `readonly unknown[]`, we lost type info so can just return the same generic type `TAbi`

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
```diff
@@ -13,7 +13,7 @@ import { useNetwork, useSigner } from '../accounts'
 import { useQuery } from '../utils'
 
 export type UsePrepareContractWriteConfig<
-  TAbi extends Abi | readonly unknown[] = Abi,
+  TAbi = Abi,
```
tmm, line 16: drop unnecessary constraint

packages/react/src/hooks/contracts/usePrepareContractWrite.ts
```diff
@@ -27,7 +27,7 @@ export type UsePrepareContractWriteConfig<
     isFunctionNameOptional: true
   }
 > &
-  QueryConfig<PrepareWriteContractResult, Error>
+  QueryConfig<PrepareWriteContractResult<TAbi, TFunctionName>, Error>
```
tmm, line 30: add inference to `data`, etc. previously only worked for `config`

PR #1335, opened 2022-11-19
fix: use `eth_accounts` for `getSigner` on InjectedConnector

### LINKED ISSUE #1334: bug: Error when calling useContractWrite write method with Core Wallet

### Package Version

0.7.15

### Current Behavior

When I connect to Core Wallet using the InjectedConnector, I get the following error when calling `useContractWrite`'s write method: `MetaMask - RPC Error: Resource unavailable. {code: -32002, message: 'Resource unavailable.'`

I am logging the connector before and after calling the write method. Before calling it I can see that the InjectedConnector is connected and ready, but after calling `write` the connected connector is `undefined`. But if I look at the Core Wallet extension it stil displays the wallet as connected.

When I use `InjectedConnector` to connect to metamask this does not happen. 

### Expected Behavior

It should call the `write` method and execute the transaction. 

### Steps To Reproduce

1. Use InjectedConnector to connect to Core Wallet
2. Use useContractWrite hook to get `write` method - using ERC20 transfer method
3. Call `write` while connected to core wallet

_No response_

### Anything else?

_No response_

jxom:
## Description

Fixes #1334

It seems that Core Wallet was reverting `eth_requestAccounts` when the account has already given permissions. We probably want to be using `eth_accounts` instead of `eth_requestAccounts` on the injected connector's `getSigner` method as the account should always be authorized with the wallet at this point.

## Additional Information

Your ENS/address: jxom.eth


tmm (commented):
Good find! Instead of updating `getSigner` seems like it might be better to update the implementation of `getAccount` to use `eth_accounts` (a la [`CoinbaseWalletConnector.getAccount`](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/connectors/coinbaseWallet.ts#L108)) AND update `connect` to explicitly call `eth_requestAccounts` instead of `getAccount`.

We'll also need to update `MetaMaskConnector.connect` to handle this logic when `UNSTABLE_shimOnConnectSelectAccount` is `true`/`false`.

jxom:
> Good find! Instead of updating `getSigner` seems like it might be better to update the implementation of `getAccount` to use `eth_accounts` (a la [`CoinbaseWalletConnector.getAccount`](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/connectors/coinbaseWallet.ts#L108)) AND update `connect` to explicitly call `eth_requestAccounts` instead of `getAccount`.
> 
> We'll also need to update `MetaMaskConnector.connect` to handle this logic when `UNSTABLE_shimOnConnectSelectAccount` is `true`/`false`.

Good idea!

ReachCambo (approved):

packages/core/src/connectors/metaMask.ts
```diff
@@ -61,25 +64,28 @@ export class MetaMaskConnector extends InjectedConnector {
...
           })
       }
 
-      const account = await this.getAccount()
+      if (!account) {
+        const accounts = await provider.request({
+          method: 'eth_requestAccounts',
+        })
+        account = getAddress(accounts[0] as string)
+      }
```
tmm, line 87: Need to call `eth_requestAccounts` for cases where:
- `UNSTABLE_shimOnConnectSelectAccount` is `false`
- `UNSTABLE_shimOnConnectSelectAccount` is `true` and account isn't connected (because `wallet_requestPermissions` is never called).

tmm (approved):

PR #1344, opened 2022-11-21
0.9.x

jxom:
Candidate branch for `0.9.x`


rsodre:
I think this page needs to be updated...
https://wagmi.sh/react/providers/configuring-chains

rsodre:
Seems ok, I think I was looking at the wrong doc

PR #1351, opened 2022-11-24
Add getProvider to InjectedConnector options

ekenigs:
## Description

Add `getProvider` to `InjectedConnector` options. This is necessary for many EVM (MetaMask) compatible wallets in the Polkadot ecosystem that are injecting a provider not to `window.ethereum`. 

- [SubWallet](https://chrome.google.com/webstore/detail/subwallet-polkadot-extens/onhogfjeacnfoofkfgppdlbmlmnplgbn) : `window.SubWallet`
- [Talisman](https://chrome.google.com/webstore/detail/talisman-polkadot-wallet/fijngjgcjhjmmpcmkeiomlglpeiijkld): `window.talismanEth`
- [Enkrypt](https://chrome.google.com/webstore/detail/enkrypt-ethereum-polkadot/kkpllkodjeloidieedojogacfhpaihoh): `window.enkrypt.providers.ethereum`

## Additional Information

### DIFF
--- a/.changeset/pink-pillows-dream.md
+++ b/.changeset/pink-pillows-dream.md
@@ -4,7 +4,7 @@
 
 Added `getProvider` to `InjectedConnector` options.
 
-Example: 
+Example:
 
 ```ts
 import { InjectedConnector } from '@wagmi/core/connectors/injected'
@@ -16,3 +16,4 @@ const connector = new InjectedConnector({
       typeof window !== 'undefined' ? window.SubWallet : undefined,
   },
 })
+```
--- a/.changeset/pink-pillows-dream.md
+++ b/.changeset/pink-pillows-dream.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/core': patch
+---
+
+Added getProvider to InjectedConnector options
--- a/.changeset/pink-pillows-dream.md
+++ b/.changeset/pink-pillows-dream.md
@@ -2,4 +2,17 @@
 '@wagmi/core': patch
 ---
 
-Added getProvider to InjectedConnector options
+Added `getProvider` to `InjectedConnector` options.
+
+Example: 
+
+```ts
+import { InjectedConnector } from '@wagmi/core/connectors/injected'
+
+const connector = new InjectedConnector({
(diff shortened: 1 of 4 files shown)


.changeset/pink-pillows-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+'@wagmi/core': patch
+---
+
+Added getProvider to InjectedConnector options
```
jxom, line 5: Nice changeset! Maybe we can just add some quick usage as well.

```suggestion
Added `getProvider` to `InjectedConnector` options.

Example: 

```ts
import { InjectedConnector } from '@wagmi/core/connectors/injected'

const connector = new InjectedConnector({
  options: {
    name: 'SubWallet',
    getProvider: () =>
      typeof window !== 'undefined' ? window.SubWallet : undefined,
  },
})
```

packages/core/src/connectors/injected.ts
```diff
@@ -39,7 +41,7 @@ export class InjectedConnector extends Connector<
 > {
   readonly id: string
   readonly name: string
-  readonly ready = typeof window != 'undefined' && !!window.ethereum
+  readonly ready: boolean = false
```
jxom, line 44: Probably don't need a default now since we assign in the constructor.

```suggestion
  readonly ready: boolean
```

ekenigs:
@jxom thanks for your review.

tmm (commented):
Added a comment. Would also be great if we could pull some of the behavior around the `MetaMaskConnector` into this so it could be completely powered by `getProvider` as well (let me know if you want to chat about this more).

ekenigs:
@tmm thanks for your review and suggestions. As you suggested I have removed `getProviderFromOptionsOrWindow` and added the default `getProvider` to the options. But I needed to add an extra interface to make `getProvider` not optional in `this.options`.

Also using `getProvider` in the `MetaMaskConnector` connector. 

Please check if it's what you like.

tmm:
Created new PR #1406 to continue work since I wasn't able to make updates to this one.

PR #1386, opened 2022-11-28
feat: pass wagmi `storage` to `createSyncStoragePersister`

### LINKED ISSUE #1383: bug: window.localStorage still being called when custom "storage" property passed to createClient

### Package Version

0.8.6

### Current Behavior

I'm trying to use wagmi while in an iframe. Because wagmi's createClient method seems to call localStorage no matter which, I always get an error. This is because in an iframe, there is no access to localstorage. (Unless both a) iframe is loaded on page of same origin and b) 'allow-same-origin' is set on iframe. Neither of these things are possible in my case. 

I tried passing in a custom `storage` property into createClient, like so: 
```
const wagmiClient = createClient({
  connectors,
  provider,
  storage: myCustomStorage
});
```
Even so, localStorage is still called up. This breaks my application in iframes.

### Expected Behavior

Local storage is never called when I run createClient and use a custom storage object. 

### Steps To Reproduce

_No response_

_No response_

### Anything else?

_No response_

jxom:
Fixes #1383

## Description

This PR resolves #1383 and passes the wagmi `storage` config option down to the React Query persister (`createSyncStoragePersister`). Previously, the React Query persister was defaulting to `window.localStorage`. 

Also hoisted up wagmi's serialization/deserialization fns to `createStorage` instead of deferring it to the persister. This also fixes a bug with our existing persist mechanisms with Zustand where we would "double serialize" values.

## Additional Information

Your ENS/address: jxom.eth


packages/core/package.json
```diff
@@ -101,7 +101,9 @@
   },
   "devDependencies": {
     "@coinbase/wallet-sdk": "^3.4.1",
+    "@types/react": "^18.0.9",
     "@walletconnect/ethereum-provider": "^1.7.8",
+    "react": "^18.1.0",
```
jxom, line 106: Just dependencies for testing serializing `React.createElement`.

tmm: Don't love this, but it's a dev dependency. Could add a step to CI that [removes dev dependencies](https://github.com/pnpm/pnpm/issues/5695) so they don't show up on npm with `pnpm pkg delete devDependencies` (not super pressing though).

tmm (approved):

tmm (approved):

PR #1422, opened 2022-12-04
feat(docs): Add more info on how to structure the TypedData

ricardo-passos:
## Description

When signing data using Ethers.js `signer._signTypedData` function, we SHOULD not include the `EIP712Domain` property inside `types` object as Ethers.js already do it for us. Doing so will cause an exception when calling the `signTypedData`/ `signTypedDataAsync` functions.

```ts
// bad code

const domain = {
  // ...props
} as const
const types = {
  EIP712Domain: [
      { name: 'name', type: 'string' },
      { name: 'version', type: 'string' }
  ],
  VerifyMessage: [
    { name: 'message', type: 'string' },
    { name: 'wallet', type: 'address' }
  ]
} as const
const value = {
  // ...props
} as const
```

```ts
// good code

const domain = {
  // ...props
} as const
const types = {
  VerifyMessage: [
    { name: 'message', type: 'string' },
    { name: 'wallet', type: 'address' }
  ]
} as const
const value = {
  // ...props
} as const
```

I think this information should be explicitly included in the docs.

## Additional Information

Your ENS/address: N/A

### DIFF
--- a/docs/pages/react/hooks/useSignTypedData.en-US.mdx
+++ b/docs/pages/react/hooks/useSignTypedData.en-US.mdx
@@ -124,6 +124,13 @@ function App() {
 
 ### types (optional)
 
+<Callout type="warning">
+  Since we use Ethers.js under the hood, you SHOULD not include EIP712Domain in
+  the types object as Ethers.js already include it by default. Doing so will
+  cause an exception when trying to sign the message. [Click here for more
+  details](https://github.com/ethers-io/ethers.js/issues/687#issuecomment-1209458789).
+</Callout>
+
 Typed data type definition.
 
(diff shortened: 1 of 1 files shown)


jxom:
Thoughts on just omitting the `EIP712Domain` attribute internally before we send it over to ethers? Means that we still stick to EIP-712 as well as still being interoperable w/ `recoverTypedSignature` and other utils.

ricardo-passos:
@jxom Since I don't know Wagmi's internals, I'm not sure what the best approach would be. I think the notice in the documentation should be enough.

jxom:
Closing in favor of #1437. Thanks for the PR!

PR #1423, opened 2022-12-05
feat: add `getClient` export + internal test exports

jxom:
## Description

Adding an export for `getClient`, as well as an internal test entrypoint, which will be used for the upcoming `@wagmi/connectors` package.

## Additional Information

Your ENS/address: jxom.eth


tmm (approved):
Overall looks good. Just one comment.

One thing we should think about is how much we want to expose publicly at an entrypoint (via `packages/*/src/test`) versus what we should keep private as specific to our testing set up (`packages/*/test`). For example, `getSigners` isn't a super robust public API yet.

PR #1439, opened 2022-12-08
feat: display .bit name & avatar

renzholy:
## Description

Discussion: #1438 

## Progress

- [x] core
- [x] react
- [x] hooks
- [ ] doc

## Additional Information

Your ENS/address:
aliez.eth


iam-frankqiu:
I think this feature is really necessary and important. I fully support it.

tmm:
Thanks for putting in this effort. Going to close for now and we should discuss whether this should be included in https://github.com/wagmi-dev/wagmi/discussions/1438.

IMO the best approach for this type of work is to keep wagmi lightweight and make this a separate wagmi-compatible [community package](https://github.com/wagmi-dev/awesome-wagmi) maintained by the .bit core team.

gakonst:
+1, ideally Wagmi is as unopinionated as possible while exposing clean interfaces for people to integrate. 

@tmm maybe worth making this explicit in your contribution docs. e.g. can set the expectation that to accept a PR the discussion should have some kind of buy-in from the maintainers. This also is in the interest of respecting the time of contributors like @renzholy and not having them to do redundant work.

PR #1444, opened 2022-12-08
fix: assert `connector` exists in `watchSigner`

### LINKED ISSUE #1440: bug: chain switching: unsupported chains, race condition

### Package Version

0.8.10

### Current Behavior

You have a connector with single supported chain (ex. Metamask, Goerli)
You connect to Metamask via useConnect().connect()

1. You are switching active chain to Mainnet inside Metamask. useNetwork().id equals 1, however provider is following: 
image
More over library also generates Signer in this situation.

Iam trying to solve this problem. Iam tracking useNetwork().id and every time it changes and chain is not supported, and call useDisconnect().disconnect()

```
  useEffect(() => {
    if (isUnssupported) {
     disconnect()
    }
  }, [chainId]);
```
2. In this case you get race condition between disconnect() call, and internal wagmi handler called on chain change. The slowest wins. Result: useProvider() returns jsonRpcProvider, and after that jsonRpcProvider  is being replaced by Web3Provider + Signer.

### Expected Behavior

1. Library should not generate Web3Provider and Signer for unsupported networks, it should generate jsonRpcProvider instead for some supported chain (previous used/most priority).

2. This race condition should never exist: you simply CANNOT change chain if something is ALREADY disconnected. So after disconnection, further onChainSwitch should be ignored

### Steps To Reproduce

```
const { chains, provider, webSocketProvider } = configureChains(
  [chain.goerli],
  [publicProvider()],
);

const metaMask = new MetaMaskConnector({
  chains,
});

const client = createClient({
  autoConnect: true,
  provider,
  connectors: [metaMask],
});

render(
  <React.StrictMode>
            <WagmiConfig client={client}>
                     <App />
             </WagmiConfig>
  </React.StrictMode>,
  document.getElementById("root"),
);

function App(): React.ReactElement {
  const { connect, connectors } = useConnect();
  const { disconnect } = useDisconnect();
  const provider = useProvider();
  const { data: signer } = useSigner();

  const { chain } = useNetwork();
  const { id: chainId } = chain || {};

  useEffect(() => {
    console.log("PROVIDER", provider);
  }, [provider]);
  useEffect(() => {
    console.log("SIGNER", signer);
  }, [signer]);

  useEffect(() => {
    if (chainId !== chain.goerli.id) {
     disconnect()
    }
  }, [chainId]);

  return (
    connectors.map((connector, id) => <button key={id} onClick={() => connect(connector)}>CONNECT</ button>})
  );
}
```

_No response_

### Anything else?

Any suggestions how to solve this problem are welcome.

jxom:
## Description

Fixes #1440 

Asserts that a `connector` exists before invoking the callback in `watchSigner`.

## Additional Information

Your ENS/address: jxom.eth

### DIFF
--- a/packages/core/src/actions/accounts/watchSigner.ts
+++ b/packages/core/src/actions/accounts/watchSigner.ts
@@ -16,8 +16,11 @@ export function watchSigner<TSigner extends Signer = Signer>(
   callback: WatchSignerCallback<TSigner>,
 ) {
   const client = getClient()
-  const handleChange = async () =>
-    callback(await fetchSigner<TSigner>({ chainId }))
+  const handleChange = async () => {
+    const signer = await fetchSigner<TSigner>({ chainId })
+    if (!getClient().connector) return callback(null)
+    return callback(signer)
+  }
   const unsubscribe = client.subscribe(
     ({ data, connector }) => ({
       account: data?.account,
--- a/.changeset/stale-onions-smell.md
+++ b/.changeset/stale-onions-smell.md
@@ -1,0 +1,6 @@
+---
+'@wagmi/core': patch
+'wagmi': patch
+---
+
+Assert that a `connector` exists before invoking the callback in `watchSigner`.
--- a/packages/core/src/actions/accounts/watchSigner.test.ts
+++ b/packages/core/src/actions/accounts/watchSigner.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, it } from 'vitest'
 
 import { setupClient } from '../../../test'
+
 import { connect } from './connect'
 import { disconnect } from './disconnect'
 import { switchNetwork } from './switchNetwork'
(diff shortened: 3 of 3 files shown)


packages/core/src/actions/accounts/watchSigner.ts
```diff
@@ -16,8 +16,11 @@ export function watchSigner<TSigner extends Signer = Signer>(
   callback: WatchSignerCallback<TSigner>,
 ) {
   const client = getClient()
-  const handleChange = async () =>
-    callback(await fetchSigner<TSigner>({ chainId }))
+  const handleChange = async () => {
+    const signer = await fetchSigner<TSigner>({ chainId })
+    if (!getClient().connector) return callback(null)
```
jxom, line 21: We need to get the client dynamically here, so can't use the client above (L18).

tmm: Don't we already check that the connector exists inside [`fetchSigner`](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/actions/accounts/fetchSigner.ts#L15)?

PR #1447, opened 2022-12-09
fix: ensure that `useSigner` does not broadcast unnecessarily

jxom:
## Description

Addresses https://github.com/wagmi-dev/wagmi/issues/987#issuecomment-1343421969

Seems like `staleTime` doesn't seem to work as intended with `setQueryData`, and can sometimes broadcast to other instances of `useSigner`. Just refactoring our `useSigner` hook to handle invalidaton all in `useQuery`.

## Additional Information

Your ENS/address: jxom.eth

### DIFF
--- a/packages/core/src/actions/accounts/getAccount.ts
+++ b/packages/core/src/actions/accounts/getAccount.ts
@@ -22,8 +22,8 @@ export type GetAccountResult<TProvider extends Provider = Provider> =
       status: 'reconnecting'
     }
   | {
-      address: undefined
-      connector: undefined
+      address: Data<TProvider>['account']
+      connector: Client<TProvider>['connector']
       isConnected: false
       isReconnecting: false
       isConnecting: true
@@ -67,8 +67,8 @@ export function getAccount<
       } as const
     case 'connecting':
       return {
-        address: undefined,
-        connector: undefined,
+        address: data?.account,
+        connector,
         isConnected: false,
         isConnecting: true,
         isDisconnected: false,
--- a/packages/react/src/hooks/accounts/useSigner.ts
+++ b/packages/react/src/hooks/accounts/useSigner.ts
@@ -1,9 +1,10 @@
 import type { FetchSignerArgs, FetchSignerResult, Signer } from '@wagmi/core'
-import { fetchSigner, watchSigner } from '@wagmi/core'
-import * as React from 'react'
+import { fetchSigner } from '@wagmi/core'
 
 import type { QueryConfig, QueryFunctionArgs } from '../../types'
-import { useChainId, useQuery, useQueryClient } from '../utils'
+
+import { useChainId, useQuery } from '../utils'
+import { useAccount } from './useAccount'
 
 export type UseSignerConfig = Omit<
   QueryConfig<FetchSignerResult, Error>,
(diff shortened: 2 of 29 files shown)


packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -28,28 +32,22 @@ export function useSigner<TSigner extends Signer>({
...
 }: UseSignerConfig = {}) {
+  const { connector } = useAccount()
   const chainId = useChainId({ chainId: chainId_ })
   const signerQuery = useQuery<
     FetchSignerResult<TSigner>,
     Error,
     FetchSignerResult<TSigner>,
     ReturnType<typeof queryKey>
-  >(queryKey({ chainId }), queryFn, {
+  >(queryKey({ chainId, connectorId: connector?.id }), queryFn, {
```
tmm, line 42: Will `connectorId` be unique enough for when people use multiple Connectors (e.g. connect modal libraries)? We only allow one "active" Connector right now in the wagmi Client, but maybe in the future we can allow multiple - does that break this assumption?

packages/react/src/hooks/accounts/useSigner.ts
```diff
@@ -28,28 +32,22 @@ export function useSigner<TSigner extends Signer>({
...
 
-  const queryClient = useQueryClient()
-  React.useEffect(() => {
-    const unwatch = watchSigner({ chainId }, (signer) =>
-      queryClient.setQueryData(queryKey({ chainId }), signer),
-    )
-    return unwatch
-  }, [queryClient, chainId])
-
   return signerQuery
```
tmm, line 52: Can we move the `return` up to the same line as `useQuery(…)` now?

packages/react/src/hooks/contracts/usePrepareContractWrite.test.ts
```diff
@@ -324,6 +324,7 @@ describe('usePrepareContractWrite', () => {
       config[property as keyof typeof config] =
         baseConfig[property as keyof typeof baseConfig]
       rerender()
+      await actConnect({ utils })
```
tmm, line 327: Interesting. Any idea why we need to call `actConnect` again now? Shouldn't the Client already be connected?

packages/react/src/hooks/accounts/useSigner.ts
jxom: True. Might be best if we can make this more unique. Will fix!

packages/react/src/hooks/contracts/usePrepareContractWrite.test.ts
jxom: Yeah, it's a weird one. The client is already connected, and I even tested this E2E, but not sure why it isn't working here. Might have another look.

packages/core/src/actions/accounts/getAccount.ts
```diff
@@ -67,8 +67,8 @@ export function getAccount<
       } as const
     case 'connecting':
       return {
-        address: undefined,
-        connector: undefined,
+        address: data?.account,
```
jxom, line 70: Noticed another weird behavior where the account & connector would be `undefined` if you are connecting to another connector.

Current behavior: If you are connected via Injected, then request connection to WalletConnect, the account + connector would be `undefined` (`connecting` status).
Expected behavior: If you are connected via Injected, then request connection to WalletConnect, the account + connector shouldn't change until you are either connected to the requested connector, or disconnected

tmm: Makes sense. What is `connector` (below on line 71) set to in this case? For example, I connect with MM and now request a connection to WC: Is `address`, the currently connected MM address and `connector`, the pending WC connector?

tmm (approved):

PR #1494, opened 2022-12-14
build: generate proxy packages

tmm:
## Description

Generate proxy packages during build process

## Additional Information

Your ENS/address: awkweb.eth


scripts/tsup.ts
```diff
@@ -52,7 +53,8 @@ export function getConfig({
             `export * from '${srcTypesFile}'`,
           )
         }
-        validateExports(exports)
+        await validateExports(exports)
+        await generateProxyPackageEntrypoints(exports)
```
tmm, line 57: Will check, but don't think we will need to do this during development (so can remove this line).

jxom (approved):

PR #1539, opened 2022-12-21
feat: support custom ensAddress for providers

0xFlicker:
## Description

Using the full `Networkish` type in providers' instantiation allows custom `ensAddress` to be applied from chain definitions.

### Background: 

Sepolia does not have an official ENS registry deployed. Since I am working on some ENS frontend tooling and want to use a testnet, and I prefer to use sepolia for reasons, I deployed my own suite of ENS contracts. I then update the sepolia chain definition before creating the providers like so....

```js
import { mainnet, goerli, sepolia } from "@wagmi/chains";

export const supportedAppChains = [mainnet, goerli, sepolia] as const;

sepolia.contracts = {
  ...sepolia.contracts,
  ensRegistry: {
    address: "0xBc4693CB0F572b99f6aB4462602b5A0B8585010B",
  },
};
```

We can then create our providers and init wagmi....

```js
  const { providers, webSocketProviders } = configureChains( [
    infuraProvider({
      apiKey: infuraKey.get(),
    }),
    alchemyProvider({
      apiKey: alchemyKey.get(),
    }),
    publicProvider(),
  ], supportedAppChains);
  return createClient({
    connectors: appConnectors.get(),
    provider,
    webSocketProvider,
    autoConnect: true,
  });
```

The problem I ran into was that the ensRegistry address was not being passed down into the providers. So I was still unable to load ENS details. This change should pass through ens registry address into providers.

## Additional Information

Your ENS/address: 0xflick.eth

### DIFF
--- a/.changeset/rude-phones-dress.md
+++ b/.changeset/rude-phones-dress.md
@@ -1,5 +1,0 @@
----
-'@wagmi/core': minor
----
-
-Pass ENS registry address of chain to providers
--- a/.changeset/rude-phones-dress.md
+++ b/.changeset/rude-phones-dress.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/core': minor
+---
+
+Pass ENS registry address of chain to providers
--- a/.changeset/dry-seals-raise.md
+++ b/.changeset/dry-seals-raise.md
@@ -1,5 +1,6 @@
 ---
-'wagmi': minor
+'@wagmi/core': patch
+'wagmi': patch
 ---
 
-Support custom ENS registries
+All Providers (ie. Alchemy, Infura, Public) now use the ENS Registry address on the wagmi `Chain` object (`chain.contracts.ensRegistry`).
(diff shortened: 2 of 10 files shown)


jxom (approved):

.changeset/rude-phones-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@wagmi/core': minor
```
tmm, line 2: We should add `wagmi` to this changeset too.

tmm (changes requested):

.changeset/rude-phones-dress.md
0xFlicker: done. Also snuck in some tests.

0xFlicker:
I don't think these test failures are me. Looks like a CI / timing issue. I sometimes see these locally but running the test individually seems fine.

PR #1570, opened 2022-12-26
feat: `watchPendingTransactions` + `useWatchPendingTransactions`

jxom:
Co-Authored-By: wslyvh <wslyvh@users.noreply.github.com>

Original PR: #1536 

## Description

Add a `watchPendingTransactions` action that subscribes to `pending` transactions as described in #1535 
I still need to add documentation, but wanted to create the PR for feedback

## Additional Information

Your ENS/address:
wslyvh.eth


packages/react/src/hooks/transactions/useWatchPendingTransactions.ts
```diff
@@ -0,0 +1,36 @@
...
+  chainId?: number
+  /** Subscribe to changes */
+  enabled?: boolean
+  /** Function fires when a pending transaction enters the mempool. */
+  listener: (transaction: Transaction) => void
+}
+
+export function useWatchPendingTransactions({
+  chainId: chainId_,
+  enabled = true,
```
wslyvh, line 18: this option seems redundant. You either subscribe to to events by using the hook or you don't.

wslyvh (commented):
Curious to hear your thoughts around the use of hooks vs. actions? Are hooks separated just to follow hooks rules or is there any other benefit from separating the 2?

For most cases there's quite some overlap, but often (like this PR/pendingTransactions) it's not really using/referencing the same code.

packages/react/src/hooks/transactions/useWatchPendingTransactions.ts
jxom: There may be a case where a consumer wants to disable subscriptions based on a depdendant state variable. wdyt?

jxom:
Generally would like the Hooks to reach parity with the core Actions. Reason why they don't share the same code is because we'd like to leverage the React lifecycle rather than relying on Zustand's reactivity (in core).

packages/react/src/hooks/transactions/useWatchPendingTransactions.ts
wslyvh: Sure, yeah. That makes sense

wslyvh:
Got it. Thanks! 

LGTM then :)

.changeset/large-steaks-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@wagmi/core': minor
+---
+
+Add watchPendingTransactions
```
tmm, line 5: Thoughts on this just being a `patch`?

```suggestion
Added `watchPendingTransactions`
```

docs/pages/core/actions/_meta.en-US.json
```diff
@@ -35,6 +35,7 @@
   "watchReadContract": "watchReadContract",
   "watchReadContracts": "watchReadContracts",
   "watchSigner": "watchSigner",
+  "watchPendingTransactions": "watchPendingTransactions",
```
tmm, line 38: should make this alphabetized

docs/pages/core/actions/watchPendingTransactions.en-US.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: 'watchPendingTransactions'
+description: 'Action for subscribing to pending transactions.'
+---
+
+import { Callout } from 'nextra-theme-docs'
+
+# watchPendingTransactions
+
+Action for subscribing to pending transactions in the Provider's transaction mempool. [What is a mempool?](https://www.alchemy.com/overviews/what-is-a-mempool)
```
tmm, line 10:
```suggestion
Action for subscribing to pending transactions in the Provider's transaction mempool. See ["What is a mempool?"](https://www.alchemy.com/overviews/what-is-a-mempool) for more info.
```

docs/pages/react/hooks/useWatchPendingTransactions.en-US.mdx
```diff
@@ -0,0 +1,71 @@
+---
+title: 'useWatchPendingTransactions'
+description: 'Action for subscribing to pending transactions.'
+---
+
+import { Callout } from 'nextra-theme-docs'
+
+# useWatchPendingTransactions
+
+Hook for subscribing to pending transactions in the Provider's transaction mempool. [What is a mempool?](https://www.alchemy.com/overviews/what-is-a-mempool)
```
tmm, line 10:
```suggestion
Hook for subscribing to pending transactions in the Provider's transaction [mempool](https://www.alchemy.com/overviews/what-is-a-mempool).
```

tmm (commented):
Couple suggestions, but overall lgtm

tmm:
Needs a changeset for `wagmi` too!

PR #1593, opened 2023-01-01
feat: cjs escape hatch

jxom:
## Description

Added a script to generate & release a CJS escape hatch bundle.

Deliberately chose to create a monolithic script in isolation so we can easily remove it when the time comes. But if we ever want to modularlize this into separate scripts, we can do that too.

## Additional Information

- [x] I read the [contributing docs](/wagmi-dev/references/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address: jxom.eth


.prettierignore
```diff
@@ -8,4 +8,6 @@ CHANGELOG.md
 docs/pages/index.en-US.mdx
 docs/pages/react/getting-started.en-US.mdx
 docs/pages/core/getting-started.en-US.mdx
+docs/pages/react/module-types.en-US.mdx
+docs/pages/core/module-types.en-US.mdx
```
tmm, line 12: I should of done this for the getting started pages lol

```suggestion
docs/pages/*/module-types.en-US.mdx
```

tmm (approved):
Looks good! Could use [`pnpm pkg ...`](https://github.com/pnpm/pnpm/issues/5695) commands to manipulate `package.json` in-place if you wanted to, but probably makes more sense to atomically like you have it now. Nice work!

Issue #1565, opened 2022-12-24
bug: awaiting switchNetwork before sendTransaction throws error


jmrossy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.9.5

### Current Behavior

The `@wagmi/core` `switchNetwork` function does successfully switch the network for most wallets, but subsequent calls to `sendTransaction` will still throw a `ChainMismatchError` even after the switch.

I guess there's stale data affecting `sendTransaction`.

### Expected Behavior

`SendTransaction` calls after awaiting `switchNetwork` succeed.

### Steps To Reproduce

```js
await switchNetwork({ chainId: 1 });
await sendTransaction({
    chainId: 1,
    request: myTxRequest,
    mode: 'recklesslyUnprepared',
});
```

jmrossy:
Note, a workaround of `await sleep(1000);` to stall for a second or two seems to solve the issue, implying there's some kind of race condition involved.

Nazeeh21:
Can I work on this issue?

legendarykamal:
Hi , can I also contribute in solving this problem ?

jxom:
PRs are welcome! Otherwise, I can pick this up when I get some spare time. Just reply to this thread so others know you are working on it.

Nazeeh21:
Cool, I'll start working on it soon

Nazeeh21:
> PRs are welcome! Otherwise, I can pick this up when I get some spare time. Just reply to this thread so others know you are working on it & so I can assign you to the issue.

I'd like to work on this, can you please assign it to me?

DanInTheD4rk:
I'll take a crack at this if no one is working it

tmm:
@DanInTheD4rk sweet! will assign it to you.

jxom:
Fixed on latest `wagmi` & `@wagmi/core` versions! Thanks @DanInTheD4rk.

PR #1595, opened 2023-01-02
Fix issue #1565

Nazeeh21:
## Description
This PR fixes issue #1565

The `@wagmi/core` `switchNetwork` function does successfully switch the network for most wallets, but subsequent calls to `sendTransaction` will still throw a `ChainMismatchError` even after the switch.

Added `sleep` function after changing network.
Lmk if any changes are required

## Additional Information

Your ENS/address: nazeeh.eth


jxom:
Thanks for the PR! 

Wondering if we could find and fix the underlying cause of the race condition described in #1565 rather than mitigating it with a `sleep`/`setTimeout` fn?

Also, seems that this PR only modifies our `references/` submodule – it might be best to re-open this PR at https://github.com/wagmi-dev/references if that's the case :)

Nazeeh21:
Sure thing! I'll try finding a fix for the race condition and then will create a PR directly to the `references` submodule

jxom:
Thank you! I am going to close this one here for now :)

PR #1638, opened 2023-01-09
feat(solid): Implement solid playground and first account hooks

NeOMakinG:
## Description

- Added a new package called `@wagmi/solid` containing reactivity for solid and some account hooks
- Added a new playground using solid base starter in order to test solid hooks

⚠️ We encountered an error while using solid start, not happening when using the default solid starter:

`TypeError: Class extends value undefined is not a constructor or null at /references/packages/connectors/src/base.ts24:10`

After investigating, it looks like `EventEmitter` from `connectors/base.ts` is undefined while using SSR with solid-start on the nodejs side, we actually can't figure out why it's working with next for the moment but not with solid start, so we did switch to the regular solid starter using vite 4 and it works perfectly

@gabrieldemian 

## Current core hooks implemented

- [x] useAccount
- [x] useConnect
- [x] useContract
- [x] useDisconnect
- [x] useNetwork
- [x] useProvider
- [x] useSigner
- [x] useSwitchNetwork
- [x] useWebSocketProvider

## Additional Information

Your ENS/address:
0xm4king.eth / 0xfd42860cD619FE2060712DC5833F0B039b23e00a

### DIFF
--- a/docs/pages/core/getting-started.en-US.mdx
+++ b/docs/pages/core/getting-started.en-US.mdx
@@ -57,7 +57,6 @@ Next, create a wagmi `Client` instance using [`createClient`](/core/client), and
 
 ```ts {14-18}
 import {
-  WagmiConfig,
   createClient,
   configureChains,
   mainnet,


gabrieldemian:
I did a test, instead of using solid-start, I used vanilla solid, and the error was gone, working normally. I also did a `console.log(EventEmitter)`. And it was undefined on the solid-start server, so I guess the problem is with the server side rendering of solid-start.

tmm:
Excited to see this! Some early feedback:

> Added a new package called wagmi-solid containing reactivity for solid and some account hooks

Let's scope it under the `wagmi` npm org and call it `@wagmi/solid`

> Changed the name package of the react hooks to wagmi-react because cannot longer call it just wagmi

Should be fine to keep as just `wagmi`. Maybe at some point we can move to `@wagmi/react`, but since it is the biggest consumer so far, let's keep as-is.

> TypeError: Class extends value undefined is not a constructor or null at /references/packages/connectors/src/base.ts24:10

Don't think this is related to a circular dependency. Seems similar to an issue we had with Vite and the `CoinbaseWalletConnector` a little while ago (see #693). Since SolidStart uses Vite, this seems like a good place to start.

---

In terms of getting this ready to merge, here's how I would prioritize the hooks (along with tests/docs):

### Required

Goal: Allows you to perform all basic functionality (connect wallets, access connected account, perform contract reads/writes in a type-safe, non-declarative manner).

- useAccount
- useConnect
- useContract
- useDisconnect
- useNetwork
- useProvider
- useSigner
- useSwitchNetwork
- useWebSocketProvider

### Nice-to-have

Goal: Extend the basic functionality to higher-level, declarative APIs.

- useBalance
- useBlockNumber
- useContractRead
- useContractReads
- useContractWrite
- useEnsAddress
- useEnsName
- usePrepareContractWrite
- usePrepareSendTransaction
- useSendTransaction
- useSignMessage
- useSignTypedData
- useWaitForTransaction

### Later

Goal: Map the remaining API surface area for lesser-used hooks.

- useContractEvent
- useContractInfiniteReads
- useEnsAvatar
- useEnsResolver
- useFeeData
- useToken
- useTransaction
- useWatchPendingTransactions

cc @jxom

NeOMakinG:
@gabrieldemian We should probably rename every hooks to `create` instead of `use` as it's a solid standard wdyt?

jxom:
Wouldn't stress too much on renaming "use" to "create" – might conflict with a lot of our naming conventions (ie. `useTransaction` would become `createTransaction`).

Solid Router still uses the ["use" prefix](https://github.com/solidjs/solid-router#useparams) too.

NeOMakinG:
> "use" prefix

As you wish sir!

NeOMakinG:
> Wouldn't stress too much on renaming "use" to "create" – might conflict with a lot of our naming conventions (ie. `useTransaction` would become `createTransaction`).
> 
> Solid Router still uses the ["use" prefix](https://github.com/solidjs/solid-router#useparams) too.

As you wish sir!

gabrieldemian:
I'm trying to test the solid hooks and having some trouble with `solid-testing-library`. Created an issue on their repo https://github.com/solidjs/solid-testing-library/issues/31

PR #1679, opened 2023-01-17
Use useEffect to avoid unsafe side effects for useAccount callbacks

aj-may:
## Description

When using the `onConnect` and `onDisconnect` callbacks in an app with react strict mode on, the current implementation ends up causing the app state to become unstable.  This has something to do with the double rendering of components when a react app, in strict mode, is running the dev server.

## Additional Information

Your ENS/address:
ajmay.eth

## Related Issues
I believe these issues are related to this fix:
- https://github.com/rainbow-me/rainbowkit/issues/837
- https://github.com/rainbow-me/rainbowkit/issues/836
- https://github.com/rainbow-me/rainbowkit/pull/891

As these issues usually occur after an `onConnect` or `onDisconnect` callback is called and when logging the state, you can see that it reverts on the second render.


aj-may:
@jxom @tmm would be happy to get your thoughts on this one.  I got the tests to run locally.  I'm getting 3 errors but none related to this functionality.  Also, tested in the playground and all seems to be fine.

bounds:
Very excited to not need to remove strict mode from projects where rainbow kit is used

tmm:
Appreciate the patience! Will have some time to take a look soon.

tmm:
@aj-may spent some time looking into this and it doesn't seem like anything is wrong here. The dev playground has [used strict mode](https://github.com/wagmi-dev/wagmi/blob/main/examples/_dev/next.config.js#L8) since April 2022. Can you describe some more what you mean by "app state to become unstable"? I'm seeing `onConnect` and `onDisconnect` callbacks fire appropriately in the dev playground and test suite.

(Side note: The issues you linked for RainbowKit are definitely issues, but quickly poked around there and they seem unrelated to wagmi.)

aj-may:
@tmm Ahh, I don't think this bug will rear its head unless there is a component that is using these callbacks to modify its own state.  in that case the dependent components state becomes unstable, or reverts to its previous state on the second render under react strict mode.

Let me see if I can make a quick branch to add something to the playground that can replicate the bug.

aj-may:
@tmm I was able to add an example to reproduce here: https://github.com/aj-may/wagmi/tree/side-effect-playground

aj-may:
@tmm Have you had a chance to pull down the branch I created that reproduces the issue?   Would love if we can revisit this and potentially merge this in.

packages/react/src/hooks/accounts/useAccount.ts
```diff
@@ -21,32 +21,31 @@ export type UseAccountConfig = {
...
   if (
     !!onDisconnect &&
-    previousAccount.status === 'connected' &&
+    previousStatus === 'connected' &&
     account.status === 'disconnected'
   )
     onDisconnect()
 
-  previousAccountRef.current = account
+  if (account.status !== previousStatus) setPreviousStatus(account.status)
```
tmm, line 48: `previousStatus` is reactive allowing it to rerender hook.

aj-may:
@tmm The changes you pushed cause the connect and disconnect callbacks to be called twice in dev mode with react strict mode on and still shows some state fluctuations over the multiple renders in the console.

Screenshot 2023-01-30 at 13 34 30
Screenshot 2023-01-30 at 13 34 59

aj-may:
Prior to that the console looked like this:
Screenshot 2023-01-30 at 13 36 02
Screenshot 2023-01-30 at 13 36 23

jxom (approved):
makes sense

aj-may:
👍 I'll take this for a spin in a couple hours

aj-may:
Looking good!

image

PR #1726, opened 2023-01-26
Add automatic foundry.toml detection

roninjin10:
## Description

Adds automatic detection of wagmi artifacts directory if one is not provided

## Additional Information

I made it synchronous for now because the outfiles are synchronously returned by the plugin

Your ENS/address: weicurry.eth

### DIFF
--- a/packages/cli/src/plugins/foundry.ts
+++ b/packages/cli/src/plugins/foundry.ts
@@ -5,6 +5,9 @@ import { globby } from 'globby'
 import { basename, extname, resolve } from 'pathe'
 import pc from 'picocolors'
 
+import toml from 'toml'
+import { z } from 'zod'
+
 import type { ContractConfig, Plugin } from '../config'
 import * as logger from '../logger'
 import type { RequiredBy } from '../types'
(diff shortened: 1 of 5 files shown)


packages/cli/src/plugins/foundry.ts
```diff
@@ -83,23 +86,65 @@ type FoundryConfig<TProject extends string> = {
...
+    forge: {
+      clean = false,
+      build = true,
+      path: forgeExecutable = 'forge',
+      rebuild = true,
+    } = {},
+    include = ['*.json'],
+    namePrefix = '',
+    project,
+  } = config
```
roninjin10, line 143:
```suggestion
export function foundry<TProject extends string>({
  artifacts = 'out',
  deployments = {},
  exclude = defaultExcludes,
  forge: {
    clean = false,
    build = true,
    path: forgeExecutable = 'forge',
    rebuild = true,
  } = {},
  include = ['*.json'],
  namePrefix = '',
  project,
}: FoundryConfig<TProject>): FoundryResult {
```

Looking at this now, this didn't need to change.  This is just a relic from my initial implementation

PR #1736, opened 2023-01-26
fix(cli): generated chain key type

tmm:
## Description

Fix generated address object key type so it's alway a number. (Won't be a number in JS `Object.keys` still, which is fine. Just need it to be for type inference.)

## Additional Information

Your ENS/address: awkweb.eth

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,11 +2,11 @@
   "name": "root",
   "scripts": {
     "anvil": "anvil --fork-url $ANVIL_FORK_URL --fork-block-number 15578840",
-    "build": "pnpm -r --filter '@wagmi/chains' build && pnpm -r --filter '@wagmi/core' build && pnpm -r --filter './packages/**' --filter './references/packages/**' build",
+    "build": "pnpm --filter '@wagmi/chains' build && pnpm --filter '@wagmi/core' build && pnpm -r --filter './packages/**' --filter './references/**' --filter '!./packages/core' --filter '!./references/**/chains' build",
     "changeset:release": "pnpm build && changeset publish",
     "changeset:version": "changeset version && pnpm install --lockfile-only",
     "cjs:release": "node -r esbuild-register ./scripts/cjs.ts",
-    "dev": "pnpm -r --filter '@wagmi/chains' build && pnpm -r --filter '@wagmi/connectors' --filter './packages/**' dev",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -2,11 +2,11 @@
   "name": "root",
   "scripts": {
     "anvil": "anvil --fork-url $ANVIL_FORK_URL --fork-block-number 15578840",
-    "build": "pnpm -r --filter '@wagmi/chains' build && pnpm -r --filter '@wagmi/core' build && pnpm -r --filter './packages/**' --filter './references/packages/**' build",
+    "build": "pnpm --filter '@wagmi/chains' build && pnpm --filter '@wagmi/core' build && pnpm -r --filter './packages/**' --filter './references/**' --filter '!./packages/core' --filter '!./references/**/chains' build",
     "changeset:release": "pnpm build && changeset publish",
     "changeset:version": "changeset version && pnpm install --lockfile-only",
     "cjs:release": "node -r esbuild-register ./scripts/cjs.ts",
-    "dev": "pnpm -r --filter '@wagmi/chains' build && pnpm -r --filter '@wagmi/connectors' --filter './packages/**' dev",
+    "dev": "pnpm --filter '@wagmi/chains' build && pnpm -r --filter './packages/**' --filter './references/**' --filter '!./references/**/chains' dev",
```
tmm, line 9: Cleaned up these scripts. Have some dependencies that pnpm is able to map out correctly with topo sort so handling it here.

packages/cli/package.json
```diff
@@ -68,7 +68,7 @@
   },
   "dependencies": {
     "@ethersproject/address": "^5.7.0",
-    "@wagmi/chains": "^0.2.0",
+    "@wagmi/chains": "workspace:../../references/packages/chains",
```
tmm, line 71: Use workspace version instead of pinned

packages/cli/src/commands/generate.ts
```diff
@@ -345,10 +342,7 @@ async function getContract({
       typeof resolvedAddress === 'string'
         ? JSON.stringify(resolvedAddress)
         : // Remove quotes from chain id key
-          JSON.stringify(resolvedAddress, null, 2).replace(
-            /^\s*"(\d)":/gm,
-            '$1:',
-          )
+          JSON.stringify(resolvedAddress, null, 2).replace(/"(\d*)":/gm, '$1:')
```
tmm, line 345: regex i used before only worked for single digits 🤦‍♂️

this one is much better: https://regexr.com/7700l

PR #1826, opened 2023-02-09
feat(core,react): ✨ `useNFT` hook

h0tw4t3r:
## Description

Adds `fetchNFT` method at `@wagmi/core` and `useNFT` hook at `wagmi`.

The params are pretty much the same as for `fetchToken` and `useToken` besides taking `id: string` as an input instead of `formatUnits: number` and returning `uri: string` instead of `decimals: BigInt...`

## Additional Information

I tried running the tests but don't have a reliable archive node by my hand to spin up the `anvil` locally.

So the Pull Request was not tested locally.

Your ENS/address:
h0tw4t3r.eth


h0tw4t3r:
CC @tmm @jxom

roninjin10:
This duplicates functionality from the erc20 plugin of the wagmi cli.   Have you tried it yet?

h0tw4t3r:
> This duplicates functionality from the erc20 plugin of the wagmi cli. Have you tried it yet?

Yes, I tried it!

I still think it is beneficial to have a `useERC721` and `useERC1155` hooks just for the same reason the `useToken` exists – for the sake of accessibility.

Indeed, it just adds up sugar code just as `useToken` does. 

> Small nit: I think it should be `useERC721` instead of `useNFT`

If so, it makes sense to rename the `useToken` hook to `useERC20` and create an `useERC1155` hook to cover all of the mostly used token interfaces.

In the end, this PR will be a feature for those who rather won't use `@wagmi/cli` 

--- 
Also, please consider the code:
https://github.com/wagmi-dev/wagmi/blob/cb707f01cbdcc62a70cf5c8a162d77948d6b6a56/packages/core/src/actions/contracts/fetchToken.ts#L38-L46

`allowFailure` is set to `false`, but in fact the `name`, `decimals` and `symbol` are considered to be optional parameters in EIP20 standard, and there could be edge cases where tokens without such properties won't be successfully fetched.

I'd propose to update/remove the `useToken` but that could be a bad backwards-compatibility choice, unless it will be included in the new major release. Please tell me if another PR is needed for that.

---
References:
- https://eips.ethereum.org/EIPS/eip-20

tmm:
Appreciate you working on this. It might make sense to add this at some point, but we aren't going to before `wagmi@1` and `@wagmi/core@1`. Going to close for now. If this stream of work is ever re-opened, will make sure your contribution history makes it in.

PR #1902, opened 2023-02-23
feat: WalletConnect v2 updates

jxom:
## Description

Dupe of https://github.com/wagmi-dev/wagmi/pull/1828

## Additional Information

Your ENS/address: asimetriq.eth


arindammojumder:
Package Version
@wagmi/core": "^0.10.10

**Current Behavior**
While using WalletConnectConnector of **"@wagmi/core": "^0.8.19"**, the below configuration was working fine with trustwallet app. The connection modal on the app was showing multichains providers. See old-version-wagmi image

```
this.walletConnector = new WalletConnectConnector({
      chains: [mainnet, bsc],
      options: {
        qrcode: true,
        version: '2',
        projectId: 'a325040d9331d6d603c4f7ae270fa5e5' // '7eeef33652f115b08e408e3a022a44d9
      },
    })
```
Unfortunately after upgradation to **@wagmi/core": "^0.10.10** THe wallet connector only connecting with the first chain only (ETH) and not BSC at the same time. See new-version-wagmi image
    ```
this.metaWalletConnector = new WalletConnectConnector({
      chains,
      options: {
        projectId: 'a325040d9331d6d603c4f7ae270fa5e5'
      },
    })
```

**Expected Behavior**
The latest version must show the mutichain while connecting, just like the old-version-wagmi. See old-version-wagmi image
old-version-wagmi
new-version-wagmi

jxom:
Not sure if this is a wagmi issue. I suspect that Trust Wallet has not implemented WalletConnect v2's optional chains API. It might be worth getting in touch with them.

jxom:
Or you can downgrade to `WalletConnectLegacyConnector`.

arindammojumder:
> Or you can downgrade to `WalletConnectLegacyConnector`.

That is not a solution especially when hey have updated the packages to v2.

jxom:
You might have to get in touch with Trust Wallet. :)

archie-wu-lb:
my package version  
viem: "^1.0.7",
wagmi: "^1.2.0"
I use imToken to connect my dapp
multi-chain can't work 
I only get one chain
and I can't change network 

but when i use imToken to connect [walletConnect demo ](https://react-app.walletconnect.com/)
it's work and I can change network

archieWu:
my package version
viem: "^1.0.7",
wagmi: "^1.2.0"
I use imToken to connect my dapp
multi-chain can't work
I only get one chain
and I can't change the network

but when I use imToken to connect [walletConnect demo](https://react-app.walletconnect.com/)
it's works and I can change the network

jxom:
imToken probably hasn't implemented WalletConnect v2 Optional Namespaces. You'll need to get in touch with them.

archieWu:
I think this is not an imToken problem
use [walletConnect demo](https://react-app.walletconnect.com/) it can work
Or am I missing something here
IMG_5952

arindammojumder:
Trust wallet is now working with multichain connection with wagmi latest version 1.2.0

archieWu:
this is work thanks

archieWu:
in my dapp more than one chain is required
can't we add more than one chain to required namespaces
not Optional Namespaces

jxom:
@archieWu – The issue there is that if the Wallet doesn't support the required namespace, then the connection will be rejected. Example: If you have Polygon as a required chain, and a user tries to connect with a Wallet that doesn't support it, then the user won't be able to connect to your DApp regardless if it is the targeted chain or not.

archieWu:
@jxom If this happens and users cannot connect to my DApp, it should be expected behavior because my DApp requires these chains. We should guide users toward a successful connection when a connection fails.
If these required chains are placed in the optional chain even if the user successfully connects to my DApp, some unexpected things will still happen.

archieWu:
if the wallet does not support one of the configured chains, the user can not get the experience we want to provide.
Assuming that I have created a Dapp for exchanging assets between two chains.
Then the user's wallet can not support one of the configured chains.
This is not unexpected. If the user uses this wallet this doesn't sound very good.

I know the required and optional chain setup we have now achieves parity with WalletConnect v1
but  WalletConnect v2 can not dynamic addition of chains.
This gives users a different experience in usage.
At V1, we can add chains anytime to provide users with a good experience.

but now we can not do this. If the user connects DApp it should approve all chains.
This may not be optional, it is necessary. but at Wagmi we only can set one chain as required.
we can not set all we need as required.

jxom:
> Then the user's wallet can not support one of the configured chains.

The preferred error here would be to display the error after connection, instead of on connection, right? Seems like a very abrupt user experience to reject the connection if one of the chains aren’t supported. Many DEXs support 10s of chains, and if one of those chains aren’t supported in the Wallet (most wallets really), then the connection will be rejected. 
Ideally, you’d want to display an error after the user has connected, but before they make the action, that the chain is not supported. ConnectKit does this quite well: https://docs.family.co/connectkit/try-it-out

> WalletConnect v2 can not dynamic addition of chains.

The lack of support of Dynamic Chain Authorization by Wallets is unfortunately a drawback of WalletConnect v2. We have been working closely with them to get Wallets to support this.

PR #1935, opened 2023-03-03
Fix useBalance select & keepPreviousData options

greg-schrammel:
## Description

fix `useBalance` select type and forward `keepPreviousData` option

## Additional Information

Your ENS/address: notgregs.eth

### DIFF
--- a/.changeset/twenty-turtles-dream.md
+++ b/.changeset/twenty-turtles-dream.md
@@ -1,0 +1,5 @@
+---
+'wagmi': patch
+---
+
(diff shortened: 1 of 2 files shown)


greg-schrammel:
`isDataEqual` and `structuralSharing` are other options that are on the type but not used, should we omit or forward?

tmm:
Thanks for looking into this! Right now, we don't expose most React Query options via the hooks so going to close in favor of https://github.com/wagmi-dev/wagmi/pull/2042. For `wagmi@1`, we'll loosen things up so you can pass any RQ option through.

Issue #2110, opened 2023-04-01
bug: can't use cli to generate hardhat code


Eyon42:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.1.14

### Current Behavior

Running `npm run wagmi generate` throws:

```

> frontend@0.1.0 wagmi
> wagmi

✔ Validating plugins
⠸ Resolving contracts
Command failed with exit code 1: npm hardhat compile
Unknown command: "hardhat"
```
 `npm hardhat compile` is invalid

### Expected Behavior

It should call `npx hardhat generate` or something else.

### Steps To Reproduce

Install the cli with npm, use the hardhat plugin and try to use the cli to generate.

### Anything else?

The issue is [here](https://github.com/wagmi-dev/wagmi/blob/main/packages/cli/src/plugins/hardhat.ts#L145):

```
      if (clean) {
        const packageManager = await getPackageManager();
        const [command, ...options] = (typeof clean === "boolean" ? `${packageManager} hardhat clean` : clean).split(" ");
        await execa2(command, options, { cwd: project });
      }
      if (build) {
        const packageManager = await getPackageManager();
        const [command, ...options] = (typeof build === "boolean" ? `${packageManager} hardhat compile` : build).split(" ");
        await execa2(command, options, { cwd: project });
      }
```

The `{packageManager} hardhat compile` patter does not work for npm, as it does for pnpm and yarn

Proposed solution: add a special case.
I might do it tomorrow, now I'm sleppy.

Also, `npm run wagmi init` does not work by default, I had to do something hacky and add `"wagmi": "wagmi"` to package.json

tomas-dukynas:
having the same issue

Eyon42:
Created a couple of PRs to address the issues

PR #2122, opened 2023-04-04
  Replace npm with npx for running hardhat commands in cli plugin.

Eyon42:
Solves #2110

## Description

Replace npm with npx for running hardhat commands.

Current behavior is calling the hardhat plugin with npm runs "npm hardhat compile" which is not a valid command. npx command is used, which is recommended by the hardhat docs.

## Additional Information

Your ENS/address: 0xc5a23c04f05D8F1bc27344E394654B21bE9CC488

### DIFF
--- a/packages/cli/src/plugins/hardhat.ts
+++ b/packages/cli/src/plugins/hardhat.ts
@@ -133,16 +133,18 @@ export function hardhat<TProject extends string>({
     async contracts() {
       if (clean) {
         const packageManager = await getPackageManager()
+        const packageManagerCommand = (packageManager === "npm") ? "npx" else packageManager
         const [command, ...options] = (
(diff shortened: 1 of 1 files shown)


tmm (changes requested):
Thanks for working on this!

Might be better to add a parameter to `getPackageManager` `executable?: boolean` (or similar) that returns the package manager executable CLI command (e.g. `npx`) so we can reduce some duplicate logic. For npm this would look like:

```ts
const packageManager = await getPackageManager()
//    ^^^^^^^^^^^^^^ "npm"
const packageManager = await getPackageManager(true)
//    ^^^^^^^^^^^^^^ "npx"
```

Once that is updated, you can add a patch changeset (`pnpm changeset`) and this is good to go!

tmm:
Thanks for working on this! Closing in favor of #2145 since I wasn't able to make changes here.

PR #2193, opened 2023-04-18
Update Sign-In with Ethereum Example 

obstropolos:
## Description
wagmi currently has an example for Sign-In with Ethereum that let's developers get started with authenticating their users. In this PR, we update that guide to use `ssx` ([documentation](https://docs.ssx.id)), an open-source / MIT-licensed library that enables developers to easily integrate Sign-In with Ethereum and more. Core to building SSX was making Sign-In with Ethereum easier to enable in applications, as well as add additional features such as signing in on behalf of a multisig and more. 

As part of this, the guide remains nearly identical, but the integration path is updated. 

## Additional Information
- [x] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution).
- [x] All updates to components and pages have been made.  
- [x] We have tested the pages locally to ensure they work. 
- [x] All steps in the relevant guide have been updated, including code examples and content.
- [x] All relevant configurations have been updated.

Your ENS/address: obstropolos.eth


roninjin10:
Would prefer this to point to ssx in docs but keep example the same.  What is nice about the current example is it shows you at a low level how sign in with ethereum works. This is a very useful library that I will likely use myself now that I know about it but bit of an indirection in the docs

PR #2298, opened 2023-05-09
Fix CLI React Plugin bug to allow hook generation when there are two similar function names and one contains underscore

### LINKED ISSUE #2297: bug: @wagmi/cli (React plugin) - HookNameError throws when a contract contains two similar named functions with / without underscore (e.g. totalSupply & _totalSupply)

### Package Version

1.0.0

### Current Behavior

In some cases a contract may contain two similar named functions which are external/public. For example totalSupply and _totalSupply, or name and _name.

If the Wagmi CLI config file includes the React plugin this will throw with the error ``Hook name "useFooTotalSupply" must be unique for contract "foo"``.

The lines which use the pascalCase function on the function names strips it off any non letters / numbers ([here](https://github.com/wagmi-dev/wagmi/blob/7b47ac330810d67be0215ae541951b1990d25ff0/packages/cli/src/plugins/react.ts#LL214C20-L214C20) and [here](https://github.com/wagmi-dev/wagmi/blob/7b47ac330810d67be0215ae541951b1990d25ff0/packages/cli/src/plugins/react.ts#LL214C20-L214C20))

This is the default behavior of the change-case package.

So it tries to generate two hooks both named "useFooTotalSupply".

### Expected Behavior

It should be able to generate hooks for all external/public functions.

In the case where there are two functions with identical names (but one beginning with an underscore) it should generate hooks for both.

### Steps To Reproduce

In a wagmi.config.ts file:

```
import { defineConfig } from "@wagmi/cli";
import { actions, react } from "@wagmi/cli/plugins";
import { comptrollerAbi } from "./src/abis/comptroller_abi.js";
import { ctokenAbi } from "./src/abis/ctoken_abi.js";

export default defineConfig({
  out: "src/generated.ts",
  contracts: [
    { name: "foo", 
       abi: [
                {
                  type: 'function',
                  name: 'totalSupply',
                  stateMutability: 'view',
                  inputs: [],
                  outputs: [{ type: 'uint256' }],
                },
                {
                  type: 'function',
                  name: '_totalSupply',
                  stateMutability: 'view',
                  inputs: [],
                  outputs: [{ type: 'uint256' }],
                },
              ] },
  ],
  plugins: [react()],
});
```

``yarn wagmi generate``

Expected error:
``Hook name "useFooTotalSupply" must be unique for contract "foo"``

_No response_

### Anything else?

_No response_

zilayo:
## Description

fixes #2297

cli react plugin currently fails to generate hooks when there are functions with the same name, but one begins with an underscore - eg totalSupply and _totalSupply.

The default behavior of the change-case package is to strip all non letters/numbers.

This PR extends the [default stripRegexp options](https://github.com/blakeembrey/change-case/blob/master/README.md#options) to ignore underscores.

Also included a test case to ensure it still throws if identical function names are found after pascal casing (this will only occur if function names are the same after stripping a symbol as [the loop already continues if there are two identical names](https://github.com/wagmi-dev/wagmi/blob/7b47ac330810d67be0215ae541951b1990d25ff0/packages/cli/src/plugins/react.ts#L211)). One function is named ``totalSupply`` and the other ``-totalSupply`` - although a real ABI will never contain a hyphen, this was required to demonstrate the test case.

## Additional Information

- [ x ] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

### DIFF
--- a/.changeset/few-icons-pump.md
+++ b/.changeset/few-icons-pump.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/cli': minor
+---
+
+Fixed react plugin bug which would throw error when there were identical function names - one starting with an underscore and one without
--- a/packages/cli/src/plugins/react.test.ts
+++ b/packages/cli/src/plugins/react.test.ts
@@ -98,5 +98,41 @@ describe('react', () => {
         format(`${imports}\n\n${content}`),
       ).resolves.toMatchSnapshot()
     })
+
+    it('throws when function names in a contract are non unique', async () => {
+      await expect(
+        react().run({
+          contracts: [
+            {
+              name: 'foo',
+              abi: [
+                {
+                  type: 'function',
+                  name: 'totalSupply',
+                  stateMutability: 'view',
+                  inputs: [],
+                  outputs: [{ type: 'uint256' }],
+                },
+                {
+                  type: 'function',
+                  name: '-totalSupply',
+                  stateMutability: 'view',
+                  inputs: [],
+                  outputs: [{ type: 'uint256' }],
+                },
+              ],
+              content: '',
+              meta: {
+                abiName: 'fooAbi',
+              },
+            },
+          ],
+          isTypeScript: true,
+          outputs: [],
+        }),
+      ).rejects.toThrowErrorMatchingInlineSnapshot(
+        `"Hook name \\"useFooTotalSupply\\" must be unique for contract \\"foo\\"."`,
+      )
+    })
   })
 })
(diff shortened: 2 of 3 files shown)


SahilPanhotra:
Any update to this? is this going to merge ? I am facing the same error

tmm:
Closing as this is now captured by the [`getHookName`](https://wagmi.sh/cli/api/plugins/react#gethookname) option in Wagmi CLI v2.

timothywangdev-cb:
I don't think this is captured by the getHookName, the `itemName` passed into the function somehow got it's underscore removed, so using getHookName doesn't really hep in this case

PR #2396, opened 2023-05-22
feat: add gas fee when getCallParameters return gas value undefined

siosio34:
## Description

I connecting to Metamask using walletconnect-legacy-provider, then using the writeContract function, the function calling the transaction failed. As a result of figuring out the cause, the prepareWriteConfig function was returning gasPrice as undefined. I decided that this was an issue of WalletConect, and got a hint from pr(https://github.com/wagmi-dev/wagmi/pull/2344) below and added logic to find a new gas fee using publicClient when the gas fee is undefined.

If it is not fixed in this way, I would appreciate it if you could tell me how to solve the problem that gasPrice returns undefined in prepareWriteContract when using walletconnect legacy.

Then have a nice day. Fighting wagmi team~

## Additional Information

- [X] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address: 0x05C844b78d17E3b4be53eEBb72ec5Bcd32f85ad6

### DIFF
--- a/packages/core/src/actions/contracts/prepareWriteContract.ts
+++ b/packages/core/src/actions/contracts/prepareWriteContract.ts
@@ -31,6 +31,7 @@ export type PrepareWriteContractResult<
 > = Omit<SimulateContractReturnType<TAbi, TFunctionName>, 'request'> & {
   request: SimulateContractReturnType<TAbi, TFunctionName>['request'] & {
     chainId?: TChainId
+    gas?: bigint | null
   }
   mode: 'prepared'
 }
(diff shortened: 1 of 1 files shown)


jxom:
Thanks for the PR! This was left out intentionally as Wallets will estimate the gas needed for a transaction to execute themselves. I tried to reproduce your issue with MetaMask + WalletConnect v1, however, I was unable to reproduce it and was successfully able to write to a contract. Do you mind providing a minimal reproducible example via https://new.wagmi.sh so I can look into it further?

If you really need to retrieve a gas estimate, you can consider using [viem's `estimateContractGas`](https://viem.sh/docs/contract/estimateContractGas.html). We will consider adding this as a Hook into wagmi soon!

PR #2401, opened 2023-05-23
feat: expose `account` on `readContract`/`useContractRead`

### LINKED ISSUE #2397: bug: readContract function gives empty output

### Package Version

1.0.5

### Current Behavior

giving empty values 
like `["",""]`

1.
function on smart contract that should be called
```
function getUser() public returns (string memory, string memory){
        a = users[msg.sender].userId;
        b = users[msg.sender].name;
        return (a, b);
    }
```
function calling it
```
const getData = async () => {
    const data = await readContract({
      address: CONTRACT_ADDRESS,
      abi: CONTRACT_ABI,
      functionName: 'getUser',
    });
    console.log(data);
    setData(data);
  };

```
should open metamask for signing the transaction but doesn't and returns empty values like above
doesn't produce any errors, just prints empty values
when this function is called

2.
Also tried changing function on smart contract
doesn't produce any errors, just prints empty values
```
function getUser() public view returns (string memory, string memory){
        User storage u = users[msg.sender];
        return (u.userId, u.name);
    }
```

### Expected Behavior

It should open metamask for signing.
And should not print empty values..
It should actually call the function on smart contract. (I think its not doing that) 

### Steps To Reproduce

Run the attached react code..
No need for any installation..
1. Connect the wallet
2. write data by filling both input fields with string data
3. now call get data function, but it doesn't work (issue)

https://stackblitz.com/edit/vite-react-tailwindcss-jtdlrf?file=src%2Fpolyfills.js

### Anything else?

_No response_

jxom:
Fixes #2397. Also updates viem.

### DIFF
--- a/.changeset/rare-turtles-visit.md
+++ b/.changeset/rare-turtles-visit.md
@@ -1,0 +1,6 @@
+---
+'@wagmi/core': patch
+'wagmi': patch
+---
+
+Exposed `account` on `readContract`/`useContractRead`.
--- a/docs/package.json
+++ b/docs/package.json
@@ -20,7 +20,7 @@
     "react-dom": "^18.2.0",
     "siwe": "^1.1.6",
     "wagmi": "workspace:*",
-    "viem": "~0.3.18"
+    "viem": "~0.3.35"
   },
   "devDependencies": {
     "@preconstruct/next": "^4.0.0",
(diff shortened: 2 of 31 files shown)


tmm (approved):
Looks good. Couple test errors to clean up.

Issue #1712, opened 2023-01-24
bug: type `Address` is not correctly picked even after configuring underlying `AddressType`


technophile-04:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

0.10.10

### Current Behavior

As mentioned in this comment https://github.com/wagmi-dev/wagmi/issues/1330#issuecomment-1320317099 if you try to configure the underlying `AddressType` it works but things such as type `ConnectorData` (specifically stuff from `wagmi/connectors`)  still have types as `0x{string}`. 

### Expected Behavior

Once configured the underlying `AddressType` to `string`  as mentioned in https://github.com/wagmi-dev/wagmi/issues/1330#issuecomment-1320317099  it should be reflected everywhere. 

### Steps To Reproduce

Example repo link: https://github.com/technophile-04/se-2/tree/issue/wagmi

if you clone the repo checkout to `issue/wagmi` branch and go to `BurnerConnector.ts` you will find some typescript errors because in some places AdressType still refers to 0x{string} even after adding the declaration file.

### Link to Minimal Reproducible Example (CodeSandbox, StackBlitz, etc.)

https://github.com/technophile-04/se-2/tree/issue/wagmi

### Anything else?

I think the reason for this is how wagmi is handling dependency of 'abitype'(I may be wrong here though) , checkout below SS : 
![Screenshot 2023-01-23 at 12 38 10 PM](https://user-images.githubusercontent.com/80153681/213962270-b2e8e37d-9a97-4d56-9b43-be325aac513e.jpg)
The above SS is of `node_modules` of `nexjs`directory. As we can see the `core` does not have `abitype` in its `node_modules` and thus using the `abitype` from our repo and using anything from `core` is inferring correct type for `AddressType : string` but if we look at `connectors` directory in above SS there is `abitype` in its `node_modules` because of which its taking `abitype` from its directory instead of targeting `abitype` of our repo. Since importing stuff from `connectors` in `BurnerConnector.ts` the types for `AddressTypes` there are `0x{string}`. 

can this line be causing issue in wagmi ? https://github.com/wagmi-dev/wagmi/blob/b846be6b1a3c0a6c967dc22c3fde7d3acdbf0607/tsconfig.json#L27 

For more description checkout https://github.com/scaffold-eth/se-2/pull/96#issuecomment-1399269480

tmm:
[...]

PR #2421, opened 2023-05-24
Fix `AddressType` when using wagmi and viem

technophile-04:
## Description

While migrating to wagmi V1 in https://github.com/scaffold-eth/scaffold-eth-2/pull/352 the `AddressType` is not correctly picked (We have [configured `AddressType`](https://github.com/scaffold-eth/scaffold-eth-2/blob/3cb837e04940ad12c01da1744b774590f4490ab9/packages/nextjs/types/abitype/abi.d.ts#L5) as `string` ) but it still complains type mismatch since its expecting `0x{string}`. 

Checkout this comment -> https://github.com/scaffold-eth/scaffold-eth-2/pull/352#discussion_r1204768900 for more context 

## Additional Information
**Solution** : My assumption is that this is caused due to version mismatch of `abiType` in `viem` which uses `0.8.2` and `wgmi` which uses `0.8.1` ( seems like related to #1712 ), So updated the version of packages using `abitype` in wagmi to `0.8.2` inline with `viem`. 

Created a direct PR because felt similar to #1712, but please feel free to close this PR in favour of better option, Tysm 🙌

- [X] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address: shivbhonde.eth

### DIFF
--- a/.changeset/cold-bears-battle.md
+++ b/.changeset/cold-bears-battle.md
@@ -1,0 +1,7 @@
+---
+'@wagmi/cli': major
+'@wagmi/core': major
+'wagmi': major
+---
+
+Bump the abitype version to 0.8.2 inline with viem
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -68,7 +68,7 @@
   },
   "dependencies": {
     "@wagmi/chains": "workspace:*",
-    "abitype": "0.8.1",
+    "abitype": "0.8.2",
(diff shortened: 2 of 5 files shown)


tmm:
this should do it https://github.com/wagmi-dev/wagmi/pull/2441

PR #2584, opened 2023-06-20
fix: cannot read properties from undefined (reading 'chains')

ViniFlores:
## Description

Fix reading from undefined, `publicClient` might be undefined

## Additional Information

Your ENS/address:
0x5b6D8Dd2bD7014AA3650c67D225FF7bd14CdE303


jxom:
Interesting. `publicClient` is a [required property](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/config.ts#L27-L29) on `createConfig`. How is it getting undefined?

PR #2664, opened 2023-07-01
fix: `useAccount` should not unsubscribe from zustand when re-render

### LINKED ISSUE #1998: bug: onConnect not triggered when another hook is using the `useAccount`

### Package Version

0.11.7 (also tried 0.12.1)

### Current Behavior

The function `onConnect` is not called when connecting a wallet to the application.
When refreshing the page (with the wallet connected) then the function is triggered.

### Expected Behavior

The function `onConnect` should be called when the wallet is connected not after reloading the page

### Steps To Reproduce

- Clone https://github.com/antho1404/issue-rainbowkit-wagmi
- Run `npm install`
- Run `npm run dev`
- Open http://localhost:3000
- Click the "connect" button
  - With normal behavior, you should receive an alert with "do action on login"
  - With the current behavior you get nothing

https://github.com/antho1404/issue-rainbowkit-wagmi

### Anything else?

The issue seems to come from a hook that is also using the `useAccount`. If this hook does not use the `useAccount`, everything works as expected.

llllvvuu:
fixes #1998

fix is mostly re-doing https://github.com/wagmi-dev/wagmi/commit/3cef111b1e30120233d8754b33587cdf94aedd8f in a more simple way. Have confirmed that:
* The problem fixed in #1679 is still fixed
* The problem in #1998 is fixed (tested example repo working after change)

## Additional Information

### DIFF
--- a/packages/react/src/hooks/accounts/useAccount.ts
+++ b/packages/react/src/hooks/accounts/useAccount.ts
@@ -22,26 +22,25 @@ export type UseAccountConfig = {
 export function useAccount({ onConnect, onDisconnect }: UseAccountConfig = {}) {
   const account = useSyncExternalStoreWithTracked(watchAccount, getAccount)
   const previousAccountRef = React.useRef<typeof account>()
-  const previousAccount: Partial<typeof account> =
-    previousAccountRef.current ?? {}
+  const previousAccount = previousAccountRef.current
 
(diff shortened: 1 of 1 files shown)


jxom:
Tests are failing b/c of remote origin. Tests passed locally for me.

PR #2669, opened 2023-07-02
feat(react): `useAccount` and `useNetwork` will react to changing `config` prop of `WagmiConfig`

### LINKED ISSUE #2665: bug: stale subscription in `useAccount()`; `config` should be a hook dependency
EDIT: PR is up: #2669

### Package Version

1.3.3

### Current Behavior

In [this minimal repo example](https://github.com/llllvvuu/wagmi-react-stale-closure), we can see that `rainbowkit`'s `<ConnectButton />` is not reacting to an update in the `wagmi` config, this is because of:
https://github.com/wagmi-dev/wagmi/blob/c8139b1cd5b9cc78961e158cc6f0de59f198908f/packages/react/src/hooks/accounts/useAccount.ts#L25
will be stale because it is inside a `useCallback` where the dependencies do not track the config. Therefore it will be subscribed to the wrong `config` and thus be not reactive at all. (ConnectButton will continue to show "Connect Wallet" despite wallet being connected)

This will also apply to `useNetwork`, since `watchNetwork`, and maybe other hooks.

Going further, technically this behavior stems from the vanilla API, but there it seems like not as much of an issue since it's one's understanding as a user of VanillaJS that they'll have to manage state themselves.

### Expected Behavior

Either:
* Subscriptions should update to use the new `config`. 
* Or there should be a way to add `connectors` to the `config` without creating a new `config` object (this would be less idiomatic from a React/FP POV)

I have a prod app in which being able to lazy load a config with more connectors would save over 1/3 of the first load JS; >80 KB gzipped.

### Steps To Reproduce

* Disconnect
* Refresh the page
* Click Connect
* Button will continue to say "Connect Wallet" even while connected

https://codesandbox.io/p/github/llllvvuu/wagmi-react-stale-closure/main

### Anything else?
Solution should not have merge conflict with https://github.com/wagmi-dev/wagmi/pull/2664

See also: https://github.com/wagmi-dev/wagmi/discussions/2666

One (not very good) workaround:
* Also update the `key` of `WagmiConfig` when updating the `config`; this causes a re-mount of the app which causes all hooks to update, at the cost of nuking all off the app's state

llllvvuu:
The use-case is to make DApps interactive using providers + one connector, while lazy-importing many connectors; this PR resolves #2665 which was a blocker.

Although this PR adds reactivity w.r.t. the `useConfig()` return value, it does not actually use said return value, since:

* the `@wagmi/core` API does not accept `config`
* I didn't want to re-implement the `@wagmi/core` methods

Further discussion here: #2666

### DIFF
--- a/packages/react/src/hooks/accounts/useAccount.ts
+++ b/packages/react/src/hooks/accounts/useAccount.ts
@@ -3,7 +3,7 @@ import type {
   PublicClient,
   WatchAccountCallback,
 } from '@wagmi/core'
-import { getAccount, watchAccount as watchAccountCore } from '@wagmi/core'
+import { getAccount, watchAccount } from '@wagmi/core'
 import * as React from 'react'
 import { useConfig } from 'wagmi/context'
 
--- a/packages/react/src/hooks/accounts/useAccount.ts
+++ b/packages/react/src/hooks/accounts/useAccount.ts
@@ -1,6 +1,11 @@
-import type { GetAccountResult } from '@wagmi/core'
-import { getAccount, watchAccount } from '@wagmi/core'
+import type {
+  GetAccountResult,
+  PublicClient,
+  WatchAccountCallback,
+} from '@wagmi/core'
+import { getAccount, watchAccount as watchAccountCore } from '@wagmi/core'
 import * as React from 'react'
+import { useConfig } from 'wagmi/context'
 
 import { useSyncExternalStoreWithTracked } from '../utils'
 
@@ -20,6 +25,15 @@ export type UseAccountConfig = {
 }
 
 export function useAccount({ onConnect, onDisconnect }: UseAccountConfig = {}) {
+  const config = useConfig()
+  const watchAccount = React.useCallback(
+    (callback: WatchAccountCallback<PublicClient>) =>
+      // ideally this would be `watchAccountCore(callback, undefined, config)`,
+      // but `watchAccountCore` does not take `config`;
+      // for now, this works due to referential inequality
+      watchAccountCore(callback),
+    [config],
+  )
   const account = useSyncExternalStoreWithTracked(watchAccount, getAccount)
   const previousAccountRef = React.useRef<typeof account>()
   const previousAccount = previousAccountRef.current
--- a/packages/react/src/hooks/accounts/useAccount.ts
+++ b/packages/react/src/hooks/accounts/useAccount.ts
@@ -34,7 +34,7 @@ export function useAccount({ onConnect, onDisconnect }: UseAccountConfig = {}) {
       watchAccount(callback),
     [config],
   )
-  const account = useSyncExternalStoreWithTracked(watchAccount, getAccount)
+  const account = useSyncExternalStoreWithTracked(watchAccount_, getAccount)
   const previousAccountRef = React.useRef<typeof account>()
   const previousAccount = previousAccountRef.current
 
--- a/packages/react/src/hooks/accounts/useAccount.ts
+++ b/packages/react/src/hooks/accounts/useAccount.ts
@@ -28,9 +28,9 @@ export function useAccount({ onConnect, onDisconnect }: UseAccountConfig = {}) {
   const config = useConfig()
   const watchAccount = React.useCallback(
     (callback: WatchAccountCallback<PublicClient>) =>
-      // ideally this would be `watchAccountCore(callback, undefined, config)`,
-      // but `watchAccountCore` does not take `config`;
-      // for now, this works due to referential inequality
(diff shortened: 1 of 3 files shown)


packages/react/src/hooks/accounts/useAccount.ts
```diff
@@ -1,6 +1,11 @@
-import type { GetAccountResult } from '@wagmi/core'
-import { getAccount, watchAccount } from '@wagmi/core'
+import type {
+  GetAccountResult,
+  PublicClient,
+  WatchAccountCallback,
+} from '@wagmi/core'
+import { getAccount, watchAccount as watchAccountCore } from '@wagmi/core'
```
jxom, line 6:
```suggestion
import { getAccount, watchAccount } from '@wagmi/core'
```

packages/react/src/hooks/accounts/useAccount.ts
```diff
@@ -20,6 +25,15 @@ export type UseAccountConfig = {
 }
 
 export function useAccount({ onConnect, onDisconnect }: UseAccountConfig = {}) {
+  const config = useConfig()
+  const watchAccount = React.useCallback(
```
jxom, line 29:
```suggestion
  const watchAccount_ = React.useCallback(
```

packages/react/src/hooks/accounts/useAccount.ts
```diff
@@ -20,6 +25,15 @@ export type UseAccountConfig = {
...
+    (callback: WatchAccountCallback<PublicClient>) =>
+      // Ideally this would be `watchAccountCore(callback, undefined, config)`,
+      // but `watchAccountCore` does not take `config` (#2666).
+      // For now, this works due to referential inequality.
+      watchAccountCore(callback),
```
jxom, line 34:
```suggestion
      watchAccount(callback),
```

packages/react/src/hooks/accounts/useAccount.ts
```diff
@@ -20,6 +25,15 @@ export type UseAccountConfig = {
...
+    [config],
+  )
   const account = useSyncExternalStoreWithTracked(watchAccount, getAccount)
```
jxom, line 37:
```suggestion
  const account = useSyncExternalStoreWithTracked(watchAccount_, getAccount)
```

packages/react/src/hooks/accounts/useNetwork.ts
```diff
@@ -1,7 +1,20 @@
-import { getNetwork, watchNetwork } from '@wagmi/core'
+import type { WatchNetworkCallback } from '@wagmi/core'
+import { getNetwork, watchNetwork as watchNetworkCore } from '@wagmi/core'
```
jxom, line 2:
```suggestion
import { getNetwork, watchNetwork } from '@wagmi/core'
```

packages/react/src/hooks/accounts/useNetwork.ts
```diff
@@ -1,7 +1,20 @@
...
+import { useCallback } from 'react'
+import { useConfig } from 'wagmi/context'
 
 import { useSyncExternalStoreWithTracked } from '../utils'
 
 export function useNetwork() {
+  const config = useConfig()
+  // rome-ignore lint/nursery/useExhaustiveDependencies: see comment below
+  const watchNetwork = useCallback(
```
jxom, line 11:
```suggestion
  const watchNetwork_ = useCallback(
```

packages/react/src/hooks/accounts/useNetwork.ts
```diff
@@ -1,7 +1,20 @@
...
+    (callback: WatchNetworkCallback) =>
+      // Ideally this would be `watchNetworkCore(callback, undefined, config)`,
+      // but `watchNetworkCore` does not take `config` (#2666).
+      // For now, this works due to referential inequality.
+      watchNetworkCore(callback),
```
jxom, line 16:
```suggestion
      watchNetwork(callback),
```

packages/react/src/hooks/accounts/useNetwork.ts
```diff
@@ -1,7 +1,20 @@
...
+    [config],
+  )
   return useSyncExternalStoreWithTracked(watchNetwork, getNetwork)
```
jxom, line 19:
```suggestion
  return useSyncExternalStoreWithTracked(watchNetwork_, getNetwork)
```

jxom (commented):
Looks good to me! Just need a patch changeset: `pnpm changeset`.

llllvvuu:
> Looks good to me! Just need a patch changeset: `pnpm changeset`.

Done, thanks!

PR #2719, opened 2023-07-10
Fixed: SignMessageArgs types from Viem

YuriGii:
## Description

signMessage message arg type was set to string only
```
export type SignMessageArgs = {
  /** Message to sign with wallet */
  message: string
}
```

used SignMessageParameters["message"] directly from viem that includes Hex and ByteArray options as per documentation https://viem.sh/docs/actions/wallet/signMessage.html 
```
export type SignMessageArgs = {
  /** Message to sign with wallet */
  message: SignMessageParameters["message"];
}
```


packages/core/src/actions/accounts/signMessage.ts
```diff
@@ -1,11 +1,11 @@
-import type { SignMessageReturnType } from 'viem'
+import type { SignMessageReturnType, Hex, ByteArray } from 'viem'
 
 import { ConnectorNotFoundError } from '../../errors'
 import { getWalletClient } from '../viem'
 
 export type SignMessageArgs = {
   /** Message to sign with wallet */
-  message: string
+  message: string | { raw: Hex | ByteArray }
```
jxom, line 8: can we use `SignMessageParameters` from viem?

```suggestion
  message: SignMessageParameters['message']
```

tmm: Should check that `SignMessageParameters` exists in this peer dep range `"viem": ">=0.3.35"`

tmm (commented):
Please add a changeset with `pnpm changeset`

PR #2737, opened 2023-07-14
update pulsechainV4 symbol to tPLS

sajadevo:
## Description

Update the symbol for **pulsechainV4** from `v4PLS` to `tPLS` since the correct symbol is `tPLS` not `v4PLS`. 

I just noticed that when I was using wagmi with rainbowkit in a NFT marketplace project.

## Additional Information


tmm:
The core team makes these updates periodically.

PR #2808, opened 2023-07-26
feat: Modify Alchemy provider's api key to be mapped by chain.

siosio34:
## Description

hello. I am using alchemyProvider, but inside alchemy, it is recommended to use different API Keys for each chain, so I changed the internal source.

The documentation code has not been included yet, but if the reflection of the code is judged to be okay, we will also modify the docs.

If you want to support the type of the previous version of the syntax, you can do additional work, but after contacting Alchemy, it seems that it is not encouraged. Same.

Finally, thank you for making a good open source. Thanks to you, my working hours are shorter.

## Additional Information

- [O] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

### DIFF
--- a/docs/components/core/Providers.tsx
+++ b/docs/components/core/Providers.tsx
@@ -12,7 +12,13 @@ import { alchemyProvider } from 'wagmi/providers/alchemy'
 
 const { chains, publicClient, webSocketPublicClient } = configureChains(
   [mainnet, goerli],
-  [alchemyProvider({ apiKey: process.env.NEXT_PUBLIC_ALCHEMY_ID! })],
+  [
+    alchemyProvider({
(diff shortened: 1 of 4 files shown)


tmm:
Appreciate the PR! The concept of providers per service like this will be removed in v2. It's more trouble than it's worth to hardcode RPC base URLs internally and much better folks specify full RPC URLs. In the meantime, you can use the [jsonRpc](https://wagmi.sh/react/providers/jsonRpc) provider. existing pattern comes from when wagmi used ethers internally.

siosio34:
@tmm  Thank you for your review. ~

PR #2862, opened 2023-08-07
Fix an issue with useContractInfiniteReads with multicall.

dankelleher:
## Description

The number of results per page was not taking into account the number of calls. As a result, if the number of calls made exceeds the page size, it failed the getNextPageParam check, which made wagmi think there were no more pages.

This fix calculates the number of expected calls per page, and uses that in getNextPageParam.

There was an existing test for this behaviour, but the inline snapshot for it was incorrect - this fixes the test, thereby covering the behaviour.

## Additional Information

- [ x] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

dankelleher.eth

### DIFF
--- a/.changeset/thick-gifts-lay.md
+++ b/.changeset/thick-gifts-lay.md
@@ -1,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+Fix an issue with useContractInfiniteReads with multicall.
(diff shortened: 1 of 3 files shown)


dankelleher:
Fix for #2863

tmm:
See https://github.com/wevm/wagmi/issues/2863#issuecomment-1878964113

PR #2941, opened 2023-08-25
feat(cli): add eslint-disable to generated output

llllvvuu:
## Description

Really quick fix to save a few seconds typing an ignore entry

## Additional Information

Your ENS/address:


tmm:
It's probably best to add this file to your eslintignore rather than add this here since this isn't as explicit/clear.

PR #2946, opened 2023-08-27
feat(cli): fallback to default public provider's chain id

tomquirk:
## Description

This PR modifies how the chain ID is sourced in `useContractRead`s generated from `wagmi/cli/react`.

**problem**: if there's no connected wallet, `useContractRead` hooks generated from `wagmi/cli/react` won't run. (more context inline: https://github.com/wagmi-dev/wagmi/pull/2946#discussion_r1306682870)

**This PR**

Now, hooks will fall back to public provider's chain ID (using wagmi's `useChainId` hook) if both the following are true:
- `useNetwork()` returns an empty object (=> no connected wallet).
- `config.chainId` is false (=> the dev didn't pass `chainId` in their call).

**The result:** hooks generated from `wagmi/cli/react` now run when the user's wallet isn't connected.

## Additional Information

Your ENS/address: aeolian.eth

### DIFF
--- a/packages/cli/src/plugins/react.ts
+++ b/packages/cli/src/plugins/react.ts
@@ -104,9 +104,11 @@ export function react(config: ReactConfig = {}): ReactResult {
             if (Object.keys(contract.address).length > 1) {
               innerHookParams.address = `${contract.meta.addressName}[chainId as keyof typeof ${contract.meta.addressName}]`
               imports.add('useNetwork')
+              imports.add('useChainId')
               innerContent = dedent`
                 const { chain } = useNetwork()
-                const chainId = config.chainId ?? chain?.id
+                const defaultChainId = useChainId()
+                const chainId = config.chainId ?? chain?.id ?? defaultChainId
               `
             } else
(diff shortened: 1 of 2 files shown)


packages/cli/src/plugins/react.ts
```diff
@@ -104,9 +104,11 @@ export function react(config: ReactConfig = {}): ReactResult {
             if (Object.keys(contract.address).length > 1) {
               innerHookParams.address = `${contract.meta.addressName}[chainId as keyof typeof ${contract.meta.addressName}]`
               imports.add('useNetwork')
+              imports.add('useChainId')
               innerContent = dedent`
                 const { chain } = useNetwork()
-                const chainId = config.chainId ?? chain?.id
```
tomquirk, line 109: if chainId is falsy, then the ABI lookup that follows this line will return undefined, and the contract read won't execute.

This is inconsistent with wagmi's `useContractRead` default behavior, where the public client's chain ID is used (using `useChainId` hook) when none given.

tmm (changes requested):
Please add a changeset with `pnpm changeset`.

PR #2977, opened 2023-09-06
Add: Allow args in watchContractEvent

illestrater:
## Description

viem provides a way to filter events via the 'args' parameter, but the Wagmi wrapper is missing this. A crucial feature of Viem's args is utilizing the default 'from' filter, in which a developer has no ability to filter this using the Wagmi wrapper.

## Additional Information

Your ENS/address: tim.eth


tmm:
This is already implemented in Wagmi v2 (#3068). Not going to make this change for v1 at the moment.

PR #3015, opened 2023-09-17
feat: like prepareWriteContract, allow walletClient for writeContract

0xFlicker:
## Description

Adds `walletClient` as a parameter to `writeContract`

This parameter is supported by `prepareWriteContract` 

I needed to make this change in order to get some custom scripts I wrote to work when using NodeJS + eth-provider + wagmi & friends

I made no updates to docs because the docs don't cover all of the input parameters to this function

This change lets me use viem + wagmi/cli generated code + eth-provider (to connect to desktop wallet frame.sh) together with code like this:
```
  const frame = ethProvider("frame");
  export const frameConnector = new InjectedConnector({
    options: {
      name: "Frame",
      getProvider: () => frame as any,
    },
  });
  const chain = [mainnet, sepolia, goerli].find(
    (chain) => chain.id === chainId
  );
  if (!chain) throw new Error(`Chain ${chainId} not found`);
  await frameConnector.connect({ chainId });
  const walletClient = createWalletClient({
    chain,
    transport: custom(frame),
  });
  const promiseClaimedProcessed = promiseClaimEvent({
    contractAddress: "0x8297AA011A99244A571190455CE61846806BF0ce",
    chainId,
  });
  // add a test allowance
  const allowanceConfig = await prepareWriteTestAllowance({
    chainId,
    account: claimingAddress,
    address: "0x8297AA011A99244A571190455CE61846806BF0ce",
    functionName: "claim",
    walletClient,
    args: [[bitcoinAddress]],
  });
  const testAllowanceResult = await writeTestAllowance({
    ...allowanceConfig,
    walletClient,
  });
  ```

otherwise I get an exception in `writeTestAllowance` because walletClient attempts to read from the global state, unlike prepare, which accepts from arguments

## Additional Information

Your ENS/address:
0xflick.xyz

### DIFF
--- a/.changeset/empty-shirts-prove.md
+++ b/.changeset/empty-shirts-prove.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/core": patch
+---
+
+Added walletClient parameter to writeContract
--- a/packages/core/src/actions/contracts/writeContract.test.ts
+++ b/packages/core/src/actions/contracts/writeContract.test.ts
@@ -34,12 +34,39 @@ describe('writeContract', () => {
     expect(hash).toBeDefined()
   })
 
+  it('prepared with walletClient', async () => {
+    const { request } = await prepareWriteContract({
+      ...wagmiContractConfig,
+      functionName: 'mint',
+      args: [getRandomTokenId()],
+      walletClient: getWalletClients()[0]!,
+    })
+    const { hash } = await writeContract({
+      ...request,
+      walletClient: getWalletClients()[0]!,
+    })
+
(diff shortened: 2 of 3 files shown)


0xFlicker:
updated description with code why I made this change

only used currently in the context of an automated test, so not blocking me in any way

but also this change seems reasonable to me

tmm:
Not going to change how this works at the moment. Wagmi v2 (see #3068) has a better abstraction for this so don't want to add this now and take it away once that is released.

PR #3034, opened 2023-09-21
fix: storage in private browsing mode

Solstice1557:
## Description

Browsing in incognito mode `SecurityError` exception can be thrown when accessing window.localStorage. This is causing `createClient` function to fail.
Here is the links about this error:

- [https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document/)](https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document/)
- [https://stackoverflow.com/questions/30481516/iframe-in-chrome-error-failed-to-read-localstorage-from-window-access-deni](https://stackoverflow.com/questions/30481516/iframe-in-chrome-error-failed-to-read-localstorage-from-window-access-deni)

This patch will fix this issue. 

## Additional Information

- [ ✅] I read the [contributing docs](/wagmi-dev/wagmi/blob/main/.github/CONTRIBUTING.md) (if this is your first contribution)

Your ENS/address:

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -5,7 +5,7 @@ import { createStore } from 'zustand/vanilla'
 import type { Connector, ConnectorData } from './connectors'
 import { InjectedConnector } from './connectors'
 import type { ClientStorage } from './storage'
-import { createStorage, noopStorage } from './storage'
+import { createStorage, selectStorage } from './storage'
 import type { PublicClient, WebSocketPublicClient } from './types'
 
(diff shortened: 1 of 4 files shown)


tmm:
Appreciate your effort here. We are now onto Wagmi v2 so closing this.

PR #3069, opened 2023-09-29
wip: alpha

tmm:
Wagmi v2 Alpha. See #3068 for more info.

https://github.com/wagmi-dev/viem/pull/1149


docs/react/guides/ssr.md
```diff
@@ -0,0 +1,162 @@
...
+})
+```
+
+Turning on the `ssr` property means that content from the external stores will be hydrated on the client after the initial mount.
+
+## Persistence using Cookies
+
+As a result of turning on the `ssr` property, external persistent stores like `localStorage` will be hydrated on the client **after the initial mount**.
+
+This means that you will still see a flash of "empty" data on the client (ie. a `"disconnected"` account instead of a `"reconnecting"` account, or an empty address instead of the last connected address) until after the first mount, when the store hydrates.
```
glitch-txs, line 33: Cool, what about adding a "loading" state? e.g. to prevent clicking on the connect button before data gets hydrated (although that should usually be pretty quick)

jxom: Wouldn't this be handled by checking `status` in `useAccount`?

tmm:
Closed in favor of https://github.com/wagmi-dev/wagmi/pull/3180

PR #3086, opened 2023-10-04
WalletConnect connector optimizations, only autoConnect the latest used connector

wille:
Fixes what is described here https://github.com/wagmi-dev/wagmi/discussions/3085


glitch-txs:
Heads up, we'll go full optional namespaces and remove required ones completely

tmm:
Not changing this behavior as Wagmi v2 supports multiple active connections so we want to reconnect to all of them. https://github.com/wagmi-dev/wagmi/discussions/3085#discussioncomment-7192138

PR #3165, opened 2023-10-27
fix: Proposal to repair refresh/reconnect problem

richtera:
## Description

When refreshing the page or tag the code runs autoConnect. Currently the providers are not being passed the desired chainId
and address making it not possible for the provider to check whether they are already connected and not prompting the user.
Adding these two fields during the "reconnecting" state and then implementing the code in the injected provider to check
if it's connected to the correct thing.

## Additional Information

Your ENS/address:

A lot of these changes are stale since they are in files which have been deleted.

### DIFF
--- a/.changeset/strange-goats-act.md
+++ b/.changeset/strange-goats-act.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/core": patch
+---
+
+Repair re-connect for generic injected wallet.
--- a/packages/core/src/connectors/injected.ts
+++ b/packages/core/src/connectors/injected.ts
@@ -147,7 +147,7 @@ export function injected(parameters: InjectedParameters = {}) {
       if (!provider) throw new ProviderNotFoundError()
 
       let accounts: readonly Address[] | null = null
-      if (!isReconnecting) {
+      if (isReconnecting) {
         accounts = await this.getAccounts().catch(() => null)
         const isAuthorized = !!accounts?.length
         if (isAuthorized)


packages/core/src/connectors/injected.ts
```diff
@@ -146,9 +146,9 @@ export function injected(parameters: InjectedParameters = {}) {
       const provider = await this.getProvider()
       if (!provider) throw new ProviderNotFoundError()
 
-      let accounts: readonly Address[] | null = null
-      if (!isReconnecting) {
-        accounts = await this.getAccounts().catch(() => null)
+      let accounts: readonly Address[] = []
+      if (isReconnecting) {
```
jxom, line 150: Is this meant to be changed?

richtera: This is where the actual change is. `!isRexonnecting` to `isRexonnecting`. The rest of the typescript specific type related stuff was to get lint a d build to work.

.changeset/strange-goats-act.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/core": major
```
tmm, line 2:
```suggestion
"@wagmi/core": patch
```

richtera:
I removed all of the typescript changes, my local biome config was a little bit different than yours so that it required more changes.

packages/core/src/connectors/injected.ts
richtera: I refactored so only the necessary changes are in the PR. This line is really the only line which needs changing.

frozeman:
Maybe to provide context. 
Currently it calls `eth_requestAccounts` again on page reload [here]( https://github.com/wevm/wagmi/blob/7b122755bb63cd28b3f94ad7081a0ee6e682eb8f/packages/core/src/connectors/injected.ts#L176), because of this our browser extension pops up again asking the user on page reload, which shouldn't happen, if there is already a connected account.

richtera:
Basically during a reconnect we don't want this to happen
https://github.com/wevm/wagmi/blob/7b122755bb63cd28b3f94ad7081a0ee6e682eb8f/packages/core/src/connectors/injected.ts#L176

We do want this to happen as the comment already is indicating
https://github.com/wevm/wagmi/blob/7b122755bb63cd28b3f94ad7081a0ee6e682eb8f/packages/core/src/connectors/injected.ts#L154 

Therefore I believe the condition on this line was in error since it was running when isReconnecting was false.
https://github.com/wevm/wagmi/blob/7b122755bb63cd28b3f94ad7081a0ee6e682eb8f/packages/core/src/connectors/injected.ts#L150

richtera:
@jxom @tmm I just tried to inject the provider with normal EIP 6963 messaging and the same problem persists. I really think the issue is with this "if" statement. Have you been able to look at it any further. I would rather not change our wallet to not respond to eth_requestAccounts without showing a dialog to the user. This would make it a lot harder to be able to change accounts and there is already eth_accounts which does the job and in fact is already being called inside of the "if" clause in question. I am not quite sure how to proceed. It seems clear that the condition should be reversed unless there is something else going on I am not aware of.

```js
async connect({ chainId, isReconnecting } = {}) {
      .
      .
      .

      let accounts: readonly Address[] | null = null

/**
 * This section of the code really only makes sense for isReconnecting and not for !isReconnecting
 * because if the user is not reconnecting then this.getAccounts() will be empty for sure.
 */
      if (!isReconnecting) {
        accounts = await this.getAccounts().catch(() => null)
        // Attempt to show another prompt for selecting account if already connected and `shimDisconnect` flag is enabled
        const isAuthorized = shimDisconnect && !!accounts?.length
        if (isAuthorized)
          try {
            const permissions = await provider.request({
              method: 'wallet_requestPermissions',
              params: [{ eth_accounts: {} }],
            })
            accounts = (permissions[0]?.caveats?.[0]?.value as string[])?.map(
              (x) => getAddress(x),
            )
          } catch (err) {
            const error = err as RpcError
            // Not all injected providers support `wallet_requestPermissions` (e.g. MetaMask iOS).
            // Only bubble up error if user rejects request
            if (error.code === UserRejectedRequestError.code)
              throw new UserRejectedRequestError(error)
            // Or prompt is already open
            if (error.code === ResourceUnavailableRpcError.code) throw error
          }
      }
      .
      .
      .

    },
```

jxom:
Sorry about the late response here. I have opened a new PR up at https://github.com/wevm/wagmi/pull/3779 to address this issue. Let me know what you think!

PR #3187, opened 2023-11-03
docs: update stalltimeout code usage

1997roylee:


jxom:
We could probably just remove these instead since it's on the `configureChains` API!

1997roylee:
> We could probably just remove these instead since it's on the `configureChains` API!

I hv removed stallTimeout options excluding configureChains API one.

PR #3300, opened 2023-12-12
Feat: Add core actions and hooks for `verifyMessage` and `verifyTypedData`

marthendalnunes:
## Description

Add core actions and hooks for the following Viem public actions:
- [verifyMessage](https://beta.viem.sh/docs/actions/public/verifyMessage.html)
- [verifyTypedData](https://beta.viem.sh/docs/actions/public/verifyTypedData.html)

Your ENS/address:
vitormarthendal.eth


tura11541:
Now see you could decipher who is who but the positions are twisted . This
is the original email using the first original contract code ...
The other that's doing all the negative things to the account is not the
owner .... delete the imposter  from access of this account and the other 3
associated with it as well..

tura

Issue #3340, opened 2023-12-22
bug: WalletConnect switchChain method uses the wrong block explorer URL


chybisov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

2.0.0-rc.0

### Current Behavior

While testing the WalletConnect connector, I found an issue where if you need to switch chain it fails with this error

> Uncaught (in promise) SwitchChainError: An error occurred when attempting to switch chain.
> 
> Details: {"code":-32602,"message":"Expected null or array with at least one valid string HTTPS URL 'blockExplorerUrl'. Received: [object Object]"}

<img width="906" alt="image" src="https://github.com/wevm/wagmi/assets/18644653/a597e482-2bc9-4ab9-9e50-a388996f0fdf">

### Expected Behavior

The switch chain works as expected.

### Steps To Reproduce

1. Connect wallet via WalletConnect
2. Select from and to chain and token (should be different from connected chain in the wallet)
3. Start swapping/bridging
4. Error will be thrown to the console and connected wallet with WalletConnect do nothing

### Link to Minimal Reproducible Example (StackBlitz, CodeSandbox, GitHub repo etc.)

https://playground.li.fi/

### Anything else?

I found that here https://github.com/wevm/wagmi/blob/rc/packages/connectors/src/walletConnect.ts#L266 we are trying to provide default block explorer URL, but actually providing object `{name: 'BlockExplorer', url: 'Some URL'}`

chybisov:
I've provided a PR for this https://github.com/wevm/wagmi/pull/3341

PR #3341, opened 2023-12-22
fix: provide correct block explorer URL to WalletConnect

chybisov:
## Description

Fixes the WalletConnect connector issue #3340 when calling `switchChain` for a chain that is not approved during initialization. We need to provide an actual block explorer URL instead of a block explorer object to an array of URLs.

## Additional Information

### DIFF
--- a/.changeset/pink-spoons-relate.md
+++ b/.changeset/pink-spoons-relate.md
@@ -1,0 +1,7 @@
+---
+"@wagmi/connectors": patch
+"wagmi": patch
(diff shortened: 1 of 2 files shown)


chybisov:
We are doing the same in the Coinbase connector as well. https://github.com/wevm/wagmi/blob/rc/packages/connectors/src/coinbaseWallet.ts#L177

Should I fix it there too? 👀

chybisov:
@tmm fix for Coinbase #3342 🙂

PR #3402, opened 2024-01-08
fix: parseCookie with special chars

### LINKED ISSUE #3401: SSR: split cookie by "=" will cause unexpected issue in some case
### Describe the bug

I am using wagmi with rainbowkit. The cookie of "wagmi.store" contains the `=` string in some fields:

SCR-20240108-mivp

This will cause unterminated json string error when parsing cookie. The cause comes from the following code:

https://github.com/wevm/wagmi/blob/b886c8dcf700344ca8dd43fb71a9e677ae490ce8/packages/core/src/utils/cookie.ts#L29-L34

_No response_

### Steps To Reproduce

_No response_

### Wagmi Version

2.0.3

### Viem Version

^2

### TypeScript Version

^5

### Check existing issues

### Anything else?

My suggestion: [split only the first `=`](https://stackoverflow.com/a/50452780/7422256)

I'm not sure if there is still some other edge case though.

I'm happy to make a PR if you think this is ok.

Songkeys:
## Description

fix #3401 

## Additional Information


.changeset/tender-keys-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/core": patch
+---
+
+fix parseCookie in ssr when the cookie value field has special chars "="
```
tmm, line 5:
```suggestion
Fixed SSR cookie support for cookies that have special characters, e.g. `=`.
```

.changeset/tender-keys-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/core": patch
```
tmm, line 2: Adding Wagmi so it pulls in this fix.

```suggestion
"@wagmi/core": patch
"wagmi": patch
```

packages/core/src/utils/cookie.test.ts
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from 'vitest'
+
+import { parseCookie } from './cookie.js'
+
+test('parseCookie', () => {
```
tmm, line 5: No need to update now, but [`test.each`](https://vitest.dev/api/#test-each) could be good for something like this in the future.

packages/core/src/utils/cookie.ts
```diff
@@ -27,8 +27,9 @@ export function cookieToInitialState(config: Config, cookie?: string | null) {
...
 
 export function parseCookie(cookie: string, key: string) {
-  return cookie
-    .split('; ')
-    .find((x) => x.startsWith(key))
-    ?.split('=')[1]
+  const keyValue = cookie.split('; ').find((x) => x.startsWith(`${key}=`))
+  if (!keyValue) {
+    return undefined
+  }
```
tmm, line 33:
```suggestion
  if (!keyValue) return undefined
```

.changeset/tender-keys-travel.md
Songkeys: Do I need to add `@wagmi/connectors` as well?

SCR-20240109-jlcf

packages/core/src/utils/cookie.test.ts
Songkeys: Did you meant I should remove this test file for this PR?

PR #3407, opened 2024-01-08
fix: prelude gas estimation

jxom:
## Description

Noticed that we didn't have the preluding gas estimate check in `sendTransaction`. Adding it means we reach parity with `writeContract` (w/ `simulateContract` being the prelude check), and consumers will be less surprised when unexpected errors happen upon tx execution.

## Additional Information

### DIFF
--- a/packages/core/src/actions/sendTransaction.ts
+++ b/packages/core/src/actions/sendTransaction.ts
@@ -5,8 +5,12 @@ import type {
   SendTransactionErrorType as viem_SendTransactionErrorType,
   SendTransactionParameters as viem_SendTransactionParameters,
   SendTransactionReturnType as viem_SendTransactionReturnType,
+  TransactionRequest,
 } from 'viem'
-import { sendTransaction as viem_sendTransaction } from 'viem/actions'
+import {
+  estimateGas as viem_estimateGas,
+  sendTransaction as viem_sendTransaction,
+} from 'viem/actions'
 
 import { type Config } from '../createConfig.js'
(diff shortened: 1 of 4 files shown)


packages/core/src/actions/sendTransaction.ts
```diff
@@ -58,16 +65,32 @@ export async function sendTransaction<
...
+    // Run gas estimation if no value is provided.
+    if (gas_ === undefined)
+      return viem_estimateGas(client, {
+        ...(rest as any),
+        chain: chainId ? { id: chainId } : null,
+      })
+
+    // Use provided gas value.
+    return gas_
+  })()
```
tmm, line 89: Fine with it if you don't want to change it, but this seems too clever :)

tmm (approved):

packages/core/src/actions/sendTransaction.ts
jxom: Fair, nested ternary is the go then lol

jxom: Not too happy about the maintainability aspects of alternatives as it either hinders type inference (declare w/ explicit type + if/else) or readability (nested ternary). I feel like this approach is more readable whilst having type inference on `gas`?

Might roll with this despite the clever syntax on lines 76 and 80.

Bullish on block expressions in JS lol.

tmm: Works for me!

PR #3427, opened 2024-01-10
Feat: add `prepareTransactionRequest` and `usePrepareTransactionRequest`

marthendalnunes:
## Description

- Added the `prepareTransactionRequest` action.
- Added the `usePrepareTransactionRequest` hook.

## Additional Information


packages/core/src/actions/prepareTransactionRequest.ts
```diff
@@ -0,0 +1,36 @@
...
+} from 'viem/actions'
+
+import { type Config } from '../createConfig.js'
+import { type ChainIdParameter } from '../types/properties.js'
+import { type Evaluate, type UnionLooseOmit } from '../types/utils.js'
+
+export type PrepareTransactionRequestParameters<
+  config extends Config = Config,
+> = Evaluate<
+  UnionLooseOmit<viem_PrepareTransactionRequestParameters, 'chain'> &
```
tmm, line 15: We should make `to` required (similar to how it is set up in [`sendTransaction`](https://github.com/wevm/wagmi/blob/main/packages/core/src/actions/sendTransaction.ts#L41)). We also need to wire up some generic slots for [`viem_PrepareTransactionRequestParameters`](https://github.com/wevm/viem/blob/main/src/actions/wallet/prepareTransactionRequest.ts#L58-L66). [`sendTransaction`](https://github.com/wevm/wagmi/blob/main/packages/core/src/actions/sendTransaction.ts)  should be a good guide how to set this up (you'll also want to add a `test-d.ts` file so we can test for chain properties).

packages/core/src/actions/prepareTransactionRequest.ts
```diff
@@ -0,0 +1,36 @@
...
+    ChainIdParameter<config>
+>
+
+export type PrepareTransactionRequestReturnType =
+  Evaluate<viem_PrepareTransactionRequestReturnType>
```
tmm, line 20: Few generics to wire up here as well: https://github.com/wevm/viem/blob/main/src/actions/wallet/prepareTransactionRequest.ts#L76-L84

tmm (changes requested):

packages/core/src/actions/prepareTransactionRequest.ts
marthendalnunes: thanks for the feedback, will add this shortly

PR #3429, opened 2024-01-11
add ability of client to override default actions

plusminushalf:
## Description

This allows clients to have their own implementation for certain actions. For example, [permissionless](https://www.npmjs.com/package/permissionless) has defined it's own `sendTransaction` function.

## Additional Information


packages/core/src/utils/getAction.ts
```diff
@@ -0,0 +1,23 @@
...
+  // Some minifiers drop `Function.prototype.name`, meaning that `action.name`
+  // will not work. For that case, the consumer needs to pass the name explicitly.
+  name: string,
+) {
+  return (params: params): returnType =>
+    (
+      client as Client & {
+        [key: string]: (params: params) => returnType
+      }
+    )[action.name || name]?.(params) ?? action(client, params)
```
tmm, line 22: We should probably check that `client[action.name || name]` is a function just in case.

```suggestion
  const clientAction = client[action.name ?? name]
  return (params: params): returnType => {
    if (typeof clientAction === 'function') clientAction(params)
    return action(client, params)
  }
```

tmm (commented):
Still taking a look! So far so good.

tmm:
Closing in favor of https://github.com/wevm/wagmi/pull/3496

Issue #3425, opened 2024-01-10
`ChainNotConfiguredError: Chain not configured` Thrown when Changing Config Chains


frankfka:
### Describe the bug

Our app has different deploy environments, where we configure chains depending on testnet/mainnet/etc.

With the latest `wagmi` update, we've found that `ChainNotConfiguredError: Chain not configured` is thrown when switching between configurations - I assume this is because there's some caching done at the localstorage level and re-initializing the app with a different config (that doesn't include the previous cached chain) will throw the error

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/new-wagmi-ddy6bs?file=src%2Fwagmi.ts

### Steps To Reproduce

The stackblitz is configured for `mainnet`. Run the app, then change the configuration to only use `arbitrum` - the error should occur

### Wagmi Version

2.1.1

### Viem Version

2.0.6

### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/wagmi/issues) for the bug I encountered.

0xdino:
Example of a good config:

```ts
import { createConfig, http } from 'wagmi'
import { Chain, mainnet, sepolia } from 'wagmi/chains'
import { createClient } from 'viem'

export const chains: Chain[] = [mainnet, sepolia];

const config = createConfig({
  chains: [chains[0], ...chains.slice(1)],
  client({ chain }) {
    return createClient({ chain, transport: http() })
  },
})
```

But if the user switches the chain to an unknown chain, then the last connected network is returned by useChainId, it would be better to return the default one - chains[0]

ramyjaiem:
getting the same issue when switching to another chain
we switch chain before connecting the user so my solution was adding `localStorage.removeItem('wagmi.store')` before changing the chain

firnprotocol:
@ramyjaiem interesting, i can confirm that this actually "worked" to "resolve" the issue on my end (of course it's a workaround). thanks.

firnprotocol:
btw, i noticed another thing. i _only_ got this error in localhost. on the other hand, i test _multiple_ apps in localhost, not just the one above (i.e., that for which the error was encountered). i wonder if there was some kind of pollution / cross-contamination of the local state, which was caused by two different apps being on the same address (namely, `http://localhost:8080`). so in other words this isn't "really" an issue i need to care about, since it was only triggered in unrepresentative conditions (two different dapps at same address).

Yuripetusko:
[...]

PR #3433, opened 2024-01-11
fix: getClient with unconfigured chainId

tmm:
## Description

Closes https://github.com/wevm/wagmi/issues/3425

Before this change, the following code would throw a `ChainNotConfiguredError`.

```ts
const config = createConfig({
  chains: [mainnet],
  transports: {
    [mainnet.id]: http(),
  },
})

useClient({ chainId: 10 }) // same with `usePublicClient`
```

Now `useClient` and `usePublicClient` return `undefined` if `chainId` is not a configured chain.

## Additional Information


packages/core/src/createConfig.ts
```diff
@@ -125,7 +125,10 @@ export function createConfig<
     // If the target chain is not configured, use the client of the current chain.
     type Return = Client<Transport, Extract<chains[number], { id: chainId }>>
     {
+      // chainId specified and not configured
+      if (config.chainId && !chain) throw new ChainNotConfiguredError()
       const client = clients.get(store.getState().chainId)
+      // client exists, but no chain
       if (client && !chain) return client as Return
```
tmm, line 132: This case is extremely unlikely to happen now since it would require `store.getState().chainId` to change from line `122` to `130` so might be worth removing this.

PR #3493, opened 2024-01-22
introduce ledger connector

Wozacosta:
## Description

This PR restores the Ledger connector. (reverting https://github.com/wevm/wagmi/commit/53ca1f7eb411d912e11fcce7e03bd61ed067959c) and migrating the old connector to v2 of wagmi. 

It now loads the [@ledgerhq/connect-kit](https://github.com/LedgerHQ/connect-kit?tab=readme-ov-file#-important-notice-) package without going through **@ledgerhq/connect-kit-loader**'s `loadConnectKit` function.  

**@ledgerhq/connect-kit** is thus added (with a fixed version, 1.1.9) in packages/connectors/package.json

## Additional Information

 **@ledgerhq/connect-kit-loader** is removed from the required packages.
 
 ### Additional security measures taken: 

Two new github workflows on the [@ledger/hq-connect-kit](https://github.com/LedgerHQ/connect-kit/) repository: 

- one checking any discrepancy between the version in that repo and the one published on npm, to alert ASAP (unlike during the exploit): https://github.com/LedgerHQ/connect-kit/actions/workflows/check_npm_repo.yml

- the second using [CodeQL](https://codeql.github.com/) to discover vulnerabilities early: https://github.com/LedgerHQ/connect-kit/actions/workflows/github-code-scanning/codeql

### DIFF
--- a/packages/connectors/src/ledger.ts
+++ b/packages/connectors/src/ledger.ts
@@ -1,0 +1,243 @@
+import type { EthereumProvider } from '@ledgerhq/connect-kit/dist/umd/index.d.ts'
+import { type EthereumProviderOptions } from '@walletconnect/ethereum-provider'
+
+import { createConnector, normalizeChainId } from '@wagmi/core'
+import type { Evaluate } from '@wagmi/core/internal'
+import {
+  type ProviderRpcError,
+  SwitchChainError,
+  UserRejectedRequestError,
+  getAddress,
+  numberToHex,
+} from 'viem'
+
+type LedgerConnectorWcV2Options = {
+  enableDebugLogs?: boolean
+  walletConnectVersion?: 2
+  projectId?: EthereumProviderOptions['projectId']
+  requiredChains?: number[]
+  requiredMethods?: string[]
+  optionalMethods?: string[]
+  requiredEvents?: string[]
+  optionalEvents?: string[]
+}
+
+export type LedgerParameters = Evaluate<
+  LedgerConnectorWcV2Options & {
+    enableDebugLogs?: boolean
+    /**
+     Target chain to connect to.
+     */
+    chainId?: number | undefined
+  }
+>
+
+ledger.type = 'ledger' as const
+export function ledger(parameters: LedgerParameters) {
+  type Provider = EthereumProvider
+  type Properties = {
+    createProvider: any
+    initProvider: any
+    setupListeners: any
+    removeListeners: any
+  }
+
+  let provider_: Provider | undefined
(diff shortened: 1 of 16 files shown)


packages/connectors/src/ledger.ts
```diff
@@ -0,0 +1,244 @@
+// @ts-ignore
+import { type EthereumProvider } from '@ledgerhq/connect-kit/dist/umd/index.d.ts'
```
W3stside, line 2: I wonder if you could skip the `// @ts-ignore` hack by setting inside the `package.json`:

```json
"alias": {
    "@ledgerhq/connect-kit": "@ledgerhq/connect-kit/lib-umd"
}
```

tmm (commented):
Can we lock [these versions](https://github.com/LedgerHQ/connect-kit/blob/main/packages/connect-kit/package.json#L56-L58) in `@ledgerhq/connect-kit`?

packages/connectors/src/ledger.ts
Wozacosta: Hey @W3stside No need to alias it, using `import type { ...` instead works fine and doesn't throw typescript errors. Thanks for bringing it up though!

tmm:
Looking pretty good overall. The only thing that gives me pause is [this CDN URL](https://github.com/LedgerHQ/connect-kit/blob/76a36642b123b68dcd3e976646410a5881ef0f64/packages/connect-kit/src/hooks/useAnalytics.tsx#L72) getting loaded by `@segment/analytics-next`. Seems like they have things covered in their SDK ([link](https://github.com/segmentio/analytics-next/blob/master/packages/browser)), but could be a possible vector. It would be nice if there was a flag disabling this, or at least a callout in the docs so folks are aware, e.g. someone can set up a [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) to block it if they need to.

Wozacosta:
@tmm I've considered if we really needed analytics (and thus the segment library) and we decided that it could be entirely removed. Less attack surface is a good thing, thanks for your review 👍

tmm:
Very happy where this is at! Thanks for the updates.

It takes time and effort to maintain third-party connectors. Since Wagmi is an OSS project that depends on sponsors and grants to continue our work, third-party connectors must [sponsor Wagmi](https://github.com/sponsors/wevm) via GitHub or another method in order to be included in the monorepo (communicated to the team via Slack). Once that is done, we can wrap up any remaining work, merge, and release.

PR #3539, opened 2024-01-31
chore: tweak estimate fees type tests

jxom:
Introduction of `maxFeePerBlobGas` in Viem seems to be causing these tests to fail in the [Wagmi regression tests on Viem](https://github.com/wevm/viem/actions/runs/7719107929/job/21041823398). 

Tweaking the test to match the return type, instead of strictly equal.


tmm (approved):
glad the tests caught something 🤣

PR #3555, opened 2024-02-02
chore: add check for Infura API key with error message pointing to the MetaMaskSDK documentation 

omridan159:
## Description

Add check for the Infura API key with an error message pointing to the MetaMaskSDK documentation if it missing 

## Additional Information


tmm:
Let's add this to the Wagmi docs instead. We try not to log out messages to the console. Also, not sure if the message "Wagmi isn't optimized for mobile use and relies heavily on read-only calls." is relevant since Wagmi doesn't use connector providers for reading data in most cases, preferring user-configured [`transports`](https://wagmi.sh/react/api/createConfig#transports) instead.

PR #3561, opened 2024-02-04
Solid no bun

glitch-txs:
# Solid.js Support

This PR integrates a new package for Solid first class support.

This covers the core "hooks"

- [x] useAccount -> createAccount
- [x] useAccountEffect -> createAccountEffect
- [x] useChainId -> createChainId
- [ ] useClient (type issue - signal?)
- [x] useConfig -> createConfig
- [x] useConnect -> createConnect
- [x] useConnections -> createConnections
- [ ] useConnectorClient -> createConnectorClient (error when being called)
- [x] useConnectors -> createConnectors
- [x] useDisconnect -> createDisconnect
- [x] useSwitchAccount -> createSwitchAccount
- [ ] useSwitchChain -> createSwitchChain (I forgot about this one)

## Philosophy

**Considerations:**
- [*Tanstack query*](https://tanstack.com/query/latest/docs/framework/solid/overview) follows the same naming convention from Solid, which uses the prefix "create" for functions returning reactive values
-  Tanstack functions return a Solid store, these stores cannot be destructured otherwise they lose reactivity.

1. To avoid confusion we are using "stores" (built on top of signals) instead of "signals", since the values are accessed differently (signals are functions). Since stores cannot be desctructured and for some "hooks" we are returning values from Query and from Wagmi at the same time, the return type is going to be an object containing these stores.

Example:
```ts
function Account(){
  const { account } = createAccount()

  return <div>{account.address}</div>
}
```

2. To follow the framework convention (adopted by Tanstack as well) we are using the prefix "create" instead of "use". This introduces a naming confict with `useConfing` and the core function `createConfig`, for a quick workaround the core function is being renamed to `initConfig`.  (we can discuss this)

-----
- [ ] Package (on process)
- [ ] Docs
- [ ] vite-solid playground (80%)
- [ ] ssr-solid playground
- [ ] tests

----

NOTE: This is currently under `solid-no-bun` folder but will alter be moved to the right place.


solid-no-bun/solid/src/signals/createSwitchChain.ts
```diff
@@ -0,0 +1,81 @@
...
+  const mutation = createMutation(()=>({
+    ...mutationParams,
+    ...mutationOptions,
+  }))
+
+  type Return = CreateSwitchChainReturnType<config, context>
+  return {
+    mutation,
+    chains: _config.chains,
+    switchChain: mutation.mutate as Return['switchChain'],
```
glitch-txs, line 78: for mutations using this pattern I guess this is not ideal since I assume reactivity is lost here? not sure if this is relevant here

fend25:
Hi! Is this project abandoned? Or how I can use wagmi with solid?

2wheeh:
@fend25 Hi I'm working on solidjs support #4904

PR #3563, opened 2024-02-05
internal: widen internal types

jxom:
## Description

[Narrowing formatter return types](https://github.com/wevm/viem/commit/b4cca34cc4a52e5f9697d958959b7f84bdc576b2) in Viem (to infer return type when no narrowed `chain` generic is passed) seems to [break internal types in Wagmi](https://github.com/wevm/viem/actions/runs/7791863392/job/21248850480). Previously, Viem returned `any` as the return type if it couldn't infer the chain, so that probably explains why it was not breaking before.


tmm:
Will take a look in the morning!

tmm (approved):

PR #3565, opened 2024-02-06
Improve tree-shakability by modularizing @wagmi/connectors package structure

coindegen:
## Description

This PR addresses this discussion: https://github.com/wevm/wagmi/discussions/1371

It allows for importing of all connectors like this:

```
import { walletConnect } from "@wagmi/connectors/walletConnect";
```

In addition to this:

```
import { walletConnect } from "@wagmi/connectors";
```

The current structure of the @wagmi/connectors NPM package does not favor tree-shaking. All the connectors are accessed via a single entry point (index.js) under the exports directory. This makes it impossible to import specific modules directly, and it produces a uniform, large bundle size regardless of which of which connector is imported.

Package size appears to be a known issue (https://wagmi.sh/core/api/connectors/metaMask#import). However, given my current build config (`vite build` targeting an index.html linking to a vanilla jsn file), the large Metamask SDK dependencies appear to be included in my JS no matter which connector I import.  

If you approve of this general approach I can include documentation.

## Additional Information

- [] Added or updated tests (and snapshots) related to the changes made.

### DIFF
--- a/packages/connectors/package.json
+++ b/packages/connectors/package.json
@@ -32,6 +32,22 @@
       "types": "./dist/types/exports/index.d.ts",
       "default": "./dist/esm/exports/index.js"
     },
+    "./coinbaseWallet": {
+      "types": "./dist/types/coinbaseWallet.d.ts",
+      "default": "./dist/esm/coinbaseWallet.js"
+    },
+    "./metaMask": {
+      "types": "./dist/types/metaMask.d.ts",
+      "default": "./dist/esm/metaMask.js"
+    },
+    "./safe": {
+      "types": "./dist/types/safe.d.ts",
+      "default": "./dist/esm/safe.js"
+    },
+    "./walletConnect": {
+      "types": "./dist/types/walletConnect.d.ts",
+      "default": "./dist/esm/walletConnect.js"
+    },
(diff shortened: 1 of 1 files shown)


tmm:
> given my current build config (vite build targeting an index.html linking to vanilla js), the large Metamask SDK dependencies appear to be included in my JS no matter which connector I import.

It shouldn't be. Can you [create an issue](https://github.com/wevm/wagmi/issues/new?assignees=&labels=&projects=&template=bug_report.yml) with a minimal reproduction?

coindegen:
You're right. I have isolated the issue, I think it has nothing to do with wagmi, but with my use of dynamic imports. I was attempting to follow Vite's suggestion to code-split my application:

```
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
```

Dynamic imports led to a big bundle, because Vite is unable to tree-shake this efficiently, it assumes I'll need the whole package at runtime:
```js
  const { walletConnect } = await import("@wagmi/connectors");
```

But I can confirm this is not a problem when using standard ES imports, Vite is able to tree-shake this properly:

```js
import { walletConnect } from "@wagmi/connectors";
```

Thanks for your time.

PR #3614, opened 2024-02-19
docs: remove unnecessary async keyword from getEthersSigner

Mostafatalaat770:
## Description

The async keyword makes the `getEtherSigner` return a promise when it shouldn't.

## Additional Information


docs/core/guides/ethers.md
tmm, line 1: we want `async` here since `getConnectorClient` is using `await` internally

tmm (changes requested):

docs/core/guides/ethers.md
Mostafatalaat770: This is only for the core or react as well?

tmm: Just for core

PR #3683, opened 2024-03-07
fix: filter out duplicate EIP-6963 connectors on mount (server-side)

magiziz:
## Description

- During on mount if `ssr` is true in `createConfig` the EIP-6963 connectors might be introduced multiple times. This is happening in [hydrate.ts](https://github.com/wevm/wagmi/blob/main/packages/core/src/hydrate.ts#L21-L31). By the time we add them they were already added in [createConfig.ts](https://github.com/wevm/wagmi/blob/main/packages/core/src/createConfig.ts#L274-L288)
- Created a simple function that would filter out duplicated connectors [here](https://github.com/KosmosKey/wagmi/blob/main/packages/core/src/createConfig.ts#L362-L375)

## Additional Information


jxom:
Closing in favor of #3715

PR #3693, opened 2024-03-11
Fix useBlockNumber multiple websocket subscriptions

### LINKED ISSUE #3656: Memory leak when using useBlockNumber({watch: true}) -- creates a new websocket subscription on every block until it runs out of memory
### Describe the bug

Consider a simple hook like this which consumes the block number for invalidation:

```typescript
export const useAccountData = () => {
  const queryClient = useQueryClient()
  const { address } = useAccount()
  const { data: blockNumber } = useBlockNumber({ chainId: activeChains.l2.id, watch: true })
  const {
    data: accountData,
    isLoading,
    queryKey,
  } = useReadContract({
    address: env.NEXT_PUBLIC_CONTRACT as `0x${string}`,
    abi: fooAbi,
    functionName: "fooFunction",
  })

  useEffect(() => {
    void queryClient.invalidateQueries({ queryKey })
  }, [blockNumber])

  return { accountData, isLoading }
}
```

When this is used just one time in the app and you look at the websocket traffic, on every single block event it's creating a new subscription and never closes the old ones. If you have 2 components consuming the hook, it will start 2 subscriptions with every block, etc for each consumer.

In the following screenshot notice it first gets 3 subscription responses, then starts a new one, then gets 4, etc. I ensured that hook above was only called/active one time, and setting `watch:false` resulted in no websocket activity.
Screenshot 2024-02-29 at 12 38 26 PM

The result is you will run out of memory and the RPC will rate limit the client because eventually it has too many subscriptions.

I tested this with QuickNode and the following:
```json
"@web3modal/wagmi": "^4.0.11",
```

I tried the same pattern but in a context and it's the same issue. The problem seems to be wagmi just keeps making new subscriptions.

If I instead use viem to get the block number, it works fine and only starts one subscription, even if I have several parts of the app consuming this hook for the block number:

```typescript
export const useOnL2Block = () => {
  const [blockNumber, setBlockNumber] = useState(BigInt(0))

  const subscribe = useCallback(() => {
    return watchBlockNumber(config.getClient(), {
      onBlockNumber: (blockNumber) => {
        setBlockNumber(blockNumber)
      },
    })
  }, [])

  useEffect(() => {
    const unsubscribe = subscribe()

    return () => unsubscribe()
  }, [subscribe])

  return { blockNumber }
}

export const useAccountData = () => {
  const queryClient = useQueryClient()
  const { address } = useAccount()

  const { blockNumber } = useOnL2Block()

  // const { data: blockNumber } = useBlockNumber({ chainId: activeChains.l2.id, watch: watch })
  const {
    data: accountData,
    isLoading,
    queryKey,
  } = useReadContract({
    address: env.NEXT_PUBLIC_CONTRACT as `0x${string}`,
    abi: fooAbi,
    functionName: "fooFunction",
  })

  useEffect(() => {
    void queryClient.invalidateQueries({ queryKey })
  }, [blockNumber])

  return { accountData, isLoading }
}
```

Screenshot 2024-02-29 at 12 54 16 PM

As another side note, what's the correct pattern to ensure there's only one bock subscription in the app so i'm not making unnecessary rpc requests?

_No response_

### Steps To Reproduce

_No response_

### Wagmi Version

2.6.5

### Viem Version

2.7.16

### TypeScript Version

5.3.3

### Check existing issues

### Anything else?

_No response_

Yuripetusko:
## Description

fixes #3656  When using `useBlockNumber` with websocket transport, the `onBlockNumber` passed to useEffect of `useWatchBlockNumber` (as property of `...rest` object), react's Object.equal would consider them to be different and run the effect every time new block arrives, causing multiple subcriptions to the same block. `watchBlockNumber` action does have code like this `if (unwatch) unwatch()` but it doesn't make sense as new `watchBlockNumber` is called every time, so `unwatch` will always be undefined, unless we define `unwatch` outside of the function, but then would need to store instances by some unique keys to only unwatch if the same subscription is being created.

The code is not ideal, and adds a bit too much probably. My head is wonky after trying to debug this for a few hours. Let me know where I can improve it, or I might review it after some rest, and come up with ways to make it smaller. Otherwise if you think this is fine as is, then that's cool too 😄 

One test snapshot for `getBlock` is failing for me, but it also fails on master with no changes. I made sure that .env file is the same as in contribution docs, so not sure what is it, perhaps the snapshot was made with different block number from the one specified in sample .env

image

## Additional Information

### DIFF
--- a/packages/react/src/hooks/useBlockNumber.ts
+++ b/packages/react/src/hooks/useBlockNumber.ts
@@ -18,7 +18,7 @@ import {
   type GetBlockNumberQueryKey,
   getBlockNumberQueryOptions,
 } from '@wagmi/core/query'
-
+import { useRef } from 'react'
 import type { ConfigParameter, QueryParameter } from '../types/properties.js'
 import { type UseQueryReturnType, useQuery } from '../utils/query.js'
 import { useChainId } from './useChainId.js'
@@ -76,6 +76,11 @@ export function useBlockNumber<
     chainId,
   })
 
+  // memoize this function for deepEqual fn.constructor comparison to work
+  const onBlockNumber = useRef((blockNumber: bigint) => {
+    queryClient.setQueryData(options.queryKey, blockNumber)
+  })
+
   useWatchBlockNumber({
     ...({
       config: parameters.config,
@@ -86,9 +91,7 @@ export function useBlockNumber<
       (query.enabled ?? true) &&
         (typeof watch === 'object' ? watch.enabled : watch),
     ),
-    onBlockNumber(blockNumber) {
-      queryClient.setQueryData(options.queryKey, blockNumber)
-    },
+    onBlockNumber: onBlockNumber.current,
   })
 
   return useQuery({ ...query, ...options })
--- a/packages/react/src/hooks/useWatchBlockNumber.ts
+++ b/packages/react/src/hooks/useWatchBlockNumber.ts
@@ -4,12 +4,13 @@ import {
   type Config,
   type ResolvedRegister,
   type WatchBlockNumberParameters,
+  type WatchBlockNumberReturnType,
   deepEqual,
   watchBlockNumber,
 } from '@wagmi/core'
 import { type UnionEvaluate, type UnionPartial } from '@wagmi/core/internal'
 import { useEffect, useRef } from 'react'
-import type { WatchBlockNumberReturnType } from 'viem'
+
 import type { ConfigParameter, EnabledParameter } from '../types/properties.js'
 import { useChainId } from './useChainId.js'
 import { useConfig } from './useConfig.js'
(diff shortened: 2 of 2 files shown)


Yuripetusko:
Thanks for testing [wkyleg.eth](https://github.com/wkyleg)! As discussed in the issue thread.

I think the most important edge case to test is switching the network, but to test this well, we need to configure multiple chains with websocket transport and don't explicitly pass chainId to the hook (let it take one from `useChainId` internally. And then observe if subscription switches to the new chain and old subscription is stopped. 

I'll look into other comments you left on the issue page right now 👍

wkyleg:
Yeah my only reservation about this approach is that I think it should either be fixed in `packages/core/src/actions/watchBlockNumber.ts` or in Viem itself, the issue with deep equality is probably present as well. 

My impression of the observable function in Viem is that it is supposed to fix issues like this

Yuripetusko:
> Yeah my only reservation about this approach is that I think it should either be fixed in `packages/core/src/actions/watchBlockNumber.ts` or in Viem itself, the issue with deep equality is probably present as well.
> 
> My impression of the observable function in Viem is that it is supposed to fix issues like this

Yeah let me check that viem's observe to see if I can use that here

wkyleg:
> > Yeah my only reservation about this approach is that I think it should either be fixed in `packages/core/src/actions/watchBlockNumber.ts` or in Viem itself, the issue with deep equality is probably present as well.
> > My impression of the observable function in Viem is that it is supposed to fix issues like this
> 
> Yeah let me check that viem's observe to see if I can use that here

I'm testing if I can fix this inside Viem so WAGMI changes aren't needed

Yuripetusko:
> > > Yeah my only reservation about this approach is that I think it should either be fixed in `packages/core/src/actions/watchBlockNumber.ts` or in Viem itself, the issue with deep equality is probably present as well.
> > > My impression of the observable function in Viem is that it is supposed to fix issues like this
> > 
> > 
> > Yeah let me check that viem's observe to see if I can use that here
> 
> I'm testing if I can fix this inside Viem so WAGMI changes aren't needed

```
/**
 * @description Sets up an observer for a given function. If another function
 * is set up under the same observer id, the function will only be called once
 * for both instances of the observer.
 */
```

Ok I see, so basically because the current wagmi hook opens 2 wss subscriptions (one for a single fetch using useQuery, and another is an observer that does `queryClient.setQueryData` when new block arrives. So if we use `observe` from the get go, we can make sure that there's just 1 connection open. Is that what you mean?

wkyleg:
> > > > Yeah my only reservation about this approach is that I think it should either be fixed in `packages/core/src/actions/watchBlockNumber.ts` or in Viem itself, the issue with deep equality is probably present as well.
> > > > My impression of the observable function in Viem is that it is supposed to fix issues like this
> > > 
> > > 
> > > Yeah let me check that viem's observe to see if I can use that here
> > 
> > 
> > I'm testing if I can fix this inside Viem so WAGMI changes aren't needed
> 
> ```
> /**
>  * @description Sets up an observer for a given function. If another function
>  * is set up under the same observer id, the function will only be called once
>  * for both instances of the observer.
>  */
> ```
> 
> Ok I see, so basically because the current wagmi hook opens 2 wss subscriptions (one for a single fetch using useQuery, and another is an observer that does `queryClient.setQueryData` when new block arrives. So if we use `observe` from the get go, we can make sure that there's just 1 connection open. Is that what you mean?

Exactly. I think the Viem case should work. 

We should probably have some better test cases for all of this though

jxom:
Fixed via Viem. Thanks @wkyleg!

Yuripetusko:
Not actually sure if that viem fix alone was enough. When I was testing it and manually applying @wkyleg fix, it did reduce the number of websocket connections, but there was still a problem, which I addressed in this PR. My memory is a bit hazy of that code though now, will try again sometimes during this week

jxom:
Think https://github.com/wevm/viem/commit/a4782bb34924bb30e25d07a7efccb4067110b502 would have fixed the other part of it.

Yuripetusko:
yeah still happening for me, will create a stackblitz

Yuripetusko:
You can observe it here in network tab: https://stackblitz.com/edit/new-wagmi-jfbebm?file=src/App.tsx,src/wagmi.ts,package.json cc @jxom

PR #3723, opened 2024-03-19
fix: @wagmi/cli optimize typescript judgment

### LINKED ISSUE #3721: The generated.js gen by @wagmi/cli has an address type error
### Describe the bug

I used the @wagmi/cli output generated.js and found that there is a type error in it,  probably as shown in the code below .
```typescript
// generated.js
export const usdcAddress = {
  42161: '0x629D****',
  421614: '0x629D****',
} 
// ...
export const useReadUsdcDecimals = /*#__PURE__*/ createUseReadContract({
  abi: usdcAbi,
/** 
Type '{ 42161: string; 421614: string; }' is not assignable to type '`0x${string}` | Record<number, `0x${string}`> | undefined'.
*/
  address: usdcAddress,  
  functionName: 'decimals',
})
``` 

 I think it's an issue with the generator, and if the check is indeed the case, I'll try to submit a PR

_No response_

### Steps To Reproduce

_No response_

### Wagmi Version

2.5.7

### Viem Version

2.8.6

### TypeScript Version

5.3.2

### Check existing issues

### Anything else?

_No response_

leecobaby:
## Description

fix #3721 
Optimized for monorepo and special directory structured projects, the user can specify the type of project.

## Additional Information

Added an option for config
### Example
```ts
import { defineConfig } from '@wagmi/cli'

export default defineConfig({
  out: 'src/generated.ts',
  isTypeScript: true,  // new
  contracts: [],
  plugins: [],
})
```        

### DIFF
--- a/.changeset/kind-laws-divide.md
+++ b/.changeset/kind-laws-divide.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/cli": patch
+---
+
+Broadened TypeScript detection.
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -55,7 +55,7 @@ export async function generate(options: Generate = {}) {
   }
 
   const resolvedConfigs = await resolveConfig({ configPath })
-  let isTypeScript = await getIsUsingTypeScript()
+  const isTypeScript = await getIsUsingTypeScript()
 
   type Watcher = FSWatcher & { config?: Watch }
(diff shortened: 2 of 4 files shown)


tmm (approved):
Thanks! Let's start with adding some more known `tsConfig` file names, and can add an explicit flag if necessary. Also, worth noting that if you use a `*.ts` Wagmi CLI config file, like `wagmi.config.ts`, then you are also opted in.

PR #3729, opened 2024-03-19
Fix return type of useWatchContractEvent which should be unsubscribe

Yuripetusko:
As per docs, the reutnr type of this hook should be unsubscribe

https://wagmi.sh/react/api/hooks/useWatchContractEvent#return-type

## Description

Bug. Returning `unsubscribe` correctly from `useWatchContractEvent` instead of void

## Additional Information


Yuripetusko:
Closing, because I noticed that 2 other wss connections actually have `eth_unsubscribe` call. So it's all good

Issue #3738, opened 2024-03-21
bug: unnecessary query invalidations in `useWalletClient` & `useConnectorClient`


BrickheadJohnny:
### Describe the bug

vm! 👋 
We've started using wagmi v2 in a bigger project and noticed that some hooks cause too many re-renders, which leads to poor application performance. Both @Yuripetusko and I tried to find the root of the issue earlier today, and turns out the problematic hooks are `useWalletClient` & possibly `useConnectorClient`. Both of these contain a `useEffect` like this:

```ts
useEffect(() => {
    // invalidate when address changes
    if (address) queryClient.invalidateQueries({ queryKey })
    else queryClient.removeQueries({ queryKey }) // remove when account is disconnected
}, [address, queryClient])
```
This looks fine at first glance, but this logic runs every single time when a hook gets called for the first time inside a component, so essentially if we mount a component which contains the hook, it'll invalidate the query.

### Link to Minimal Reproducible Example

https://codesandbox.io/p/sandbox/wagmi-use-wallet-client-2mfmgd

### Steps To Reproduce

1. Open the provided Codesandbox
2. Connect your wallet
3. Click on the "Re-render child component" button (one or more times)

You'll notice that the WalletClient UID changes every time the child component's `key` changes.

### Wagmi Version

2.5.11

### Viem Version

2.8.14

### TypeScript Version

5.4.2

### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/wagmi/issues) for the bug I encountered.

### Anything else?

We also tried to find a quick solution for this issue & figured I could post it here, maybe it'll serve as an inspiration for the final solution. So, I essentially stored the value of the `account` variable in a `ref`, and since refs don't trigger re-renders, I could use it to compare the `account` with its previous value inside this `useEffect`: 

```ts
const prevAddress = useRef<UseAccountReturnType<config>["address"]>(address)

useEffect(() => {
  if (!address) {
    // remove when account is disconnected 
    queryClient.removeQueries({ queryKey })
  } else if (address !== prevAddress.current) {
    // invalidate when address changes
    queryClient.invalidateQueries({ queryKey })
  }

  prevAddress.current = address
}, [address, queryClient])
```

tmm:
Great info. Definitely an area for improvement.

BrickheadJohnny:
> Why assign the value if it's never even used

[...]

PR #3740, opened 2024-03-21
fix `useWalletClient` and `useConnectorClient` performance issues

BrickheadJohnny:
## Description

Fixes the performance issues discussed in #3738 
Thank you for all the help @Yuripetusko!

## Additional Information

### DIFF
--- a/.changeset/olive-ravens-eat.md
+++ b/.changeset/olive-ravens-eat.md
@@ -1,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+fix `useWalletClient` and `useConnectorClient` performance issues
--- a/.changeset/olive-ravens-eat.md
+++ b/.changeset/olive-ravens-eat.md
@@ -2,4 +2,4 @@
 "wagmi": patch
 ---
 
-fix `useWalletClient` and `useConnectorClient` performance issues
+Removed unnecessary re-renders from  `useConnectorClient` and `useWalletClient`.
--- a/packages/react/src/hooks/useConnectorClient.ts
+++ b/packages/react/src/hooks/useConnectorClient.ts
@@ -83,11 +83,12 @@ export function useConnectorClient<
   const addressRef = useRef(address)
   // biome-ignore lint/nursery/useExhaustiveDependencies: `queryKey` not required
   useEffect(() => {
-    if (!address) {
+    const previousAddress = addressRef.current
+    if (!address && previousAddress) {
       // remove when account is disconnected
       queryClient.removeQueries({ queryKey })
-      if (addressRef.current) addressRef.current = undefined
-    } else if (address !== addressRef.current) {
+      addressRef.current = undefined
+    } else if (address !== previousAddress) {
       // invalidate when address changes
       queryClient.invalidateQueries({ queryKey })
       addressRef.current = address
--- a/packages/react/src/hooks/useConnectorClient.ts
+++ b/packages/react/src/hooks/useConnectorClient.ts
@@ -22,7 +22,7 @@ import {
   type UseQueryReturnType,
   useQuery,
 } from '../utils/query.js'
-import { type UseAccountReturnType, useAccount } from './useAccount.js'
+import { useAccount } from './useAccount.js'
 import { useChainId } from './useChainId.js'
 import { useConfig } from './useConfig.js'
 
@@ -63,9 +63,9 @@ export function useConnectorClient<
 >(
   parameters: UseConnectorClientParameters<config, chainId, selectData> = {},
 ): UseConnectorClientReturnType<config, chainId, selectData> {
-  const { query = {} } = parameters
+  const { query = {}, ...rest } = parameters
 
-  const config = useConfig(parameters)
+  const config = useConfig(rest)
   const queryClient = useQueryClient()
   const { address, connector, status } = useAccount()
   const chainId = useChainId()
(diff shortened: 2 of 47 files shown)


packages/react/src/hooks/useConnectorClient.ts
```diff
@@ -80,11 +80,22 @@ export function useConnectorClient<
...
   const enabled = Boolean(status !== 'disconnected' && (query.enabled ?? true))
 
+  const addressRef = useRef<UseAccountReturnType<config>['address']>(address)
+
   // biome-ignore lint/nursery/useExhaustiveDependencies: `queryKey` not required
   useEffect(() => {
-    // invalidate when address changes
-    if (address) queryClient.invalidateQueries({ queryKey })
-    else queryClient.removeQueries({ queryKey }) // remove when account is disconnected
+    if (!address) {
```
Yuripetusko, line 87: wouldn't this always removeQueries on page refresh? Not sure if there are edge cases like offline-first apps etc. Although perhaps re-initialising new walletClient on page refresh is a good thing

BrickheadJohnny: Yup, good point & I think it is actually okay to revalidate on refresh

Yuripetusko: Yeah I suppose you are right, client is not that critical to re-create on page load and forget the old one. As oppose to things like account, that is more important to get from persistent storage for offline-first apps

jxom: idk if we should be removing queries on refresh if we don't need to

Yuripetusko: ok yea, in this case @BrickheadJohnny perhaps you can check for `!!addressRef.current` ? and I would probably remove it from useRef initialisation, this feels like can lead to weird behaviour. I think it's better to have it null or undefined by default. wdyt?

packages/react/src/hooks/useConnectorClient.ts
```diff
@@ -80,11 +80,22 @@ export function useConnectorClient<
   })
   const enabled = Boolean(status !== 'disconnected' && (query.enabled ?? true))
 
+  const addressRef = useRef<UseAccountReturnType<config>['address']>(address)
```
Yuripetusko, line 83:
```suggestion
  const addressRef = useRef<UseAccountReturnType<config>['address']>()
```

packages/react/src/hooks/useConnectorClient.ts
Yuripetusko, line 87:
```suggestion
    if (!address && !!addressRef.current) {
```

Yuripetusko: Actually I take it back

BrickheadJohnny:
Hey @tmm, according to GitHub, you are the code owner of the `useWalletClient.test.ts` & `useConnectorClient.test.ts` files, could I ask for some guidance here please? 👀 
I'd like to add tests for this bugfix, but wonder how could we re-create this case which I described in the original issue description (or basically the one which you could try in the linked Codesandbox) with the `renderHook` function? 🤔 So the test should look something like this:

1. render the `useWalletClient` hook
2. render a child component, which also calls this hook
3. WalletClient's `uid` shouldn't change after the child (re)renders

If I just call the `renderHook` function 2 times, it'll wrap the 2 hooks in 2 different `WagmiProvider`s, so that wouldn't really work. Should we create & mount actual components in tests with the [render](https://testing-library.com/docs/react-testing-library/api#render) method? (I haven't tried that solution yet, because after all wagmi is a hooks library, and I see that you don't do that in other tests)

BrickheadJohnny:
> Hey @tmm, according to GitHub, you are the code owner of the `useWalletClient.test.ts` & `useConnectorClient.test.ts` files, could I ask for some guidance here please? 👀 I'd like to add tests for this bugfix, but wonder how could we re-create this case which I described in the original issue description (or basically the one which you could try in the linked Codesandbox) with the `renderHook` function? 🤔 So the test should look something like this:
> 
> 1. render the `useWalletClient` hook
> 2. render a child component, which also calls this hook
> 3. WalletClient's `uid` shouldn't change after the child (re)renders with a different `key`
> 
> If I just call the `renderHook` function 2 times, it'll wrap the 2 hooks in 2 different `WagmiProvider`s, so that wouldn't really work. Should we create & mount actual components in tests with the [render](https://testing-library.com/docs/react-testing-library/api#render) method? (I haven't tried that solution yet, because wagmi is a hooks library after all, and I see that you don't do that in other tests)

I ended up rendering components & basically re-created the behavior which I mentioned in the original issue. I'm pretty happy with it (except that `wait(200)`), but please let me know if you think there's a simper solution for this ;)

I also fixed the `useWalletClient` tests, so now they actually test the proper hook (we talked about those on Discord yesterday)

packages/react/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": "./tsconfig.build.json",
-  "include": ["src/**/*.ts", "test/**/*.ts"],
+  "include": ["src/**/*.ts?(x)", "test/**/*.ts?(x)"],
```
Yuripetusko, line 3: Was testing this, and this pattenr matching wasn't working well for me (had ts error), but this worked:

```
"include": ["src/**/*.ts","src/**/*.tsx", "test/**/*.ts", "test/**/*.tsx"],
  "compilerOptions": {
    "jsx": "preserve"
  },
```

Yuripetusko:
Just ran these tests with and without your fix, and indeed, without the fix, the uuid of wallet client in re-rendered component differs, but with fix, it's the same. 👍

packages/react/tsconfig.json
BrickheadJohnny: Hmm, interesting, changed it in 88ec9072e3b7b74d6b84aeecc3e866d7296db239 to make sure it works for everyone, thanks!

.changeset/olive-ravens-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+fix `useWalletClient` and `useConnectorClient` performance issues
```
tmm, line 5:
```suggestion
Removed unnecessary re-renders from  `useConnectorClient` and `useWalletClient`.
```

PR #3777, opened 2024-04-01
Fix `writeContract` `chainId` forwarding during contract call simulation

desfero:
## Description

`chainId` was not forwarded to the simulation call which was leading to calling wrong RPC node


packages/core/src/actions/writeContract.ts
```diff
@@ -106,6 +106,7 @@ export async function writeContract<
   else {
     const { request: simulateRequest } = await simulateContract(config, {
       ...rest,
+      chainId,
       account,
```
tmm, line 110:
```suggestion
      account,
      chainId,
```

PR #3784, opened 2024-04-03
feat: @coinbase/wallet-sdk upgrade

nateReiners:
## Description

- This is just a draft to show what changes are needed in order to upgrade and get the ball rolling
- Upgrade `@coinbase/wallet-sdk` from v3 to v4
- v4 is almost ready and includes Coinbase smart wallets!

### Bundle size has been cut in half!
https://bundlephobia.com/package/@coinbase/wallet-sdk@4.0.0 


packages/connectors/package.json
```diff
@@ -45,7 +45,7 @@
     }
   },
   "dependencies": {
-    "@coinbase/wallet-sdk": "3.9.1",
+    "@coinbase/wallet-sdk": "4.0.0-beta.2",
```
nateReiners, line 48: TODO: upgrade to v4 latest when published

jxom:
What are the breaking changes in v4?

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -12,43 +10,14 @@ import {
...
-    reloadOnDisconnect?: boolean | undefined
-  }
->
-
 coinbaseWallet.type = 'coinbaseWallet' as const
-export function coinbaseWallet(parameters: CoinbaseWalletParameters) {
-  const reloadOnDisconnect = false
-
-  type Provider = CoinbaseWalletProvider
+export function coinbaseWallet(parameters: CoinbaseWalletSDKOptions) {
```
tmm, line 14: Let's keep `CoinbaseWalletParameters` around, even if it just proxies `CoinbaseWalletSDKOptions`. Gives us more control over the connector's public API.

nateReiners:
> What are the breaking changes in v4?

Hey @jxom , For most apps there won't be any necessary code changes other than a version bump. However we did remove a bunch of optional SDK options so apps that were passing in any of those will need to remove them. Details [here](https://github.com/coinbase/coinbase-wallet-sdk/blob/master/packages/wallet-sdk/docs/migration_guide.md)

arein:
We need a `disableSmartWallet` option please - also can you please provide a bundle size comparison @nateReiners ? we have thousands of apps using web3modal and we serve your SDK

glitch-txs:
hi @nateReiners we are also using `enableMobileWalletLink` in Web3Modal, do you know if this would cause some sort of error or conflict in the SDK? It would be a bit tricky for us since Wagmi is just a peer dependency and we cannot control which version devs are using.

nateReiners:
> hi @nateReiners we are also using `enableMobileWalletLink` in Web3Modal, do you know if this would cause some sort of error or conflict in the SDK? It would be a bit tricky for us since Wagmi is just a peer dependency and we cannot control which version devs are using.

Hey @glitch-txs `enableMobileWalletLink` is gone, but the ability to connect via mobile is enabled by default in v4. When the SDK receives an `eth_requestAccounts` request, it will open a popup that allows the user to choose between connecting via smart wallet or Coinbase Wallet mobile. You can test out the UX for smart wallet or mobile wallet link here: https://www.smart-wallet.xyz/ Feedback welcome 

To remove the option to connect via mobile, dapps can pass in `smartWalletOnly: true` as an optional SDK option

wilsoncusack:
> We need a `disableSmartWallet` option please - also can you please provide a bundle size comparison @nateReiners ? we have thousands of apps using web3modal and we serve your SDK

Hey Derek can you say more why the disable option is needed?

glitch-txs:
@nateReiners thanks for the quick response! My question is if the SDK will throw an error if I pass it down the `enableMobileWalletLink` option in v4. In which case it would break Web3Modal if the dev is using an "old" version.

nateReiners:
> also can you please provide a bundle size comparison

@arein Yep, just added it to the PR description. gzipped size was reduced by ~20 KB 🎉

jxom:
> We need a disableSmartWallet option please - also can you please provide a bundle size comparison @nateReiners ? we have thousands of apps using web3modal and we serve your SDK

Curious why you'd want to disable smart wallets? Shouldn't Web3Modal be agnostic in this sense?

nateReiners:
> @nateReiners thanks for the quick response! My question is if the SDK will throw an error if I pass it down the `enableMobileWalletLink` option in v4. In which case it would break Web3Modal if the dev is using an "old" version.

No the SDK wont throw an error. All deprecated options including `enableMobileWalletLink` are just ignored in v4. It would just be a Typescript issue.

arein:
@jxom @wilsoncusack I can imagine some dapps not wanting this so I think it makes sense to allow them to configure. What rationale went into the option to exclusively use the smart wallet.

glitch-txs:
Hi Coinbase team! Is this issue solved here? Also curious why the SDK is using the Solana package, is it meant to be multi-chain?
https://github.com/wevm/wagmi/issues/3836

nateReiners:
Hey just wanted to drop an update here since I know this has been sitting for some time. We have been making some SDK interface tweaks and I wanted to wait until all of those decisions are solidified before cleaning up this PR. 

We are adding the ability to disable smart wallet as part of these tweaks via an 'eoaOnly' option cc: @arein

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -134,10 +111,10 @@ export function coinbaseWallet(parameters: CoinbaseWalletParameters) {
     },
     async getChainId() {
       const provider = await this.getProvider()
-      const chainId = await provider.request<number>({ method: 'eth_chainId' })
```
bangtoven, line 137: do we need this change?

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -122,7 +100,6 @@ export function coinbaseWallet(parameters: CoinbaseWalletParameters) {
       }
 
       provider.disconnect()
-      provider.close()
```
bangtoven, line 125: what if `provider`'s version is 3.x?

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -85,7 +63,7 @@ export function coinbaseWallet(parameters: CoinbaseWalletParameters) {
         }
 
         // Switch to chain if provided
-        let currentChainId = await this.getChainId()
+        let currentChainId: number = await this.getChainId()
```
bangtoven, line 66: is this necessary?

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -1,268 +1,226 @@
...
-     * Fallback Ethereum Chain ID
-     * @default 1
-     */
-    chainId?: number | undefined
-    /**
-     * Whether or not to reload dapp automatically after disconnect.
-     * @default false
-     */
-    reloadOnDisconnect?: boolean | undefined
+  Mutable<Omit<ConstructorParameters<typeof CoinbaseWalletSDK>[0], "appChainIds">> & {
```
bangtoven, line 14:
```suggestion
  Mutable<
    Omit<
      ConstructorParameters<typeof CoinbaseWalletSDK>[0],
      'appChainIds' // remove property and use `config.chains`
    >
  > & {
```

bangtoven (commented):
nit: to minimize diff

packages/connectors/src/coinbaseWallet.ts
nateReiners: good catch, addressed in [886fb51](https://github.com/wevm/wagmi/pull/3784/commits/886fb51fd306b14e81ec5b1d3afeac2d00a42c15)

packages/connectors/src/coinbaseWallet.ts
nateReiners: no, removed

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -17,37 +14,22 @@ import {
...
-    chainId?: number | undefined
-    /**
-     * Whether or not to reload dapp automatically after disconnect.
-     * @default false
-     */
-    reloadOnDisconnect?: boolean | undefined
-  }
+      'appChainIds' // set via wagmi config
+    > & {
+      preference?: Preference
```
tmm, line 25: Let's inline `Preference`

```suggestion
      preference?: Parameters<CoinbaseWalletSDK['makeWeb3Provider']>[0] | undefined
```

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -17,37 +14,22 @@ import {
   numberToHex,
 } from 'viem'
 
-// TODO(@3): Set `enableMobileWalletLink` to `true`
+type Preference = Parameters<CoinbaseWalletSDK['makeWeb3Provider']>[0]
+
```
tmm, line 18:
```suggestion
```

tmm (commented):
Docs will need updating too

https://wagmi.sh/react/api/connectors/coinbaseWallet

PR #3831, opened 2024-04-17
feat: update metamask sdk

abretonc7s:
## Description

Update `@metamask/sdk` to 0.18.5 which includes multiple connection and improvements and reduced bundle size.

## Additional Information

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -57,9 +57,6 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         provider.on('connect', this.onConnect.bind(this) as Listener)
     },
     async connect({ chainId, isReconnecting } = {}) {
-      console.log(
-        `MetaMask SDK connect chainId=${chainId} isReconnecting=${isReconnecting}`,
-      )
       const provider = await this.getProvider()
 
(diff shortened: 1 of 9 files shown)


packages/connectors/src/metaMask.ts
```diff
@@ -60,35 +56,16 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         provider.on('connect', this.onConnect.bind(this) as Listener)
     },
     async connect({ chainId, isReconnecting } = {}) {
+      if (isReconnecting && isMobileBrowser) {
+        // Prevent reconnection on mobile, it needs to be done manually via eth_requestAccount + deeplink
```
jxom, line 60: Interesting. Can we get this to work?

deeeed: We cannot at the moment since the wallet needs to be active and online. You need to make a deeplink to wake up the wallet. We are working on updating our protocol to prevent that. The reason for that limitation is that unlike wallet connect and coinbase we don't store encryption keys locally which in turn prevents us from storing rpc message on a relayer without having the wallet online.

jxom: I see.

jxom:
Closed in favor of #3848

PR #3829, opened 2024-04-16
Fix shimDisconnect ignored

mqklin:
Fix https://github.com/wevm/wagmi/issues/3828

## Description

What changes are made in this PR? Is it a feature or a bug fix?

## Additional Information

Before submitting this issue, please make sure you do the following.

- [ ] Read the [contributing guide](https://wagmi.sh/dev/contributing)
- [ ] Added documentation related to the changes made.
- [ ] Added or updated tests (and snapshots) related to the changes made.

PR #3849, opened 2024-04-24
fix: `shimDisconnect: false` behavior

### LINKED ISSUE #3828: shimDisconnect is ignored
### Describe the bug

Ref:
https://github.com/wevm/wagmi/issues/3603

I've put breakpoints inside injected.js file debugger. You can see that `injected` is called two times, and for the second time it's called with shimDisconnect undefined => true:
image

And after that the code that should not be run, is actually run: 
image

I've extensively described the process here: https://github.com/wevm/wagmi/pull/3608

_No response_

### Steps To Reproduce

_No response_

### Wagmi Version

@wagmi/core@2.6.17

### Viem Version

2.9.19

### TypeScript Version

_No response_

### Check existing issues

### Anything else?

_No response_

tmm:
### Description

fixes #3828
closes #3829

### DIFF
--- a/packages/core/src/connectors/injected.ts
+++ b/packages/core/src/connectors/injected.ts
@@ -223,13 +223,13 @@ export function injected(parameters: InjectedParameters = {}) {
           currentChainId = chain?.id ?? currentChainId
         }
 
-        if (shimDisconnect) {
-          // Remove disconnected shim if it exists
+        // Remove disconnected shim if it exists
+        if (shimDisconnect)
           await config.storage?.removeItem(`${this.id}.disconnected`)
-          // Add connected shim if no target exists
-          if (!parameters.target)
-            await config.storage?.setItem('injected.connected', true)
-        }
+
+        // Add connected shim if no target exists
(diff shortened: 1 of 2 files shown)


packages/core/src/connectors/injected.ts
```diff
@@ -223,13 +223,13 @@ export function injected(parameters: InjectedParameters = {}) {
...
+        if (shimDisconnect)
           await config.storage?.removeItem(`${this.id}.disconnected`)
-          // Add connected shim if no target exists
-          if (!parameters.target)
-            await config.storage?.setItem('injected.connected', true)
-        }
+
+        // Add connected shim if no target exists
+        if (!parameters.target)
+          await config.storage?.setItem('injected.connected', true)
```
tmm, line 232: flag needs to be set even if `shimDisconnect: false`

https://github.com/wevm/wagmi/blob/f612323f2dd9a80cc5be43f0bc9fd08c31aeb65c/packages/core/src/connectors/injected.ts#L316

PR #3859, opened 2024-04-25
fix: metamask chain switch with missing net_version

holic:
### Description

This is a workaround for the issue described in https://github.com/MetaMask/metamask-extension/issues/24247.

I've left a big inline comment explaining everything.

### DIFF
--- a/packages/core/src/connectors/injected.ts
+++ b/packages/core/src/connectors/injected.ts
@@ -388,13 +388,11 @@ export function injected(parameters: InjectedParameters = {}) {
               method: 'wallet_switchEthereumChain',
               params: [{ chainId: numberToHex(chainId) }],
             })
-            // MetaMask makes a `net_version` RPC call to the target chain during `wallet_switchEthereumChain`.
-            // If this request fails, MetaMask doesn't emit a `chainChanged` event, but will still switch the chain.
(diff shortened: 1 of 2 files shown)


packages/core/src/connectors/injected.ts
```diff
@@ -381,21 +380,34 @@ export function injected(parameters: InjectedParameters = {}) {
...
               }
-            }
-            provider.on('chainChanged', listener)
-          }),
+            }),
+          new Promise<void>((resolve) =>
+            config.emitter.once('change', ({ chainId: currentChainId }) => {
+              if (currentChainId === chainId) resolve()
+            }),
+          ),
```
holic, line 410: I switched this to use the internal emitter rather than binding to the provider's emitter so that we can emit our own `changed` event in the workaround case. This also matches the MetaMask connector:
https://github.com/wevm/wagmi/blob/72a25ee00a7b6b1b41c1a4825d08440a852f9057/packages/connectors/src/metaMask.ts#L180-L184

holic: I believe that either approach can leave this listener open indefinitely if we never get a chain switch event. Should we time out after some time and remove the listener?

PR #3868, opened 2024-04-28
feat: connector property to skip contract simulation

jxom:
### Description

Adds a property to the connector to indicate whether it supports contract simulation. We can defer contract write simulation to the connector's wallet if it can accurately simulate the contract write & display contract revert reasons correctly. 

As a result, the time between user interaction and the wallet opening[^1] will become nearly instantaneous. 

This means it will resolve the issue of iOS Safari popup & universal link dismissal[^2] which some connectors rely on (ie. Coinbase Smart Wallet & the MetaMask SDK), provided that their wallets can accurately simulate contract write calls.

[^1]: TTOW: Time To Open Wallet.
[^2]: These dismissals occur if the intent does not get invoked within ~500ms of user interaction.

### DIFF
--- a/.changeset/twelve-walls-deny.md
+++ b/.changeset/twelve-walls-deny.md
@@ -4,4 +4,4 @@
 "wagmi": minor
 ---
 
-Added `supportsSimulation` property to connectors.
+Added `supportsSimulation` property to connectors that indicates if the connector's wallet supports contract simulation.
--- a/.changeset/twelve-walls-deny.md
+++ b/.changeset/twelve-walls-deny.md
@@ -1,0 +1,7 @@
+---
+"@wagmi/connectors": minor
+"@wagmi/core": minor
+"wagmi": minor
+---
+
+Added `skipSimulateContract` property to connectors.
--- a/.changeset/twelve-walls-deny.md
+++ b/.changeset/twelve-walls-deny.md
@@ -4,4 +4,4 @@
 "wagmi": minor
 ---
 
-Added `skipSimulateContract` property to connectors.
+Added `supportsSimulation` property to connectors.
--- a/packages/core/src/actions/sendTransaction.ts
+++ b/packages/core/src/actions/sendTransaction.ts
@@ -82,7 +82,7 @@ export async function sendTransaction<
     if (!('data' in parameters) || !parameters.data) return undefined
 
     // Skip gas estimation if connector specifies.
-    if (activeConnector?.skipSimulateContract) return undefined
+    if (activeConnector?.supportsSimulation) return undefined
 
     // Skip gas estimation if `null` is provided.
     if (gas_ === null) return undefined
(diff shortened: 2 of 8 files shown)


packages/core/src/connectors/injected.ts
```diff
@@ -34,6 +34,9 @@ export type InjectedParameters = {
   target?: TargetId | Target | (() => Target | undefined) | undefined
 }
 
+// Regex of wallets/providers that can accurately simulate contract calls & display contract revert reasons.
+const skipSimulateContractRegex = /(rabby|trustwallet)/
```
jxom, line 38: Sadly, from my testing of the top wallets, only Rabby and Trust can accurately simulate contract writes.

packages/core/src/actions/sendTransaction.ts
```diff
@@ -74,7 +75,15 @@ export async function sendTransaction<
   else
     client = await getConnectorClient(config, { account, chainId, connector })
 
+  const { connector: activeConnector } = getAccount(config)
+
   const gas = await (async () => {
+    // Skip gas estimation if `data` doesn't exist (likely not a contract interaction).
+    if (!('data' in parameters) || !parameters.data) return undefined
```
jxom, line 82: We should skip estimating gas if we know that the transaction is not a contract interaction as all wallets will disallow value transfer if the user does not have enough funds.

tmm (approved):
love it!

.changeset/twelve-walls-deny.md
```diff
@@ -0,0 +1,7 @@
+---
+"@wagmi/connectors": minor
```
tmm, line 2: When the Changesets version PR is created, double check that `@wagmi/connectors` doesn't get bumped to `5.0.0` (has in the past, need to take a look at the Changesets config to figure out why).

packages/core/src/actions/writeContract.ts
```diff
@@ -101,8 +102,11 @@ export async function writeContract<
...
     client = await getConnectorClient(config, { account, chainId, connector })
 
+  const { connector: activeConnector } = getAccount(config)
+
   let request
-  if (__mode === 'prepared') request = rest
+  if (__mode === 'prepared' || activeConnector?.supportsSimulation)
+    request = rest
   else {
     const { request: simulateRequest } = await simulateContract(config, {
```
lukaw3d, line 111: This looks wrong

jxom: ?

lukaw3d: If a connector supports simulation, then this skips it

jxom: Correct

jxom: To be clear, we want to skip simulation if the Wallet supports it.

- If Wallet **does** support simulation, then Wagmi **skips the simulation** since the Wallet perform simulation instead.
- If Wallet **does not** support simulation, then Wagmi **performs the simulation**.

PR #3873, opened 2024-05-01
fix: TrustWallet getAccounts error

oktapodia:
### Description

It seems that TrustWallet returns an error instead of a `[]` on some `getAccounts` call, this PR should fallback to `[]` all the time instead of throwing an error.
image


dalechyn (changes requested):
can we instead raise an upstream issue in Trust Wallet?

tmm:
better to raise this upstream like @dalechyn mentioned

PR #3906, opened 2024-05-09
feat: @wagmi/vue

tmm:
### Description

Adds `feat: @wagmi/vue` package


packages/vue/src/composables/useBlockNumber.ts
```diff
@@ -75,7 +75,7 @@ export function useBlockNumber<
 
   const queryOptions = computed(() => {
     const {
-      chainId = configChainId.value,
+      chainId = configChainId,
```
tmm, line 78: not reactive unless `.value` is removed, but messes with the type for `get*QueryOptions`

tmm: need to figure out a nice way to use existing `get*QueryOptions` while also unwrapping these type of framework-specific values (in this case `ref`)

Related https://github.com/TanStack/query/pull/7390

tmm: resolved https://github.com/wevm/wagmi/pull/3906/commits/ff73c95510ae256febf693f58a5bdb30214964f8

reslear:
wow amazing

tmm:
@reslear agree!

PR #3919, opened 2024-05-11
feat: support Vue2.x

unicape:
Currently, Vue2.x has 500,000+ downloads per week, which is a very large group.

When I was developing use-wagmi, many people wanted to get support from Vue2.x.

The cost for us to support Vue2.x is very low

### DIFF
--- a/packages/test/package.json
+++ b/packages/test/package.json
@@ -63,7 +63,7 @@
     "react-dom": ">=18",
     "typescript": ">=5.0.4",
     "viem": "2.x",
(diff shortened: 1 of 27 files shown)


tmm:
Going to release initial Vue support for v3 only since it's been out for a while, has the most downloads, and we like to stay on cutting edge (e.g. we dropped support for React 17 and it has just over 4M downloads/week). If we add support for Vue 2 in the future, we will resurrect this PR and give you credit.

Issue #3635, opened 2024-02-25
bug: waitForTransactionReceipt doesn't accept client with multiple possible chains


LufyCZ:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

2.7.13

### Current Behavior

![image](https://github.com/wevm/viem/assets/20911347/f84f2fa1-86a7-4846-99c2-fdcbced78a79)

### Link to Minimal Reproducible Example (StackBlitz, CodeSandbox, GitHub repo etc.)

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwBYxmAvnAZlCI4DkAbsAKYgEBQoksiA7gIbAwBi0AKlIwHYDOjAMYxgEHgCVSg0sDAxMOPIRLkA9EJFi+VGtHgJGUAEYsoAVxAA5CEUYAaOBDmhgfEAtz5iZEKsFJmfh1wPUQzPlIABTMjABtgQQBheNIeeWxPQiYAcxBgYNp9QShSRhhSRLEsYGyPJQIAARy8v2hSKkpBLXh-QL44AF44AG1DExhzKxt7R2c8twBdOEZ+rv4YTu64Cd4+Qv6hhEo4EbHTC2tbADpgABMFgC44FDQACgAid4BKO2ORpxE8xAN3uTxeYA+30omBWcDWfA2m3WcKqNUGiD+vWA-F+Jx2-H20OWq26lEoMAAnmBSHBKjxqtl0ZIYGYoDwOFTSAAeP6U6kQLBwkplCqo7I8k4nV580gClH0mpfYYELFBBa4yXSzlytYMpUEfF7PTaBZ-AB8lAtZKwZh4wlEPDg4VIAAlSLFYhBXl8MSd4T0Uml0c7onEEskyGkuXSGWbvWSTkwWOwoFxdhoHZJpLIYK9BIGYA5kCskE93gAGAAe73QX2hQA

tmm:
If anyone wants to look into this, this is the test that you need to pass:

`getClient.test-d.ts`
```ts
test('behavior: viem actions', () => {
  const config = createConfig({
    chains: [arbitrumNova, optimism],
    transports: {
      [arbitrumNova.id]: http(),
      [optimism.id]: http(),
    },
  })
  const client = getClient(config)
  waitForTransactionReceipt(client, { hash: '0x…' })
})
```

KedziaPawel:
[...]

PR #3929, opened 2024-05-14
pass client to waitForTransactionReceipt

benschac:
### Description
Looking into `waitForTransactionReceipt` this _seems_ like a no-op.

- adds an example in playground of passing a client with `useWaitForTransactionReceipt`
- passes the test case provided here: https://github.com/wevm/wagmi/issues/3635#issuecomment-2033265084
- remove `getClient.test-d.ts` and move test cases to `getClient.test.ts` locally, vitest wasn't running `getClient.test-d.ts`.
- I didn't update docs because examples show a config being past already. This change just shows it a bit more explicitly.

### Notes
- snapshot tests were failing when initial clone of repo was pulled. I can update them in another PR.
- this is my first _real_ PR here. Forgive me if there's any large gap in context I might be missing.

 

fixes: #3635


playgrounds/vite-react/src/App.tsx
```diff
@@ -138,7 +140,7 @@ function SwitchChain() {
         <button
           disabled={chainId === chain.id}
           key={chain.id}
-          onClick={() => switchChain({ chainId: chain.id })}
+          onClick={() => switchChain({ chainId: chain.id as ChainIds })}
```
benschac, line 143: un-related. Saw an error, fixed an error.

playgrounds/vite-react/src/App.tsx
```diff
@@ -255,8 +259,15 @@ function SendTransaction() {
...
-  const { isLoading: isConfirming, isSuccess: isConfirmed } =
+  const { isLoading: isConfirming, data: receipt } =
     useWaitForTransactionReceipt({
+      config: {
+        ...config,
+        // Additionally, we can pass a client to the config
+        // to use a different client than the one passed to the hook
+        client,
+      },
+      chainId: connectorClientData?.chain?.id,
```
benschac, line 270: assuming this will help someone in the future when they run into / search this issue.

tmm:
Thanks for this! Think I'm going to take this on since there seems to be a larger internal types refactor.

PR #3949, opened 2024-05-20
refactor(react): optimize use connectors in use disconnect hook

jeasonstudio:
### Description

Optimize the way to obtain connectors in `useDisconnect`


tmm:
The reason `useConnections` is used to get the `connectors` is so `connectors` only includes connected `connectors` since calling `disconnect` on unconnected connectors does not work.

PR #3950, opened 2024-05-20
Update createConfig.md

danielsimao:


danielsimao:
No need to have `createStorage` import

Issue #3961, opened 2024-05-21
Safe Connector causes wagmi account to be stuck in connecting state


cbachmeier:
### Describe the bug

Having the [safe connector](https://wagmi.sh/core/api/connectors/safe#safe) included in the `connectors` list of a web app in the wagmi config will cause the wagmi `account.status` to be stuck indefinitely as `connecting` when the web app is iframed.

### Link to Minimal Reproducible Example

https://stackblitz.com/edit/new-wagmi-jhjax3?file=index.html

### Steps To Reproduce

The above minimal setup is just an iframe of a given url. As a default I've included the link to [uniswap](https://app.uniswap.org), which included `safe` as a connector. Looking at the preview you'll see that the connection logic in the top right hand corner while be stuck in the `connecting` status.

### Wagmi Version

2.5.19

### Viem Version

2.x

### Check existing issues

- [X] I checked there isn't [already an issue](https://github.com/wevm/wagmi/issues) for the bug I encountered.

PR #3962, opened 2024-05-21
feat(safe): add getInfo timeout

tmm:
### Description

Adds timeout `getInfo` called in `safe` connector since [non-Safe App iFrames cause it to not resolve](https://github.com/safe-global/safe-apps-sdk/issues/263#issuecomment-1029835840).

Fixes https://github.com/wevm/wagmi/issues/3961

### DIFF
--- a/packages/connectors/src/safe.ts
+++ b/packages/connectors/src/safe.ts
@@ -20,8 +20,10 @@ export type SafeParameters = Evaluate<
      */
     shimDisconnect?: boolean | undefined
     /**
-     * `getInfo` (from the Safe SDK) does not resolve when not used in Safe App iFrame. This allows the connector to force a timeout.
-     * @default 100
+     * Timeout in milliseconds for `getInfo` (from the Safe SDK) to resolve.
+     *
+     * `getInfo` does not resolve when not used in Safe App iFrame. This allows the connector to force a timeout.
+     * @default 10
      */
     unstable_getInfoTimeout?: number | undefined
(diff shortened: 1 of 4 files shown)


packages/connectors/src/safe.ts
```diff
@@ -19,6 +19,13 @@ export type SafeParameters = Evaluate<
      * @default false
      */
     shimDisconnect?: boolean | undefined
+    /**
+     * Timeout in milliseconds for `getInfo` (from the Safe SDK) to resolve.
+     *
+     * `getInfo` does not resolve when not used in Safe App iFrame. This allows the connector to force a timeout.
+     * @default 10
```
tmm, line 26: tested with `"Good 3G"` throttling on app.safe.global and has no problem resolving.

https://app.safe.global/apps/open?safe=eth%3A0x4557B18E779944BFE9d78A672452331C186a9f48&appUrl=http%3A%2F%2Flocalhost%3A5173

packages/core/src/actions/reconnect.ts
```diff
@@ -65,7 +65,7 @@ export async function reconnect(
   const connections: Connection[] = []
   const providers: unknown[] = []
   for (const connector of sorted) {
-    const provider_ = await connector.getProvider()
+    const provider_ = await connector.getProvider().catch(() => undefined)
```
tmm, line 68: Surprised we weren't doing this before.

PR #3970, opened 2024-05-22
fix cookie path in cookieStorage

nanxiaobei:
If the path of cookie is not set, there will be multiple `wagmi.store` in the cookie (under different paths), and the values are inconsistent. At this point, obtaining the value of the cookie will be uncontrollable.


nanxiaobei:
vSnbR5d

packages/core/src/utils/cookie.ts
```diff
@@ -10,7 +10,7 @@ export const cookieStorage = {
   },
   setItem(key, value) {
     if (typeof window === 'undefined') return
-    document.cookie = `${key}=${value}`
+    document.cookie = `${key}=${value};path=`
```
tmm, line 13: what's the advantage of `path=` over `path=/`?

tmm (changes requested):
Please add a patch changeset with `pnpm changeset`.

packages/core/src/utils/cookie.ts
nanxiaobei: It's a mistake... fixed.

PR #3975, opened 2024-05-23
chore: add store param

nathan1530:
Let `unstable_connector` have ability to use provider from specific connector instead of the first injected provider.

ref:
https://github.com/wevm/wagmi/discussions/3803

### DIFF
--- a/packages/core/src/createConfig.ts
+++ b/packages/core/src/createConfig.ts
@@ -176,7 +176,7 @@ export function createConfig<
         chain,
         batch: properties.batch ?? { multicall: true },
         transport: (parameters) =>
-          rest.transports[chainId]({ ...parameters, connectors }),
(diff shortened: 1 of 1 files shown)


nathan1530:
in my forked repo I have patched wagmi with the following to select the current connector in `unstable_connector`

```
    const request: EIP1193RequestFn = async ({ method, params }) => {
      let connector: Connector | undefined

      const storeState = store?.getState()

      if (preferCurrentConnector && storeState?.current) {
        connector = storeState.connections.get(storeState.current)?.connector
      }

      connector =
        connector ?? connectors?.getState().find((c) => c.type === type)
 ```

tmm:
Thanks for taking a look at this! Going to spend some time noodling on the API design for this.

tmm:
Closing for now and adding to future roadmap.

PR #3978, opened 2024-05-23
Fix: update WalletConnect connector - getNamespaceChainsIds

glitch-txs:
### Description

The `chains` object from the `session` object is an optional property. [See type](https://github.com/WalletConnect/walletconnect-monorepo/blob/022e4d492c9862ab9b17c1a7b12884bdcd992ae4/packages/types/src/sign-client/session.ts#L11). So it's possible to get an `undefined` value when there are actually approved chains in the session. However the `accounts` object is required, so we can rely on it to obtain the approved chains in a session.

### DIFF
--- a/.changeset/dry-penguins-talk.md
+++ b/.changeset/dry-penguins-talk.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/connectors": patch
+---
+
+Fixed walletConnect connector: get chain ids from the accounts object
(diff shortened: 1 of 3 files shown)


ganchoradkov (approved):
Awesome!💯

Why is the diff so large though

glitch-txs:
> https://github.com/wevm/wagmi/pull/3978

When installed the lock file was automatically updated, should I revert? cc @tmm

tmm:
Closing in favor of https://github.com/wevm/wagmi/pull/3979

PR #3997, opened 2024-05-31
cbw sdk version bump

nateReiners:
v4.0.3 fixes the `eth_chainId` response to match the spec. It was returning a number and now returns a hex, in line with https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_chainid

### DIFF
--- a/packages/connectors/src/coinbaseWallet.ts
+++ b/packages/connectors/src/coinbaseWallet.ts
@@ -18,7 +18,6 @@ import {
   type ProviderRpcError,
   SwitchChainError,
   UserRejectedRequestError,
-  fromHex,
   getAddress,
   numberToHex,
 } from 'viem'
--- a/packages/connectors/src/coinbaseWallet.ts
+++ b/packages/connectors/src/coinbaseWallet.ts
@@ -18,6 +18,7 @@ import {
   type ProviderRpcError,
   SwitchChainError,
   UserRejectedRequestError,
+  fromHex,
   getAddress,
   numberToHex,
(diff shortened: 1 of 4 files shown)


packages/connectors/src/coinbaseWallet.ts
```diff
@@ -163,10 +164,10 @@ function version4(parameters: Version4Parameters) {
     },
     async getChainId() {
       const provider = await this.getProvider()
-      const chainId = await provider.request<number>({
+      const chainId = await provider.request<`0x${string}`>({
         method: 'eth_chainId',
       })
-      return Number(chainId)
+      return fromHex(chainId, 'number')
```
jxom, line 170:
```suggestion
      return Number(chainId)
```

packages/connectors/src/coinbaseWallet.ts
```diff
@@ -401,8 +402,10 @@ function version3(parameters: Version3Parameters) {
     },
     async getChainId() {
       const provider = await this.getProvider()
-      const chainId = await provider.request<number>({ method: 'eth_chainId' })
-      return Number(chainId)
+      const chainId = await provider.request<`0x${string}`>({
+        method: 'eth_chainId',
+      })
+      return fromHex(chainId, 'number')
```
jxom, line 408:
```suggestion
      return Number(chainId)
```

nateReiners:
TIL `Number` can convert from hex. Thanks @jxom

tmm (approved):

PR #4009, opened 2024-06-04
fix: Make to optional in sendTransaction

roninjin10:
To should be optional since you can not set it when deploying a contract.

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwFAEgDOBTAdgEwCpQCGWahAxjMBFgDRIC+cAZlBCHAOQDuhA5iMAD05StTQckoSLETYyEHBgAiGMABsIATyWEYhRizacAbsAwgJSeSXhQMaOAF44hHsHiZcBYqQpUsABQIjIQOxJo0iDi6hABccHIKyqoa2jFBIWFYmgCU9DlAA


roninjin10:
In draft because I need to add a changeset when I wake up

PR #4012, opened 2024-06-05
fix: exclude forge-std Mock.sol files by default to prevent duplicate contract names

### LINKED ISSUE #3951: Foundry Plugin Error - Contract Name "IERC721TokenReceiver" Must Be Unique
# Bug description

When setting up a new create wagmi project and configuring Foundry according to the documentation, an error occurs during the wagmi generate command. The error message indicates that the contract name "IERC721TokenReceiver" must be unique.

## Steps to Reproduce

1. Initialize a new `create wagmi` project.
2. Configure Foundry wagmi plugin.
3. Create a new Foundry project using `forge init`.
4. Attempt to generate wagmi artifacts with `wagmi generate`.

```
> wagmi generate
✔ Validating plugins
⠹ Resolving contracts
Contract name "IERC721TokenReceiver" must be unique.
 ELIFECYCLE  Command failed with exit code 1.
```

My Configuration File (wagmi.config.ts)

```typescript
export default defineConfig({
  out: 'src/generated.ts',
  contracts: [],
  plugins: [
    foundry({
      project: '../hello_foundry',
    }),
  ],
})
```

### Cause
The issue is caused by a conflict with forge-std version 1.8.0 or higher, which is installed by default with Foundry.

### Solution
To resolve this issue, i downgraded forge-std to version 1.7.6. Following these steps:

1. Remove the current version of forge-std:
```bash
forge remove foundry-rs/forge-std
git commit -m "Removed foundry-rs/forge-std"
```
2. Install version 1.7.6 of forge-std:
```bash
forge install foundry-rs/forge-std@v1.7.6
forge clean
```

_No response_

### Steps To Reproduce

_No response_

### Wagmi Version

latest

### Viem Version

latest

### TypeScript Version

5.2.2

### Check existing issues

### Anything else?

Let me know if i can help somehow!

eli-lim:
- Fixes #3951 
- Took this approach since the Mock files should probably be excluded anyway
- To prevent this from repeating in the future as `forge-std` drifts, might be a good idea to write a dedicated test to sanity check that the plugin works with `forge init`

### Note 
- An alternative is to use the `Mock*.sol/**` glob pattern, but to prevent some edge-case wonkiness down the line where the user introduces a `MockingBird.sol` or something and it doesn't get picked up

### DIFF
--- a/packages/cli/src/plugins/foundry.ts
+++ b/packages/cli/src/plugins/foundry.ts
@@ -14,6 +14,8 @@ import type { Evaluate, RequiredBy } from '../types.js'
 const defaultExcludes = [
   'Common.sol/**',
   'Components.sol/**',
+  'MockERC20.sol/**',
+  'MockERC721.sol/**',
   'Script.sol/**',
   'StdAssertions.sol/**',
   'StdInvariant.sol/**',


tmm:
> To prevent this from repeating in the future as forge-std drifts, might be a good idea to write a dedicated test to sanity check that the plugin works with forge init

Good idea! Definitely down for this if you want to open a PR.

PR #4020, opened 2024-06-08
refactor(connectors): metaMask mobile reconnect

tmm:
`metaMask` now supports mobile reconnection and uses deeplinks by default.

### DIFF
--- a/playgrounds/vite-react/src/App.tsx
+++ b/playgrounds/vite-react/src/App.tsx
@@ -70,6 +70,8 @@ function Account() {
         <br />
         chainId: {account.chainId}
         <br />
+        connector id: {account.connector?.uid}
+        <br />
         status: {account.status}
       </div>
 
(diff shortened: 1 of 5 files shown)


playgrounds/vite-react/src/App.tsx
```diff
@@ -158,8 +160,9 @@ function SignMessage() {
       <h2>Sign Message</h2>
 
       <form
-        onSubmit={(event) => {
+        onSubmit={async (event) => {
           event.preventDefault()
+          await new Promise((resolve) => setTimeout(resolve, 1_000))
           const formData = new FormData(event.target as HTMLFormElement)
           signMessage({ message: formData.get('message') as string })
```
tmm, line 167: `signMessage` does not redirect over to MM iOS app because 1s delay above cause universal link to expire.

PR #4023, opened 2024-06-09
feat: chain check

jxom:
Adds a chain check to `getConnectorClient` since it's possible for connection state chain ID to get out of sync with the connector (e.g. [wallet bug](https://github.com/MetaMask/metamask-extension/issues/25097)).


tmm (commented):
Wonder if we should push this check into `getConnectorClient` instead since all the higher-level mutation features (e.g. `sendTransaction`) use that? Also protects against someone dropping down to Viem via `sendTransaction(await getConnectorClient(), { ... })`.

PR #4033, opened 2024-06-11
chore: add `loadDefault` function to manage dynamic imports

zouhangwithsweet:
- Added `loadDefault` function to manage dynamic imports
- Ensured type-safe handling for modules with and without default exports


tmm:
Wasn't able to push to your branch so created https://github.com/wevm/wagmi/pull/4036 (your commits were added so you still get credit)

PR #4040, opened 2024-06-12
fix: vite import utility

HuberTRoy:
Hi~, When I trying new version of wagmi@2.9.12 & wagmi@2.10.0, the `CoinbaseConnector` & `MetamaskConnector` not working for me.

I found it's because this PR #4036 introduce a wrong utility.

The dynamic import from vite will be 
```
{ 
 default: {
   default: ...
 }
}
```

In the previous code:

`const { default: CoinbaseSDK_ } = await import('@coinbase/wallet-sdk')`, you get the first layer `default` first and then get the second layer `default`: 
```
if (typeof CoinbaseSDK_ !== 'function' && typeof CoinbaseSDK_.default === 'function') return CoinbaseSDK_.default
```
It works well. 

and When you change to utility, the judgement `if (typeof module !== 'function' && typeof module.default === 'function')` is not work because the first layer of `default` is object not a function.

image
image
image

### DIFF
--- a/packages/connectors/src/coinbaseWallet.ts
+++ b/packages/connectors/src/coinbaseWallet.ts
@@ -404,7 +404,7 @@ function version3(parameters: Version3Parameters) {
     },
     async getProvider() {
       if (!walletProvider) {
-        const SDK = await loadDefault<
+        const { default : SDK } = await loadDefault<
           Awaited<typeof import('cbw-sdk')>['default']
         >(import('cbw-sdk'))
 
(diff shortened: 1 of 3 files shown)


tmm:
Appreciate this! Ended up reverting the change entirely in https://github.com/wevm/wagmi/commit/ff083b4fc9941c1878e5415a2696af9727081cae. Will work with MetaMask and Coinbase to get named exports into their SDKs instead.

PR #4043, opened 2024-06-12
feat: Add connector icons for coinbase, metamask and walletconnect

jtaaa:
Adds icons to connectors exported by `wagmi/connectors`. Connector icons can be used as follows
```
const { connectors } = useConnectors()
return connectors.map(c => (
  <div>
    {c.icon && <img src={c.icon} />}
    {\* other content *\}
  </div>
))
```
Injected wallet connectors that are discovered via 6963 have icons set but currently when using the `coinbaseWallet` connector from `wagmi/connectors`, the connector object returned by `useConnectors` does not include an icon. It is very convenient to have the icon set here and there is not an easy way for the client to set/override the icon.

Please let me know if this belongs here or if we should make it easier for the user to set the icon and also if I should add icons to the playground connect buttons + if tests are required. I got the MM icon from the injected connector and the coinbase icon from their popup. The wallet connect one I got an svg from their website and converted it to base64 can definitely swap out if they would like different assets used! Thanks.

Also I wasn't sure what icon asset to use for the `safe` connector.


tmm:
Appreciate this, but at the moment we don't want to be the source of truth for this info.

jtaaa:
@tmm would it make sense to make it easier for the icon property on the connector to be set by users? currently I am doing this 😅
```
wagmiCreateConfig({
  // ...
  connectors: [
    // ...
    (...args) => {
      const connector = coinbaseWallet({
        appName: "OpenSea",
      })(...args)
      // @ts-expect-error Monkey patching the icon into the connector
      connector.icon = "data:image/svg+xml;base64,..."
      return connector
    },
  ],
})
```

not sure if there is a better way I am missing but there is no option to set the icon on the connector imported from `wagmi/connectors`. I could wrap `useConnectors` and include it there or map against the connector id in the UI but it would be convenient to be able to rely on the icon directly from the connector. Curious what you think. Maybe I just need to wrap it and add it myself :)

tmm:
Since icons are a view-level concern, I wouldn't modify the connector unless you need to.

```ts
const connectorIdToIconMap = {
  coinbaseWalletSDK: 'data:image/svg+xml;base64,...',
  metaMaskSDK: 'data:image/svg+xml;base64,...',
  ...
}
const { connectors } = useConnectors()
return connectors.map(c => (
  <div>
    {c.icon || connectorIdToIconMap[c.id] && <img src={c.icon ?? connectorIdToIconMap[c.id]} />}
    {\* other content *\}
  </div>
))
```

jtaaa:
Thanks for the reply. Makes sense. I do still think it would just be convenient to have it on the connector though. I guess in a perfect world Coinbase would maintain a connector and include their icon

PR #4053, opened 2024-06-13
feat: `useWatchAsset` hook

dalechyn:


packages/core/src/actions/watchAsset.ts
```diff
@@ -0,0 +1,44 @@
...
+  | viem_WatchAssetErrorType
+
+/** https://wagmi.sh/core/api/actions/watchAsset */
+export async function watchAsset(
+  config: Config,
+  parameters: WatchAssetParameters,
+): Promise<WatchAssetReturnType> {
+  const { connector, ...rest } = parameters
+
+  const client = await getConnectorClient(config, { connector })
```
dalechyn, line 40: @tmm, I don't think `account` nor `chainId` are needed here or lmk if my assumption is wrong.

PR #4059, opened 2024-06-18
chore: remove useless code

zouhangwithsweet:
I notice that just donot use default export, it works fine.


tmm:
Wasn't able to edit so closing in favor of https://github.com/wevm/wagmi/pull/4096

PR #4124, opened 2024-07-09
FIX: useConnectorClient to be enabled only when connected or reconnecting

t0rbik:
This PR aims to prevent `useConnectorClient` running when account is **not** in `connected` state.

Took this along as a quick fix, feel free to close if this is a desired behaviour. Cheers.

### DIFF
--- a/.changeset/short-eyes-hide.md
+++ b/.changeset/short-eyes-hide.md
@@ -1,5 +1,0 @@
----
-"wagmi": patch
----
-
-useConnectorClient only enabled when account status is 'connected'.
--- a/.changeset/short-eyes-hide.md
+++ b/.changeset/short-eyes-hide.md
@@ -1,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+useConnectorClient only enabled when account status is 'connected'.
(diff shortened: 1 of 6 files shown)


packages/react/src/hooks/useConnectorClient.ts
```diff
@@ -81,7 +81,7 @@ export function useConnectorClient<
     chainId: parameters.chainId ?? chainId,
     connector: parameters.connector ?? connector,
   })
-  const enabled = Boolean(status !== 'disconnected' && (query.enabled ?? true))
+  const enabled = Boolean(status === 'connected' && (query.enabled ?? true))
```
tmm, line 84: Agree with filtering out `'connecting'`. Let's add `'reconnecting'` here though as it enables optimistic updates.

.changeset/short-eyes-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+useConnectorClient only enabled when account status is 'connected'.
```
tmm, line 5:
```suggestion
Updated `useConnectorClient` to enable when status is `'reconnecting'` or `'connected'` (previously was also enabled when status was `'connecting'`).
```

tmm (changes requested):
Thanks! Let's add the same change (and changeset) for the [`useConnectorClient` Vue Composable](https://github.com/wevm/wagmi/blob/main/packages/vue/src/composables/useConnectorClient.ts#L100) too.

t0rbik:
Thank you, I've applied the suggestions.

PR #4128, opened 2024-07-11
feat: `useWatchAsset` hook

dalechyn:


dalechyn:
@tmm, I think I've fixed all of the test cases regarding the `useWatchAsset` hook, but the last `sendTransaction` test still fails.

Is it expected? https://github.com/wevm/wagmi/actions/runs/9898919460/job/27346778416?pr=4128

PR #4146, opened 2024-07-16
chore: update `@safe-global/safe-apps-sdk`

jxom:

### DIFF
--- a/packages/connectors/src/safe.ts
+++ b/packages/connectors/src/safe.ts
@@ -89,16 +89,7 @@ export function safe(parameters: SafeParameters = {}) {
       if (!isIframe) return
 
       if (!provider_) {
-        const { default: SafeAppsSDK } = await import(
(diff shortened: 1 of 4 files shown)


packages/connectors/src/safe.ts
```diff
@@ -89,16 +89,7 @@ export function safe(parameters: SafeParameters = {}) {
...
-          '@safe-global/safe-apps-sdk'
-        )
-        let SDK: typeof SafeAppsSDK.default
-        if (
-          typeof SafeAppsSDK !== 'function' &&
-          typeof SafeAppsSDK.default === 'function'
-        )
-          SDK = SafeAppsSDK.default
-        else SDK = SafeAppsSDK as unknown as typeof SafeAppsSDK.default
+        const { default: SDK } = await import('@safe-global/safe-apps-sdk')
```
jxom, line 92: Workaround no longer applies.

PR #4185, opened 2024-08-01
chore: update 'useDeeplink' default state to 'false' in the Metamask-SDK connector

omridan159:
This PR updates the default state of the `useDeeplink` parameter in the MetaMask SDK connector.

Previously:
`useDeeplink: parameters.useDeeplink ?? true`

Now changed to:
`useDeeplink: parameters.useDeeplink ?? false`

This change ensures that the `useDeeplink` feature is disabled by default unless explicitly set in the parameters.
It improves the user experience on mobile iOS devices.


OKEAMAH (approved):
Update

tmm:
If this isn't actively being worked on or ready for review (still in draft state), let's close it out for now.

PR #4186, opened 2024-08-01
feat: improve UX by using connectWith for MetaMask network sync

omridan159:
This PR improves the user experience (UX) when connecting to the MetaMask wallet by using the `connectWith` method instead of the `connect` method. This change ensures proper network configuration when the user changes the network on the dApp side and it is not in sync with the MetaMask wallet.

Key change:
- Replaced `sdk.connect()` with `sdk.connectWith()` to handle network switching or adding based on the provided `chainId`.

This enhancement provides a smoother and more seamless network switching experience for users on the dApp.

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -28,6 +28,7 @@ import {
   withRetry,
   withTimeout,
 } from 'viem'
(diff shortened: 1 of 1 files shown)


tmm:
If this isn't actively being worked on, let's close it out for now.

PR #4188, opened 2024-08-02
fix: set cookie path correctly

### LINKED ISSUE #4097: Disconnected session cookie not removed after connecting again
### Check existing issues

### Describe the bug

When calling `disconnect`, wagmi would normally store a session cookie like `wagmi.io.metamask.disconnected` so the page stays disconnected after a refresh.

However, this cookie should be cleared/removed if the user connects again. This removal of the "disconnected" cookie currently only works when the the "connect" call is made on the root `/` and first level pages (e.g. `/foo`) but **NOT** on pages that are deeper (e.g. `/bar/qux` or `/a/b/c`).

https://wagmi-bug.vercel.app/

### Steps To Reproduce

(Code is available at https://github.com/dengpeng/wagmi-bug)

- Correct behavior:
   1. Go to https://wagmi-bug.vercel.app/ or https://wagmi-bug.vercel.app/foo.
   2. Connect wallet then disconnect wallet
   3. Now a session cookie `wagmi.io.metamask.disconnected` is set.
   4. After a successful connect on that page once again, the cookie `wagmi.io.metamask.disconnected` is removed. ✅
   
- Incorrect behavior:
   1. Go to https://wagmi-bug.vercel.app/bar/qux or https://wagmi-bug.vercel.app/a/b/c
   2. Connect wallet then disconnect wallet
   3. Now a session cookie `wagmi.io.metamask.disconnected` is set.
   4. After a successful connect on that page once again, the cookie `wagmi.io.metamask.disconnected` is NOT removed. ❌
   6. This will lead to wallet not automatically re-connecting after the page is reloaded. ⚠️

wagmi, @wagmi/core

### Wagmi Version

@wagmi/core@2.11.5, wagmi@2.10.8

### Viem Version

2.16.3

### TypeScript Version

5

### Anything else?

_No response_

goosewobbler:
The path of the shimDisconnect cookie is being defaulted in all cases to `/`, this means that when wagmi is re-connected on the page `/a/b/c`, the cookie is not removed.  By setting the cookie path to `/a/b` upon disconnect,  the cookie is recognised when the app is re-connected.

Fixes #4097

### DIFF
--- a/packages/core/src/utils/cookie.ts
+++ b/packages/core/src/utils/cookie.ts
@@ -10,7 +10,9 @@ export const cookieStorage = {
   },
   setItem(key, value) {
     if (typeof window === 'undefined') return
-    document.cookie = `${key}=${value};Path=/;SameSite=Lax`
+    const pathBits = window.location.pathname.split('/')
+    const path = pathBits.slice(0, pathBits.length - 1).concat('/')
+    document.cookie = `${key}=${value};Path=${path};SameSite=Lax`
   },
   removeItem(key) {
     if (typeof window === 'undefined') return


tmm:
Thanks for the PR! This doesn't quite fix the issue as `path=/` is what we want so the cookie works on all paths under the domain. Instead `removeItem` needed [`path` added to it](https://github.com/wevm/wagmi/commit/1b9b523fa9b9dfe839aecdf4b40caa9547d7e594#diff-970fbf8a70495fb5f5795e547cef0483433db67e6302e598f5a85daa3ad8b39eR17).

goosewobbler:
Makes sense, I was close but no cigar

tmm:
Next time my friend!

PR #4200, opened 2024-08-10
Fixed wrongly passing 'abi' to 'useWriteContract'

Vedant-asati:
Fixed wrongly passing of parameter 'abi' to 'useWriteContract' hook.


Vedant-asati:
This problem is across all the examples, and the checks might be failing due to wagmi v1 being used for tests or any tests' specific issue. But for my case it wasnt working.
@tmm Kindly review. this

tmm:
There's nothing wrong here. `abi` is a [shorthand property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer).

PR #4202, opened 2024-08-13
Fix typo Update contributing

Ocheretovich:
Fix typo Update contributing


tmm:
This is actually correct. npm is the registry pnpm publishes to.

PR #4207, opened 2024-08-17
Fix race condition in injected.ts

Smert:
It can hangs if "once" will catch an event with unexpected chainId.

UPD: But need to correctly unsubscribe handler( Maybe you can suggest better fix.


tmm:
What's the issue? Mind dropping a minimal reproduction.

Smert:
Its heisenbug. `switchChain` can hang. I don't know why, but `config.emitter` can emit `change` event with another `chainId` and the promise will not be reloved

PR #4227, opened 2024-08-27
Fix the switchChain error encountered when using the "metaMask" connector

### LINKED ISSUE #4223: switchChainAsync throw an error when user approved
### Check existing issues

### What I am doing:

1. Connect MetaMask wallet using the "metaMask" connector.
2. Refresh page
3. Switch to a network (e.g., BSC) that is not yet added to your MetaMask wallet.
4. A popup to switch to BSC appears, click “Approve” and then “Switch Network”.

### What I expected:

The network is added and switched without any errors.

### What is actually happening:

switchChainAsync throws an error: User rejected the request.
Details: The user rejected the switch after the network was added.
Version: 2.20.0

https://codesandbox.io/p/sandbox/wagmi-switchchain-6m86q5

### Steps To Reproduce

Here is the link: https://codesandbox.io/p/sandbox/wagmi-switchchain-6m86q5

Steps to reproduce the issue:

1. Remove BNB Smart Chain from your MetaMask wallet.
2. Click the Connect button.
3. Refresh the page.
4. Click the BNB Smart Chain button.
5. A popup to switch to BSC will appear. Click Approve, then Switch Network.
6. An error message will be displayed below the button.

wagmi

2.12.7

### Viem Version

2.20.0

### TypeScript Version

5.5.4

### Anything else?

_No response_

xianchenxc:
fix #4223


packages/connectors/src/metaMask.ts
```diff
@@ -283,11 +283,17 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
-            await provider.request({
-              method: 'wallet_addEthereumChain',
-              params: [addEthereumChain],
-            })
-
+            await Promise.all([
+              provider.request({
+                method: 'wallet_addEthereumChain',
+                params: [addEthereumChain],
+              }),
```
DNR500, line 290: Should this also have the additional have the additional then block to fire the change event if its needed?

https://github.com/wevm/wagmi/blob/ceadbbe99b987149aeccc9411897431e4c7f0ab2/packages/connectors/src/metaMask.ts#L230-L239

tmm:
Calling `'eth_chainId'` directly solves it. It seems `provider.getChainId` does not always update `this.state.chainId` immediately.

PR #4229, opened 2024-08-27
fix: prevent component re-rendering with the same account snapshot

weilliao05621:
In addressing performance issues with a specific React hook, I identified that components using this hook were re-rendering unnecessarily. This was caused by the hook destructuring an object, which created a new reference each time it executed.

```ts
// `trackedResult` will always has new reference when the hook executed
 const trackedResult = { ...result }
```

To optimize this, I use `useMemo` to wrap the returned object. This change prevents the creation of a new object reference on every execution, ensuring that the object is only recalculated when its dependencies change. As a result, the number of re-renders for components utilizing this hook has significantly decreased.

## Screenshots

Have tried the updated version in the project's react playground.

### Before

https://github.com/user-attachments/assets/68a5190f-8843-4cb4-a807-8f9814544ad4

### After

https://github.com/user-attachments/assets/847b875a-66d3-4f09-b551-08c841253ba6


0x33dm:
Nice catch! i hope this gets merged soon.

tmm:
Can you post the playground code that demonstrates this behavior or [add a test](https://github.com/wevm/wagmi/blob/main/packages/react/src/hooks/useSyncExternalStoreWithTracked.test.ts) that reproduces?

weilliao05621:
sure! will be working on the test case

weilliao05621:
screenshot for passing unit test

截圖 2024-09-27 下午2 48 30

--- 

I also do manual test with the playground:) 
Here's the playground code for testing:
```ts
const MemoAccount = memo(function (props: {
  store: ReturnType<typeof useAccount>
}) {
  console.log("account update!")
  return <div>{props.store.address}</div>
})

function RerenderWithAccount() {
  const account = useAccount()
  const [count, setCount] = useState<number>(0)

  return (
    <>
      <MemoAccount store={account} />
      <button
        onClick={() => {
          setCount((prev) => prev + 1)
        }}
      >
        Update count: {count}
      </button>
    </>
  )
}
```

PR #4285, opened 2024-09-20
Add slicedAddress to getAccount action

LukaMartin:
- Added slicedAddress return value to getAccount action.
- Added sliceAddress function and tests.

I have created this PR to add a slicedAddress return value to the getAccount action and useAccount hook.
The returned value is in the following format: 0x12...fe12
I feel this is a very common use case for dapps that are using wagmi/viem and have wanted this myself on past projects


tmm:
Appreciate the PR, but think it's best for folks to add this to their project directly.

PR #4311, opened 2024-10-02
fix: injected connector throws error after switching to a chain that was just added via `'wallet_addEthereumChain'`.

chybisov:
Fixes the same issue encountered in #4227 and #4223 where the connector throws an error after switching to a chain that was just added via `'wallet_addEthereumChain'` but for the injected connector.

### DIFF
--- a/.changeset/unlucky-boats-dance.md
+++ b/.changeset/unlucky-boats-dance.md
@@ -2,4 +2,4 @@
 "@wagmi/core": patch
 ---
 
-Fixed injected connector throwing error after switching to a chain that was just added via 'wallet_addEthereumChain'
+Fixed `injected` connector race condition after calling `'wallet_addEthereumChain'` in `switchChain`.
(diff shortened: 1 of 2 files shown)


tmm:
Not able to reproduce this issue anymore. If you can supply a reproduction, happy to work on merging this.

chybisov:
@tmm hey, the issue is still there. 

https://codesandbox.io/p/sandbox/wagmi-switchchain-forked-47mx8q

Steps
1. Make sure BSC is not added to MetaMask
2. Switch to Ethereum in MetaMask
3. Connect and try to switch to BSC

Code that fails

```ts
const switchedChain = await switchChain(config, {
  chainId: chain.id,
});
const client = await getConnectorClient(config, {
  chainId: switchedChain.id,
});
```

Error

```
The current chain of the connector (id: 56) does not match the connection's chain (id: 1). Current Chain ID: 56 Expected Chain ID: 1 Version: @wagmi/core@2.14.3
```

tmm:
@chybisov thanks! made a small update.

PR #4313, opened 2024-10-03
feat: addChain

marthendalnunes:
Adds `addChain` action and `useAddChain` hook.


tmm:
Working on the next major version of Wagmi and going to make it easier to add actions and framework-specific function with much less code. Going to close for now and will add your work into there so you still get credit.

PR #4318, opened 2024-10-05
docs: fix description of 'value' in usePrepareTransactionRequest

jaxxjj:
Fix an issue in the documentation for the `usePrepareTransactionRequest` hook. Specifically, the description of the `value` parameter was incorrect. It previously stated:

> "The transaction recipient or contract address."

The correct description should be:

> "The amount of ether (in wei) to send with the transaction."

### DIFF
--- a/site/react/api/hooks/usePrepareTransactionRequest.md
+++ b/site/react/api/hooks/usePrepareTransactionRequest.md
@@ -284,7 +284,7 @@ function App() {
 
 `bigint | undefined`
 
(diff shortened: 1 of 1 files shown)


jaxxjj:
**Note:** I've noticed that the description for the `value` parameter in `prepareTransactionRequest.md` also contains the same incorrect wording.

The current description is: 
> "The transaction recipient or contract address."

It should be: 
> "The amount of ether (in wei) to send with the transaction."

PR #4323, opened 2024-10-08
feat: update metamask sdk and implement connectSign

abretonc7s:
Update metamask sdk to 0.29.1

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -64,7 +64,6 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
   let sdk: MetaMaskSDK
   let provider: Provider | undefined
   let providerPromise: Promise<typeof provider>
-  let cachedDisplayUri: string | undefined
 
   let accountsChanged: Connector['onAccountsChanged'] | undefined
   let chainChanged: Connector['onChainChanged'] | undefined
@@ -97,10 +96,6 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         let signResponse: string | undefined
         let connectWithResponse: unknown | undefined
         if (!accounts?.length) {
-          if (cachedDisplayUri) {
-            provider.emit('display_uri', cachedDisplayUri)
-          }
-
           let requestedAccounts: `0x${string}`[]
           if (parameters.connectAndSign) {
             signResponse = await sdk.connectAndSign({
@@ -430,9 +425,6 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
 
       config.emitter.emit('disconnect')
 
-      // Remove cached display URI
-      cachedDisplayUri = undefined
-
       // Manage EIP-1193 event listeners
       if (!accountsChanged) {
         accountsChanged = this.onAccountsChanged.bind(this)
@@ -452,13 +444,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       }
     },
     onDisplayUri(uri) {
-      // Cache display uri
-      cachedDisplayUri = uri
-
-      config.emitter.emit('message', {
-        type: 'display_uri',
-        data: uri,
-      })
+      config.emitter.emit('message', { type: 'display_uri', data: uri })
     },
   }))
 }
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -96,7 +96,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         let signResponse: string | undefined
         let connectWithResponse: unknown | undefined
         if (!accounts?.length) {
-          let requestedAccounts: `0x${string}`[]
+          let requestedAccounts: Address[]
           if (parameters.connectAndSign) {
             signResponse = await sdk.connectAndSign({
               msg: parameters.connectAndSign,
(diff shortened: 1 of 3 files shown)


packages/connectors/src/metaMask.ts
```diff
@@ -186,8 +239,11 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
               return [chain.id, url]
             }),
           ),
-          dappMetadata: parameters.dappMetadata ?? { name: 'wagmi' },
-          useDeeplink: parameters.useDeeplink ?? true,
+          dappMetadata,
```
EdouardBougon, line 242: Why do we overwrite the dappMetaData provided by in parameters ?

packages/connectors/src/metaMask.ts
```diff
@@ -76,11 +93,26 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       if (isReconnecting) accounts = await this.getAccounts().catch(() => [])
 
       try {
+        let signResponse: string | undefined
+        let connectWithResponse: unknown | undefined
         if (!accounts?.length) {
-          const requestedAccounts = (await sdk.connect()) as string[]
-          accounts = requestedAccounts.map((x) => getAddress(x))
+          let requestedAccounts: `0x${string}`[]
```
EdouardBougon, line 99:
```typescript
import { type Address } from 'viem';

let requestedAccounts: Address[];
```

EdouardBougon (commented):
Do you you know if other wallets have the same features of `connectAndSign` and `connectWith`?

If the app only uses the MetaMask connector, it works.
But if they use multiple connectors, this means that in the Dapp code, the connector will need to be tested:
- If it’s MetaMask, then just call `connect`.
- If it’s not MetaMask, you need to call `connect` and then `signMessage`.

--

The best solution would be to propose a new `connectAndSign` hook to Wagmi, where the above logic would be wrapped. We could also add a new optional `signMessage` parameter to the connect method of the connector: https://wagmi.sh/dev/creating-connectors#methods

I'm available to discuss about it

abretonc7s:
> Do you you know if other wallets have the same features of `connectAndSign` and `connectWith`?
> 
> If the app only uses the MetaMask connector, it works. But if they use multiple connectors, this means that in the Dapp code, the connector will need to be tested:
> 
> * If it’s MetaMask, then just call `connect`.
> * If it’s not MetaMask, you need to call `connect` and then `signMessage`.
> 
> --
> 
> The best solution would be to propose a new `connectAndSign` hook to Wagmi, where the above logic would be wrapped. We could also add a new optional `signMessage` parameter to the connect method of the connector: https://wagmi.sh/dev/creating-connectors#methods
> 
> I'm available to discuss about it

The idea here is to have a least intrusive solution and allow the dapps using metamask to have a  better U/X (no need for two deeplink for the mobile sign request)

packages/connectors/src/metaMask.ts
```diff
@@ -31,10 +31,27 @@ import {
...
+      | 'forceDeleteProvider'
+    >
+  >
 >
 
+export type MetaMaskParameters = WagmiMetaMaskSDKOptions & {
+  // Shortcut to connect and sign a message
+  connectAndSign?: string
+  // Allow connectWith any rpc method
+  connectWith?: { method: string; params: unknown[] }
```
tmm, line 52:
```suggestion
  connectWith?: { method: string; params: unknown[] } | undefined
```

What `method`s are supported? We should be able to strongly type this with the utility types in Viem (e.g. `EIP1474Methods`) if we want.

packages/connectors/src/metaMask.ts
```diff
@@ -168,8 +214,11 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
         // See: https://github.com/vitejs/vite/issues/9703
         const MetaMaskSDK = await (async () => {
           const { default: SDK } = await import('@metamask/sdk')
+          // @ts-ignore
           if (typeof SDK !== 'function' && typeof SDK.default === 'function')
+            // @ts-ignore
             return SDK.default
+          // @ts-ignore
```
tmm, line 221: I'll take a crack on getting TS to play nice.

packages/connectors/src/metaMask.ts
```diff
@@ -186,8 +235,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
               return [chain.id, url]
             }),
           ),
-          dappMetadata: parameters.dappMetadata ?? { name: 'wagmi' },
-          useDeeplink: parameters.useDeeplink ?? true,
+          dappMetadata: parameters.dappMetadata ?? {
+            url: `${window.location.protocol}//${window.location.host}`,
```
tmm, line 239: We should check that `window` is available (`typeof window !== 'undefined'`)

packages/connectors/src/metaMask.ts
```diff
@@ -186,8 +235,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
+          },
+          useDeeplink: true,
+          injectProvider: false,
+          forceInjectProvider: false,
+          forceDeleteProvider: false,
```
tmm, line 244: What were the defaults internally before? Just want to make sure changes these isn't too disruptive for folks. Probably should allow folks to override these manually so it's not a breaking change and remove that ability in the next major version. (Very unlikely anyone will override manually since people just use defaults, but want to make sure semver is respected.)

tmm: Is this PR dependent on that one? Haven't had a chance to wrap up our thinking on that quite yet.

tmm:
Closing in favor of https://github.com/wevm/wagmi/pull/4337

PR #4334, opened 2024-10-11
fix: returning walletClient and reconnecting mock client

### LINKED ISSUE #4317: `mock` Connector Results in `undefined` result from `useWalletClient`
### Check existing issues

### Describe the bug

I'm currently experimenting with `mock` connector to establish a "fake" connection within our app. I've found that:
- `useWalletClient` result is undefined, but I expect some mock signer to be returned instead
- `reconnect: true` does not result in a reconnection after refreshing the page

https://stackblitz.com/edit/new-wagmi-f3azfv?file=src%2FApp.tsx

### Steps To Reproduce

See the stackblitz

wagmi

2.10.10

### Viem Version

2.17.4

### TypeScript Version

_No response_

### Anything else?

_No response_

marthendalnunes:
Fixed #4317 by adding a catch on `getAccounts` method in the mock client and retrieving correctly a walletClient, and also fixed the reconnection behaviour of the mock client by using `config.storage`

### DIFF
--- a/.changeset/strange-hounds-rest.md
+++ b/.changeset/strange-hounds-rest.md
@@ -2,4 +2,4 @@
 "@wagmi/core": patch
 ---
 
-Fixed the getAccounts method for the mock client.
+Fixed the getAccounts method and reconnecting logic for the mock client.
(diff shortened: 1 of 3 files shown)


tmm:
See https://github.com/wevm/wagmi/issues/4317#issuecomment-2418105760

PR #4339, opened 2024-10-14
fix: waitForTransactionReceipt proper eth_call usage

AndriyAntonenko:
This PR addresses a bug in the @wagmi/core library, specifically in the waitForTransactionReceipt function. The bug occurs when arguments are provided to skip simulation and gas estimation. The data field was not being passed to the eth_call function, which resulted in unexpected errors.

I've made a simple repository that has a script that reproduces this issue: https://github.com/AndriyAntonenko/wagmi-waitfortransaction-bug-poc

### DIFF
--- a/packages/core/src/actions/waitForTransactionReceipt.ts
+++ b/packages/core/src/actions/waitForTransactionReceipt.ts
@@ -71,7 +71,9 @@ export async function waitForTransactionReceipt<
       maxFeePerGas: txn.type === 'eip1559' ? txn.maxFeePerGas : undefined,
       maxPriorityFeePerGas:
         txn.type === 'eip1559' ? txn.maxPriorityFeePerGas : undefined,
+      data: txn.input ? txn.input : undefined,
     })
+
     const reason = code?.data
       ? hexToString(`0x${code.data.substring(138)}`)
       : 'unknown reason'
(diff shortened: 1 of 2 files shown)


packages/core/src/actions/waitForTransactionReceipt.ts
```diff
@@ -65,17 +76,34 @@ export async function waitForTransactionReceipt<
...
+      if (
+        contract &&
+        error instanceof CallExecutionError &&
+        error.cause instanceof ExecutionRevertedError &&
+        error.cause.cause instanceof RpcRequestError
+      ) {
+        throw new ContractFunctionRevertedError({
+          abi: contract.abi as Abi,
+          functionName: contract.functionName,
+          data: (error.cause.cause.cause as any).data,
```
dalechyn, line 103: can typing be slightly improved here?

tmm:
Let's update this PR so it only fixes the issue and doesn't add new features.

> Also, I've added the possibility to pass abi and functionName as optional params in waitForTransactionReceipt. These arguments will be used to parse the possible error thrown by action_call function

Ideally this type of thing is added to Viem so Wagmi gets it for free. Please [create a discussion](https://github.com/wevm/viem/discussions) over there before creating a PR.

packages/core/src/actions/waitForTransactionReceipt.ts
AndriyAntonenko: Removed this code as suggested in commend below

AndriyAntonenko:
@tmm thanks, I've removed the part with error handling. Now this PR only fix the bug

PR #4343, opened 2024-10-16
feat(connectors): filter mipd by connector rdns

tmm:
Allows connectors to specify `rdns` property and filters out matching EIP 6963 connectors from being created.

Related https://github.com/wevm/wagmi/pull/4328

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -263,7 +263,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
             parameters.dappMetadata ??
             (typeof window !== 'undefined'
               ? { url: window.location.origin }
-              : { name: 'wagmi' }),
+              : { name: 'wagmi', url: 'https://wagmi.sh' }),
           useDeeplink: parameters.useDeeplink ?? true,
         })
         await sdk.init()
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -97,9 +97,18 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
     type: metaMask.type,
     async setup() {
       const provider = await this.getProvider()
-      if (provider && !connect) {
-        connect = this.onConnect.bind(this)
-        provider.on('connect', connect as Listener)
+      if (provider?.on) {
+        if (!connect) {
+          connect = this.onConnect.bind(this)
+          provider.on('connect', connect as Listener)
+        }
+
+        // We shouldn't need to listen for `'accountsChanged'` here since the `'connect'` event should suffice (and wallet shouldn't be connected yet).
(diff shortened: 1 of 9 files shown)


tmm:
Open thoughts:
- Could imagine someone setting an invalid RDNS to filter out another connector. We can make sure this doesn't happen for built-in connectors, but someone could easily do downstream (e.g. Connector A sets rdns to `'com.connector.b'` so their connector overrides Connector B).
- Maybe there is an approach that inverts control to make this easier for folks, where we don't need to change core code?
- Want to avoid hardcoding values (e.g. `'com.example.wallet'`) if possible. It's annoying and error prone.

tmm:
Let's roll this out and see how it goes.

abretonc7s:
> Open thoughts:
> 
> * Could imagine someone setting an invalid RDNS to filter out another connector. We can make sure this doesn't happen for built-in connectors, but someone could easily do downstream (e.g. Connector A sets rdns to `'com.connector.b'` so their connector overrides Connector B).
> * Maybe there is an approach that inverts control to make this easier for folks, where we don't need to change core code? Lots of devs use higher-level "Connect Wallet" libraries so maybe less of an issue for us to solve.
> * Want to avoid hardcoding values (e.g. `'com.example.wallet'`) if possible. It's annoying and error prone.

I totally get the concern about setting an invalid rdns. To reduce that risk, we can limit the use of rdns to default built-in connectors... A simple validation check could ensure that only valid or registered rdns values are used, which would prevent any issues with third-party connectors while keeping things stable for the built-ins.

As for hardcoded values, the rdns for well-known connectors like MetaMask and Coinbase is stable and specific to those connectors, so it’s not really arbitrary or prone to errors. For third-party connectors, developers could handle these identifiers however they want. Wagmi could also expose the mechanism used for the default connectors so developers have the option to extend it if needed, without messing with the core code.

BjornGunnarsson:
hey @tmm, thanks for drafting this up.

To clarify, we are specifically addressing scenarios where developers are integrating the MetaMask SDK[it's opt-in], aiming to offer a more feature-rich experience, including mobile wallet support. 
Our goal is to resolve the pain point for users who wish to use the mobile wallet, which is not compatible with the injected provider.

The core package is focused on supporting the most widely used and popular connectors, including MetaMask. As such, hardcoding the reverse DNS (RDNS) for these specific connectors is a logical approach. This would not disrupt downstream libraries but would instead enhance the core functionality by not adding extra filtering logic there where you have already made a decision on the wagmi/connector level.

packages/connectors/src/metaMask.ts
```diff
@@ -257,7 +263,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
             parameters.dappMetadata ??
             (typeof window !== 'undefined'
               ? { url: window.location.origin }
-              : { name: 'wagmi' }),
+              : { name: 'wagmi', url: 'https://wagmi.sh' }),
```
tmm, line 266: error spamming SSR (e.g. Next.js) without a URL set

PR #4347, opened 2024-10-16
fix: MetaMask disconnecting during the connection process.

EdouardBougon:
In certain cases, the MetaMask SDK, when connected via mobile, may emit an `accountsChanged` event with an empty list, followed shortly by another event containing the correct account list.

In the current implementation of the MetaMask provider, receiving an empty list causes it to disconnect, leading to unintended side effects.

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -386,10 +386,10 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
       }
     },
     async onAccountsChanged(accounts) {
-      // Disconnect if there are no accounts
-      if (accounts.length === 0) this.onDisconnect()
+      // Do nothing if there are no accounts
+      if (accounts.length === 0) return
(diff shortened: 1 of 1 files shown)


packages/connectors/src/metaMask.ts
```diff
@@ -96,9 +96,18 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
+          connect = this.onConnect.bind(this)
+          provider.on('connect', connect as Listener)
+        }
+
+        // We shouldn't need to listen for `'accountsChanged'` here since the `'connect'` event should suffice (and wallet shouldn't be connected yet).
+        // Some wallets, like MetaMask, do not implement the `'connect'` event and overload `'accountsChanged'` instead.
+        if (!accountsChanged) {
+          accountsChanged = this.onAccountsChanged.bind(this)
+          provider.on('accountsChanged', accountsChanged as Listener)
+        }
```
tmm, line 110: Noticed that the MetaMask Connector does not work with connecting directly from extension.

PR #4362, opened 2024-10-23
fix: MetaMask readonlyRPCMap type

EdouardBougon:
The option readonlyRPCMap should have a chainId in hexString as a key, not a number.

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -1,6 +1,7 @@
 import type {
   MetaMaskSDK,
   MetaMaskSDKOptions,
+  RPC_URLS_MAP,
   SDKProvider,
 } from '@metamask/sdk'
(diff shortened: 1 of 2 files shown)


packages/connectors/src/metaMask.ts
```diff
@@ -243,30 +244,28 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
-          ),
+          readonlyRPCMap,
           dappMetadata:
             parameters.dappMetadata ??
             (typeof window !== 'undefined'
               ? { url: window.location.origin }
               : { name: 'wagmi', url: 'https://wagmi.sh' }),
           useDeeplink: parameters.useDeeplink ?? true,
         })
-        await sdk.init()
```
tmm, line 269: deprecated and not required

PR #4365, opened 2024-10-25
@wagmi/svelte

ByteAtATime:
This PR is an implementation of reactive Wagmi hooks in Svelte. It implements about half of the hooks, mainly building the foundation for future contributions.

### Reactivity

In order to make return types reactive, I have opted to return a **function**. It's the *function* that then returns the value. While it is a little unintuitive, it works perfectly with [`$derived.by`](https://svelte.dev/docs/svelte/$derived#$derived.by):

```ts
const { data, isLoading } = $derived.by(useEnsName(...));
```

**Design consideration**: should we use React-style `.current` instead of returning a function?

<details>
  <summary>Why?</summary>

Unfortunately, simply returning a primitive from a function doesn't make it reactive. The workaround is to return an object with a getter:

```ts
const reactiveReturn = () => {
  // ... do stuff

  return {
    get count() {
      return count;
    }
  }
}
```

However, I find this a little ugly when used, especially because it cannot be destructured:

```ts
const countObj = reactiveReturn();
countObj.count // repetitive?
```
</details>

---

Similarly, a similar problem occurs with parameters: if a reactive object is directly passed into a function, it won't be reactive:

```ts
let counter = $state(69);
const output = useCounter(counter);
                       // ^ State referenced in its own scope will never update. Did you mean to reference it inside a closure?
```

Happily, this isn't the only framework that has this problem (SolidJS shares a similar problem). In both frameworks, the workaround is to simply pass in a function as a parameter:

```ts
let reactiveAddress = $state("0x...");
$derived.by(useEnsName(() => ({ address: reactiveAddress })));
```

~~**Design consideration**: In the `RuneParameters` type, I have defined it as `(() => T) | T`. In other words, it accepts either a reactive function or a non-reactive object. **Will this lead to confusion?**~~
Edit: Now we can only pass in a function. This makes it easier to compose types *and* reduces confusion for users.

**Design consideration**: should the return types and parameters include the function or not?
- Including the function -> actual return types/parameters
- Excluding the function -> easier to be composed in types (currently, you would need to use `ReturnType<UseChainIdReturnType>` or `ReturnType<UseChainIdParameters>`)

### Testing

~~So far, I don't see a way of testing the Svelte package without adding the dependencies to the root. For now, I'm simply running `vitest` in the package instead of adding it to the Vitest workspace -- I will look into it in the future.~~
Edit: I'm still using a seperate `vite.config.ts`, but it's added into the workspace


packages/svelte/.npmrc
tmm, line 1: Let's move this to the root `.npmrc` if it's required.

packages/svelte/.gitignore
tmm, line 1: can we consolidate this file with the root `.gitignore`.

packages/svelte/README.md
tmm, line 1: Let's do a stripped down README similar to what we do for [other packages](https://github.com/wevm/wagmi/blob/main/packages/react/README.md).

packages/svelte/package.json
```diff
@@ -0,0 +1,63 @@
...
+      "types": "./dist/exports/chains.d.ts",
+      "default": "./dist/exports/chains.js"
+    }
+  },
+  "peerDependencies": {
+    "@tanstack/svelte-query": "https://pkg.pr.new/@tanstack/svelte-query@ccce0b8",
+    "viem": "2.x",
+    "svelte": "^5.0.0"
+  },
+  "devDependencies": {
```
tmm, line 37: Would like to avoid Prettier and ESLint if [Biome support](https://biomejs.dev/internals/language-support) is good enough.

packages/svelte/src/lib/hooks/test.svelte.ts
tmm, line 1: Let's move this to `@wagmi/test/svelte`.

tmm (commented):
Looking good! Some comments. Let's also move the playground to `playgrounds/sveltekit` so `packages/svelte/src/` follows the same directory structure as the other packages.

packages/svelte/.npmrc
ByteAtATime: Resolved in https://github.com/wevm/wagmi/pull/4365/commits/2c96b3858bf2283d0208b55cb336a0f7b48b057f

packages/svelte/.gitignore
ByteAtATime: Resolved in https://github.com/wevm/wagmi/pull/4365/commits/2c96b3858bf2283d0208b55cb336a0f7b48b057f

packages/svelte/package.json
ByteAtATime: Mainly resolved in https://github.com/wevm/wagmi/pull/4365/commits/2917d4254af926f51af667a0b8e801b9adaf0854 with minor changes in https://github.com/wevm/wagmi/pull/4365/commits/9d2c30adddd6fa66101ceb3985531066537cf71b

packages/svelte/src/lib/hooks/test.svelte.ts
ByteAtATime: Resolved in https://github.com/wevm/wagmi/pull/4365/commits/ffb16a13d92ebffbc8c75e8dbfd933d1cade8ca2

roninjin10:
Hey @ByteAtATime . Thank you for this pr. I'm gonna use it as reference for [krome](https://github.com/evmts/krome)

roninjin10:
@ByteAtATime @tmm are there any blockers to me helping push this pr over the finish line? I think it's easier and more useful for everybody to just help merge this pr than port it to a different project

packages/svelte/src/hooks/useAccount.svelte.ts
```diff
@@ -0,0 +1,40 @@
...
+import { useConfig } from './useConfig.svelte.js'
+
+export type UseAccountParameters<config extends Config = Config> =
+  RuneParameters<ConfigParameter<config>>
+
+export type UseAccountReturnType<config extends Config = Config> =
+  RuneReturnType<GetAccountReturnType<config>>
+
+/** https://wagmi.sh/react/api/hooks/useAccount */
+export function useAccount<config extends Config = ResolvedRegister['config']>(
```
roninjin10, line 22: Is useFoo normal for svelte apps? Would prefer it to just be `getAccount`

roninjin10: Noticing tanstack docs does this pattern of useFoo
image

Morinyo1907 (approved):

PR #4371, opened 2024-10-29
fix: Safe connector not working in some Vite apps

### LINKED ISSUE #4364: Safe connector not working, throwing a `SafeAppProvider` is not a constructor error
### Check existing issues

### Describe the bug

I'm trying to open my Vite dApp as a Safe app, it works well inside the Safe wallet context, but the Safe connector will not connect. After digging for a bit and doing some debugging I found out there is a `SafeAppProvider is not a constructor` error thrown by the connector.

I see there has been similar issues fixed for other connectors before, like [this one](https://github.com/wevm/wagmi/pull/3940), so perhaps the solution for this connector is similar

https://github.com/iceanddust/wagmi-safe-connector-issue

### Steps To Reproduce

- Install packages and run dev server
- Go to https://app.safe.global/ and select (or open) a safe
- Go to the `Apps` tab, and then to `My custom apps` tab
- Click `Add custom Safe App`, and enter the URL (Should be `http://localhost:5173/` if you're running locally)
- Check the checkbox and click `Add`, then open the Safe app
- Inside the app, click on the `Safe` connector and see the error pop up, the connector will never connect

wagmi

2.12.25

### Viem Version

2.21.34

### TypeScript Version

_No response_

### Anything else?

_No response_

iceanddust:
### Description
Fixes #4364 by using the same logic applied in the [MetaMask](https://github.com/wevm/wagmi/blob/main/packages/connectors/src/metaMask.ts#L240-L245) and [CoinbaseWallet](https://github.com/wevm/wagmi/blob/main/packages/connectors/src/coinbaseWallet.ts#L176-L181) connectors

### DIFF
--- a/.changeset/olive-chefs-learn.md
+++ b/.changeset/olive-chefs-learn.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/connectors": patch
+---
+
+Fixed Safe connector not working in some Vite apps
--- a/packages/connectors/src/safe.ts
+++ b/packages/connectors/src/safe.ts
@@ -101,13 +101,13 @@ export function safe(parameters: SafeParameters = {}) {
         // Unwrapping import for Vite compatibility.
         // See: https://github.com/vitejs/vite/issues/9703
         const SafeAppProvider = await (async () => {
-          const providerImport = await import('@safe-global/safe-apps-provider')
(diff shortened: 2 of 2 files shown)


1997roylee:
I don't suggest make a change for it, please reference this link: https://github.com/vitejs/vite/issues/17529

PR #4397, opened 2024-11-11
feat: Improve chain switching via MetaMask connector

christopherferreira9:
This PR aims to solve an issue where the original logic of switchChain -> add if it's not present would cause an extra back and forth between dapp and the MetaMask wallet, particularly on Mobile to Mobile interactions.

- Adds a retry to the eth_chainId call to avoid a race condition


tmm:
looks like this got include as part of https://github.com/wevm/wagmi/pull/4422

PR #4409, opened 2024-11-13
Use new init method + expose preference object

fan-zhang-sv:
In latest `@coinbase/wallet-sdk`, we enriched one of the config param, `preference`, to be able to take `attribution` for onchain analytics, `keysUrl` for testing against different smart wallet env. So updating `@wagmi/connectors` to expose this new preference object to end users.

* this preference object can take arbitrary param, so it provides more flexibility to end user as we push out more new params
* this PR also replaced the deprecated class `CoinbaseWalletSDK` with the newly-introduced `createCoinbaseWalletSDK` to better align w js norm

### Demo

https://github.com/user-attachments/assets/a8e0c81d-8f57-46de-b15a-8745b8c99d83

### DIFF
--- a/.changeset/bright-suns-rush.md
+++ b/.changeset/bright-suns-rush.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/connectors": minor
+---
+
+Added `preference` object for Coinbase Wallet connector.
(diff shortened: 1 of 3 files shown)


cb-jake (commented):
thanks Felix. Looks good. This likely warrants a major version bump on the connectors pkg

spencerstock (approved):
Looks sane and good to me

Svenne1980 (commented):
I dont understand it completely,  sorry

PR #4417, opened 2024-11-20
feat: remove simulation

jxom:
Modern wallets (and wallets in general) should be smart enough to perform simulation & estimate gas on their side. This PR removes the need to perform simulations & gas estimations from the send trasnaction & contract write actions.


packages/core/src/actions/codegen/createWriteContract.ts
```diff
@@ -86,7 +86,7 @@ export type CreateWriteContractReturnType<
               | undefined
           }
         : Compute<ChainIdParameter<config, chainId>>) &
-      ConnectorParameter & { __mode?: 'prepared' }
```
tmm, line 89: Might be good to leave `__mode` here and just add a `@deprecated` comment so people that upgrade don't experience unexpected type errors if they were passing this in directly (to skip simulation).

packages/core/src/actions/simulateContract.ts
tmm, line 1: Probably good to keep and deprecate `__mode` here too.

packages/core/src/connectors/injected.ts
tmm, line 1: I'd keep this around too

Mahmoudhpp (approved):

Magkoooh (approved):

packages/core/src/actions/sendTransaction.ts
```diff
@@ -78,37 +74,10 @@ export async function sendTransaction<
...
-
-  const gas = await (async () => {
-    // Skip gas estimation if `data` doesn't exist (not a contract interaction).
-    if (!('data' in parameters) || !parameters.data) return undefined
-
-    // Skip gas estimation if connector supports simulation.
-    if ((connector ?? activeConnector)?.supportsSimulation) return undefined
-
-    // Skip gas estimation if `null` is provided.
-    if (gas_ === null) return undefined
```
cgero-eth, line 91: @jxom @tmm This is breaking, `viem` throws error if gas is set to `null`:
```
Details: Cannot convert null to a BigInt
Version: viem@2.21.51
    at getTransactionError (getTransactionError.js:18:12)
    at sendTransaction (sendTransaction.js:219:105)
    at async sendTransaction (sendTransaction.js:24:18)Caused by: TypeError: Cannot convert null to a BigInt
    at BigInt (<anonymous>)
    at numberToHex (toHex.js:135:19)
    at formatTransactionRequest (transactionRequest.js:37:89)
    at sendTransaction (sendTransaction.js:114:29)
    at async sendTransaction (sendTransaction.js:24:18)
```
Documentation should also be updated:
Screenshot 2024-11-27 at 10 00 28

PR #4447, opened 2024-12-09
Fix: pass config parameter to useConnectorClient in useSimulateContract and useEstimateGas

Aerilym:
Added the config parameter to `useConnectorClient` parameters when used in `useSimulateContract` and `useEstimateGas`. `useConnectorClient` would always call `useConfig` as config wasn't provided in the parameters, this broke providing a config to either hook and would mean `useConfig` was triggered in both hooks. This caused an error when `WagmiProvider` wasn't available in the context, as `useConnectorClient` would call `useContext(WagmiContext)` even if a config parameter was provided to the parent hooks.

### DIFF
--- a/.changeset/proud-knives-behave.md
+++ b/.changeset/proud-knives-behave.md
@@ -1,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+Fixed config parameter passing in useSimulateContract and useEstimateGas
--- a/packages/react/src/hooks/useEstimateGas.ts
+++ b/packages/react/src/hooks/useEstimateGas.ts
@@ -51,6 +51,7 @@ export function useEstimateGas(
 
   const config = useConfig(parameters)
   const { data: connectorClient } = useConnectorClient({
+    config,
     connector,
     query: { enabled: parameters.account === undefined },
   })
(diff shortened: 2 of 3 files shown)


Aerilym:
Looks like some of the tests are throwing errors, not 100% if its an issue with my changes or with the tests, will take a look.

Aerilym:
Ran the tests locally, looks like 2 tests are timing out, and as far as i can tell the issue is unrelated to the code changes:
```
 FAIL  |@wagmi/core| packages/core/src/actions/getBlock.test.ts > default
 FAIL  |@wagmi/core| packages/core/src/actions/getBlock.test.ts > args: blockNumber
 FAIL  |@wagmi/core| packages/core/src/actions/getBlock.test.ts > args: includeTransactions
 
 FAIL  |wagmi| packages/react/src/hooks/useInfiniteReadContracts.test.ts > default
```

.changeset/proud-knives-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+Fixed config parameter passing in useSimulateContract and useEstimateGas
```
tmm, line 5:
```suggestion
Fixed `useSimulateContract` and `useEstimateGas` behavior by passing `config` parameter to internal `useConnectorClient`.
```

tmm (approved):
Nice catch!

PR #4452, opened 2024-12-09
feat: swaps out execa for nanoexec

douglance:
swaps out execa for nanoexec to reduce deps tree as asked for in this tweet https://x.com/awkweb/status/1866200894078918744


tmm:
Appreciate this! Decided to drop third-party packages entirely for this purpose https://github.com/wevm/wagmi/pull/4457

PR #4458, opened 2024-12-12
fix: metadata handling for empty URL

EdouardBougon:
Fix metadata handling for empty URLs.

In my previous commit, I used `||` to test for an empty string, but you replaced it with `??`, which no longer allows testing for `''`.
I have updated the test to use a ternary operator.

Let me know if you prefer a more explicit test, such as `name === ''`.

### DIFF
--- a/.changeset/nasty-planets-play.md
+++ b/.changeset/nasty-planets-play.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/connectors": patch
+---
+
+Fixed metamask internal metadata handling.
--- a/.changeset/nasty-planets-play.md
+++ b/.changeset/nasty-planets-play.md
@@ -2,4 +2,4 @@
 "@wagmi/connectors": patch
 ---
 
-Fixed metamask internal metadata handling.
+Fixed MetaMask internal metadata handling.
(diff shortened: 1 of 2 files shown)


.changeset/nasty-planets-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/connectors": patch
+---
+
+Fixed metamask internal metadata handling.
```
tmm, line 5:
```suggestion
Fixed MetaMask internal metadata handling.
```

packages/connectors/src/metaMask.ts
```diff
@@ -262,12 +262,15 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
           readonlyRPCMap,
           dappMetadata: {
             ...parameters.dappMetadata,
-            name: parameters.dappMetadata?.name ?? 'wagmi',
-            url:
-              parameters.dappMetadata?.url ??
-              (typeof window !== 'undefined'
+            // Test if name and url are set AND not empty
+            name: parameters.dappMetadata?.name
+              ? parameters.dappMetadata?.name
```
reslear, line 267: We already know that there's a value defined remove `?`: 

`parameters.dappMetadata?.name ? parameters.dappMetadata.name : ''`

PR #4463, opened 2024-12-15
Fix bug when connector does not have any methods

mqklin:
Sometimes connector does now have any methods, `connector.disconnect is not a function` thrown
image

This is how `config.state.connections[current].connector` look like (there is no other connections):
image


mqklin:
Closed it cause it doesn't solve the issue. Thread related: https://discord.com/channels/1156791276818157609/1156791580938739822/threads/1317862357191495760

PR #4472, opened 2024-12-20
fix: default storage `setItem`

tmm:
Handles `setItem` errors in default storage


sergeichestakov (approved):
looks good to me, thank you! 🙏

PR #4518, opened 2025-01-24
Chains are displayed correctly on the site

fe-dudu:
# as-is
image
Previously, not all chains were visible, and the versions were mismatched.

# to-be
image


tmm:
we periodically bump the site's viem version. that's why the note is there to check viem's source for the latest.

PR #4599, opened 2025-03-27
chore: broken link replaced

Ocheretovich:
The old link led to a github page that didn't exist at all. I added a new link taken from the official site with the tag


tmm:
works fine for me

https://github.com/user-attachments/assets/b597a32d-4a89-423e-901d-8ad7b72b8ab0

Ocheretovich:
Yes indeed on the main site everything works fine, but on Github it leads to an error

PR #4622, opened 2025-04-08
fix: rdns filter for coinbase v3 connector

dan1kov:
PR fixes Coinbase connector duplication. The configuration for v3 is missing a filter for rdns

### DIFF
--- a/.changeset/cool-masks-hang.md
+++ b/.changeset/cool-masks-hang.md
@@ -2,4 +2,4 @@
 "@wagmi/connectors": patch
 ---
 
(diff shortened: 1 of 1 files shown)


.changeset/cool-masks-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/connectors": patch
+---
+
+Add rdns filter for Coinbase v3 connector
```
tmm, line 5:
```suggestion
Added `rdns` property to Coinbase Wallet v3 connector
```

PR #4629, opened 2025-04-14
Use Sourcify v2 API in sourcify plugin

allezxandre:
This PR upgrades the Sourcify plugin API endpoint from the deprecated V1 to the V2 endpoint.

FYI, it should also solve an issue where it was not possible to fetch partial-matched contracts. I hope this doesn’t break anything, but the V2 API does not seem to make the distinction anymore.

See here: <https://docs.sourcify.dev/docs/api/>

### DIFF
--- a/.changeset/loose-mirrors-yawn.md
+++ b/.changeset/loose-mirrors-yawn.md
@@ -2,4 +2,4 @@
 "@wagmi/cli": minor
 ---
 
-Use Sourcify v2 API in `sourcify` plugin
+Upgraded to Sourcify v2 API in `sourcify` plugin
--- a/.changeset/loose-mirrors-yawn.md
+++ b/.changeset/loose-mirrors-yawn.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/cli": minor
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/loose-mirrors-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/cli": minor
+---
+
+Use Sourcify v2 API in `sourcify` plugin
```
tmm, line 5:
```suggestion
Upgraded to Sourcify v2 API in `sourcify` plugin
```

tmm (changes requested):
Can we update this script too

https://github.com/wevm/wagmi/blob/main/scripts/updateBlockExplorerPluginChains.ts#L30

allezxandre:
It appears the script you mentionned @tmm is using a non-deprecated API endpoint. The Sourcify documentation still lists it under ”Others”: <https://docs.sourcify.dev/docs/api/#other>

Do you think there is something to be done there?

allezxandre:
> This works too but you can also use `?fields=abi` directly and get `response.abi` instead of the metadata.

Got it, I'll update this PR to use that instead

PR #4657, opened 2025-05-08
fix: compute proper chainId value on useReadConstracts to avoid unnecessary queryKey mutations

agsola:
fix: compute proper chainId value on useReadConstracts to avoid unnecessary queryKey mutations

Derives a unified chainId from contracts when all share the same value, otherwise falls back to the default (chainId from useChainId).

This fixes 'queryKey' mutation when connecting or changing network despite all contracts sharing the same chainId, avoid unnecessary mutations and equating to the default behaviour on useReadContract.

### DIFF
--- a/.changeset/sixty-mugs-serve.md
+++ b/.changeset/sixty-mugs-serve.md
@@ -1,0 +1,5 @@
+---
+"wagmi": minor
+---
+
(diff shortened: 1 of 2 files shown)


tmm:
Couldn't push to your branch. Closing in favor of https://github.com/wevm/wagmi/pull/4872

PR #4661, opened 2025-05-12
fix: prevent reconnect prompt when calling `connect`

magiziz:
### Description

Fixes an issue where switching between wallets would prompt the user to connect again instead of overriding the existing connection. This improves both the UX and wagmi state


tmm:
How do you reproduce this behavior?

magiziz:
@tmm try connecting to metamask -> connect to trust and connect to metamask again. You'll be prompted to connect your wallet again

tmm:
Not convinced this extra code is worth it as some wallets don't show a repeat prompt (MetaMask, Coinbase, etc.) while others do (Trust). Probably best handled on the wallet side or in your application code.

PR #4686, opened 2025-05-26
chore: update .connect to .mutate

o-az:
`useConnect()`:
- deprecate `.connect` and `.connectAsync`,
- add `.mutate` and `.mutateAsync`,
- updates tests to use `mutate`

`useDisconnect()`:
- deprecate `.disconnect` and `.disconnectAsync`,
- add `.mutate` and `.mutateAsync`,
- updates tests to use `mutate`


tmm:
Will take this on after v3 #4857

PR #4691, opened 2025-05-29
switch chain on connect in walletconnect connector

frolic:
I noticed that connecting Rainbow wallet app via WalletConnect would fail for my local app when attempting to use the connected account, because Rainbow doesn't support adding networks. This appears to happen because the `walletConnect` connector calls `switchChain` inside `getProvider`, whereas other connectors appear to do their chain check and `switchChain` call inside of `connect`:
https://github.com/wevm/wagmi/blob/f5b717ccf8a5b283263cadc984ba00b354bcefae/packages/core/src/connectors/injected.ts#L174 https://github.com/wevm/wagmi/blob/f5b717ccf8a5b283263cadc984ba00b354bcefae/packages/connectors/src/coinbaseWallet.ts#L132 https://github.com/wevm/wagmi/blob/f5b717ccf8a5b283263cadc984ba00b354bcefae/packages/connectors/src/metaMask.ts#L147

For my app, I've set `initialChainId` to `0` to kinda bypass the chain switching step here, which works because my app only needs signatures from the connected wallet, regardless of the chain it's connected to.

I'm not really sure how to go about testing this change, but this PR aligns the switch chain logic across all connectors.


frolic:
I forked this connector locally with these changes and I was able to connect without issues!

PR #4694, opened 2025-06-02
feat: Soften duplicate contract name in the case contracts have identical ABIs

silasdavis:
This is useful for instance when contracts appear as duplicates when ingesting from a single Foundry project that depends on others foundry projects.

In this case `getArtifactPaths()` in the Foundry plugin pick up the same contracts twice since they end up existing in both `out/` directories. I don't think there is a reliable to heuristic to de-duplicate them at this level.

It seems like duck-typing contracts by the ABI should be fine, it could always be enabled via an option.


silasdavis:
I just want to put this out for a sense check before going to town on the rest of contribution requirements, let me know what you think.

silasdavis:
closing for correct description of root cause after debugging some more: https://github.com/wevm/wagmi/pull/4695

PR #4695, opened 2025-06-02
feat: Soften duplicate contract name in the case contracts have identical ABIs

silasdavis:
Foundry seems to duplicate some imported sub-files within a contract, for example:

`<foundry root>/src/parts/primitives/Foo.sol`

Is defined only once, and imported from `<foundry root>/src/Bar.sol`

But ends up appearing twice in artifacts directory:

```
<foundry root>/out/Foo.sol/Foo.json
<foundry root>/out/primitives/Foo.sol/Foo.json
```

In this case `getArtifactPaths()` in the Foundry plugin pick up the same contracts twice since they end up existing in multiple subdirectories of `out/`. I don't think there is a reliable to heuristic to de-duplicate them at this level.

It seems like duck-typing contracts by the ABI should be fine and may help other plugin sources. it could always be disabled by default and enabled via an option at the cost of yet more config

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -6,7 +6,7 @@ import { watch } from 'chokidar'
 import { default as dedent } from 'dedent'
 import { basename, dirname, resolve } from 'pathe'
 import pc from 'picocolors'
-import { type Abi, type Address, getAddress } from 'viem'
+import { type Abi, type Address, getAddress, keccak256 } from 'viem'
 import { z } from 'zod'
 
 import type { Contract, ContractConfig, Plugin, Watch } from '../config.js'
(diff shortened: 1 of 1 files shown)


silasdavis:
I just want to put this out for a sense check before going to town on the rest of contribution requirements, let me know what you think.

silasdavis:
A bit more on why these duplicates might be seen from foundry,

These snippets are are target from two duplicated foundry artifacts.

the first appears under `out/`:

```json
      "compilationTarget": {
        "src/Foo.sol": "Foo"
      },
```

the second under `out/src/`:

```json
      "compilationTarget": {
        "../project/src/Foo.sol": "Foo"
      },
```

Where we are compiling in the Foundry project `project`, but we do have some other Foundry projects that import from each other, so it seems to be that the way Foundry processes the Solidity compiler output causes these duplicates, which probably isn't really a bug, so would be nice if wagmi supports this.

silasdavis:
I've implemented a similar fix in the `foundry.ts` plugin source itself which I am using as a local vendored patch, let me know if you'd prefer that version.

tmm:
@silasdavis sorry it's taken so long to get back to you!

> I've implemented a similar fix in the `foundry.ts` plugin source itself which I am using as a local vendored patch, let me know if you'd prefer that version.

yes

PR #4705, opened 2025-06-10
chore: remove connectors dependency from @wagmi/react

o-az:
- **chore: rm connectors dependency from @wagmi/react**
- **feat: update playgrounds and doc page**


tmm:
Will take care of this in the next major version!

PR #4711, opened 2025-06-25
fix(core): update chainId retrieval logic to consider syncConnectedChain configuration

### LINKED ISSUE #4680: [wagmi-cli] hooks default to account chanId instead of global config
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/wagmi/issues) for the bug I encountered.

### Describe the bug

generated hooks use the connected account chainId instead of the `configChainId`:
https://github.com/wevm/wagmi/blob/f5b717ccf8a5b283263cadc984ba00b354bcefae/packages/core/src/actions/codegen/createReadContract.ts#L80-L83

because of that, setting [`syncConnectedChain`](https://wagmi.sh/react/api/createConfig#syncconnectedchain) to `false` has no effect on wagmi-cli generated hooks, so you have to explicitly provide a chainId to every hook if you want them to work regardless of the connected network

_No response_

### Steps To Reproduce

1. generate hooks with wagmi-cli targetting an Ethereum mainnet contract
2. use one of those hooks in your dapp, without specifying chainId
3. open dapp, switch network to any chain other than Ethereum mainnet
4. hook won't work

@wagmi/cli

2.3.1

### Viem Version

2.29.4

### TypeScript Version

5.8.3

### Anything else?

_No response_

yigiterdev:
## Problem
Fixes #4680 
Generated hooks from wagmi-cli were ignoring the syncConnectedChain configuration setting and always prioritizing the connected account's chainId over the config's default chainId. This meant that setting syncConnectedChain: false had no effect on generated hooks, requiring users to explicitly provide a chainId parameter to every hook call when they wanted hooks to work regardless of the connected network.

## Solution
Updated the chainId resolution logic to respect the syncConnectedChain configuration:

### Behavior after fix:
```
When syncConnectedChain: true (default): Uses account.chainId if available, falls back to configChainId
When syncConnectedChain: false: Uses configChainId directly, ignoring connected account's chain
```

### DIFF
--- a/packages/core/src/actions/codegen/createReadContract.ts
+++ b/packages/core/src/actions/codegen/createReadContract.ts
@@ -79,8 +79,9 @@ export function createReadContract<
       const account = getAccount(config)
       const chainId =
         (parameters as { chainId?: number })?.chainId ??
-        account.chainId ??
-        configChainId
+        (config._internal.syncConnectedChain
+          ? (account.chainId ?? configChainId)
+          : configChainId)
       return readContract(config, {
         ...(parameters as any),
         ...(c.functionName ? { functionName: c.functionName } : {}),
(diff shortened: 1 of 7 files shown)


tmm:
Thanks for this! Sorry in the delay for taking a look. Wasn't able to push to your branch so closing in favor of https://github.com/wevm/wagmi/pull/4773

PR #4735, opened 2025-07-17
Adding baseAccount Connector for Base Account

fan-zhang-sv:
We’ve recently rebranded Coinbase Smart Wallet to Base Account, and launched a dedicated SDK for it [here](https://github.com/base/account-sdk).

Our recommended migration path for this migration/transition is to introduce a separate Base Account connector specifically for the Base Account (Smart Wallet), while the existing Coinbase Wallet connector will continue to support both EOA users and Smart Wallets to ensure a smooth migration. More [details](https://docs.base.org/base-account/guides/migration-guide) 

This PR adds the new connector for Base Account.

### DIFF
--- a/packages/connectors/package.json
+++ b/packages/connectors/package.json
@@ -45,6 +45,7 @@
     }
   },
   "dependencies": {
+    "@base-org/account": "1.0.2",
     "@coinbase/wallet-sdk": "4.3.6",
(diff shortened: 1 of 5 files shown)


jxom:
Duplicate of #4734. Let’s consolidate for that one!

PR #4737, opened 2025-07-22
feat: add nuxt 4 support for wagmi nuxt module

microHoffman:
I've recently started a new Nuxt 4 project, but I found out that officially supported version on Wagmi Nuxt module is only Nuxt 3. In this PR I've added support for Nuxt 4 as well. I've published the updated package with Nuxt 4 support here: https://www.npmjs.com/package/@microhoffman/vue , to test if it's working fine in my Nuxt 4 codebase and it seems to work just fine, so I'm creating this PR with extending support to Nuxt 4.

Let me know please if there is anything else I should tackle, adjust some tests, or maybe something else... Thanks!


SwartZCoding:
Please merge this ! No changes are required. Just bump the nuxt version on package.json :)

PR #4740, opened 2025-07-24
feat: upgrade vitest, biome

tmm:
Upgrades Vitest, Biome


packages/create-wagmi/templates/vite-react/biome.json
```diff
@@ -1,13 +1,26 @@
 {
+  "root": false,
```
tmm, line 2: remove this

PR #4741, opened 2025-07-25
fix: ReferenceError indexedDB is not defined during SSR

eliezerbasubi:
Fix issue with indexedDB being undefined during SSR. This issue can be experienced when using `rainbowkit`.

https://github.com/rainbow-me/rainbowkit/issues/2476


tmm:
Best to fix upstream in `@walletconnect/ethereum-provider`

PR #4752, opened 2025-08-04
fix(deepEqual): support objects with null prototypes

iamakulov:
While reviewing a project crash for a client, I discovered that the `deepEqual` function ([invoked by getConnectors](https://github.com/wevm/wagmi/blob/e97836b255b4d5a3d7d9b32a9b99c6340045c6c5/packages/core/src/actions/getConnectors.ts#L14)) crashes if some field inside a connector is an object without a prototype:

```js
const a = Object.create(null)

if (a.valueOf !== Object.prototype.valueOf)
    return a.valueOf() === b.valueOf();
          // ↑ fails with “TypeError: a.valueOf is not a function” since an object
          // with no prototype will have `a.valueOf === undefined`
```

This PR fixes that.

### DIFF
--- a/.changeset/rich-teams-care.md
+++ b/.changeset/rich-teams-care.md
@@ -2,4 +2,4 @@
 "@wagmi/core": patch
 ---
 
-Fix deepEqual crashing for objects with no prototype
+Fixed `deepEqual` crashing for objects with no prototype
--- a/packages/core/src/actions/getConnectors.test.ts
+++ b/packages/core/src/actions/getConnectors.test.ts
@@ -5,4 +5,5 @@ import { getConnectors } from './getConnectors.js'
 
 test('default', () => {
   expect(getConnectors(config)).toEqual(config.connectors)
+  expect(getConnectors(config)).toEqual(config.connectors)
 })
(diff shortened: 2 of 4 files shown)


iamakulov:
Dug deeper into this. Another issue this `deepEqual` implementation has is it also crashes with `Maximum call stack size exceeded` if any passed object has circular references.

Critically, it’s pretty easy to end up that state. I still don’t have the full picture of how exactly it’s happening, but it appears that:
- Wagmi adds an `emitter` (instance of [eventemitter3](https://github.com/primus/eventemitter3)) object onto every connector created through it. This object has both `__proto__ === null` and a circular reference to itself
- Whenever you call `getConnectors()`, Wagmi [deep-compares previous and new connectors](https://github.com/wevm/wagmi/blob/e97836b255b4d5a3d7d9b32a9b99c6340045c6c5/packages/core/src/actions/getConnectors.ts#L14)
- So if you ever recreate a connector from scratch (which can easily happen during SSR, where the React state gets reset on every request), Wagmi will end up deep-comparing the previous and the new connector – and crashing inside `deepEqual` when it tries to compare `emitter` objects

I guess the answer to this is “don’t store Wagmi connectors in your React state, duh”. But it would’ve been nice if Wagmi had a better DX around this (rather than just crashing with `a.valueOf is not a function` or `Maximum call stack size exceeded`).

tmm:
@iamakulov thanks for this! removed `deepEqual` from `getConnectors` and can revisit more improvements later.

PR #4756, opened 2025-08-07
Use getAction for sendCalls & getCallsStatus

plusminushalf:
Adds the ability for clients to implement custom sendCalls and getCallsStatus. This will enable libraries like permissionless to support useSendCalls for smart accounts.


johanneskares:
👏

PR #4784, opened 2025-09-15
feat: expose connect capabilities

tmm:
Adds `withCapabilities` flag to expose response capabilities in `connector.connect()`, `connect()`, `useConnect().connect(Async)()`. This will become the default behavior in Wagmi v3.

https://eips.ethereum.org/EIPS/eip-7846#capabilities


stephancill:
hey @tmm i'm having a hard time understanding where the capabilities for connection are configured in this model e.g. how to pass the nonce to the signInWithEthereum capability, would you mind elaborating a bit?

tmm:
@stephancill Passing connect parameters based on the connector was already supported in https://github.com/wevm/wagmi/pull/4453. You just need to add a [`connect` property](https://github.com/wevm/wagmi/blob/main/packages/core/src/connectors/mock.ts#L52-L61) to the `Properties` you pass to `createConnector`. For example, the `mock` connector has a custom [`foo` parameter](https://github.com/wevm/wagmi/blob/main/packages/core/src/connectors/mock.ts#L56). Can see it in action in the [type](https://github.com/wevm/wagmi/blob/main/packages/core/src/actions/connect.test-d.ts#L21) [tests](https://github.com/wevm/wagmi/blob/main/packages/react/src/hooks/useConnect.test-d.ts#L82).

The same concept now works for connector return types (in this case `capabilities` when the `withCapabilities` flag is `true`). `mock` connector returns a [`foo` capability](https://github.com/wevm/wagmi/pull/4784/files#diff-f23248209841a6752b44445e48fb2b5f9e90c2f5755049e8eb6afa2f1e56d7dc). If you want `baseAccount` to support this, then you need to [restore the `Properties` type](https://github.com/wevm/wagmi/pull/4784/files#diff-b4090895fcaf82f372184f35827f7f5e88cf6273699efcd2b04695568017c762L28-L36) and add in you capabilities to parameters and return type. There's another example of this in https://github.com/ithacaxyz/porto/pull/862

---

Wagmi v3 will make `capabilities` in the return type the default so `withCapabilities` is short term for backwards compatibility. Might as add a top-level `capabilities: Record<string, unknown>` property to connect parameters in v3 too so connectors that don't explicitly define a `Properties` will allow passing `capabilities` (although not strongly typed) as well.

stephancill:
@tmm thanks for your response, that makes sense. so will this require the developer to narrow the type of the connector before connecting? or is there a generic way to implement using the `signInWithEthereum` capability with `nonce` if available?

tmm:
@stephancill no need to narrow. as long as you [register your Wagmi config with TypeScript or pass it directly](https://wagmi.sh/react/typescript#config-types), works fine. Added an example type test in https://github.com/wevm/wagmi/pull/4784/commits/8303900b4f8c8ed943050cc8b58053af3f80162a

PR #4824, opened 2025-10-08
chore: upgrade `@walletconnect/ethereum-provider` to `2.22.1`

svenvoskamp:
chore: upgrade `@walletconnect/ethereum-provider` to `2.22.1`


tomiiide:
@svenvoskamp  ah ah, looks like I'm not the only one having issues with this

Issue #4828, opened 2025-10-13
eth_call should not contains `nonce` in `waitTransactionReceipt`


clouds56:
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/wagmi/issues) for the bug I encountered.

### Describe the bug

The txn returned by `action_getTransaction` has `"accessList", "data", "gas", "maxFeePerGas", "maxPriorityFeePerGas", "nonce", "to", "value"`, where according to [eth_call](https://www.quicknode.com/docs/ethereum/eth_call) doc, only `from, to, gas, gasPrice, value, data` are accepted

https://github.com/wevm/wagmi/blob/6159db8c9c45b724c0de0fc44a90651d66449546/packages/core/src/actions/waitForTransactionReceipt.ts#L66-L75

### What Wagmi package(s) are you using?

@wagmi/vue

### Wagmi Package(s) Version(s)

0.2.11

### Viem Version

2.38.0

clouds56:
And it SHOULD also include `from` in `eth_call` which is recovered from the transaction.

PR #4829, opened 2025-10-13
fix nonce and account when eth_call in waitForTransactionReceipt

clouds56:
fixes #4828

PR #4832, opened 2025-10-13
fix(core): adds `account` to internal `call` in `waitForTransactionReceipt` for reverted transaction

tmm:
Related https://github.com/wevm/wagmi/pull/4829
Closes https://github.com/wevm/wagmi/issues/4828

### DIFF
--- a/.changeset/wise-foxes-smash.md
+++ b/.changeset/wise-foxes-smash.md
@@ -2,4 +2,4 @@
 "@wagmi/core": patch
 ---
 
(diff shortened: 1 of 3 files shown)


clouds56:
May I ask why we pass the same nonce to eth_call?
Since the transaction is settled (even reverted), the next nonce should be at least original `nonce+1`.

AFAIK, anvil would check the nonce for eth_call, we would get `NonceTooLow` error if the RPC endpoint is anvil.

Ref: https://github.com/foundry-rs/foundry/issues/12077

PR #4833, opened 2025-10-16
Bump to latest base account version

KodiKraig:
- Update connectors to use latest `@base-org/account` version

### DIFF
--- a/packages/connectors/package.json
+++ b/packages/connectors/package.json
@@ -53,7 +53,6 @@
     "@safe-global/safe-apps-sdk": "9.1.0",
     "@walletconnect/ethereum-provider": "2.21.1",
     "cbw-sdk": "npm:@coinbase/wallet-sdk@3.9.3",
-    "fastestsmallesttextencoderdecoder": "1.0.22",
     "porto": "0.2.19"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/connectors/package.json
```diff
@@ -45,14 +45,15 @@
...
-    "@base-org/account": "1.1.1",
+    "@base-org/account": "2.4.0",
     "@coinbase/wallet-sdk": "4.3.6",
     "@gemini-wallet/core": "0.2.0",
     "@metamask/sdk": "0.33.1",
     "@safe-global/safe-apps-provider": "0.18.6",
     "@safe-global/safe-apps-sdk": "9.1.0",
     "@walletconnect/ethereum-provider": "2.21.1",
     "cbw-sdk": "npm:@coinbase/wallet-sdk@3.9.3",
+    "fastestsmallesttextencoderdecoder": "1.0.22",
```
jxom:
```suggestion
```

tmm: `fastestsmallesttextencoderdecoder` is not a peer dependency of `@base-org/account` (you can check its package.json) so seems like it's the responsibility for `@solana/kit` to set it since a bunch of dependencies it uses (e.g. `@solana/*`) have `fastestsmallesttextencoderdecoder` as a peer dependency.

tmm:
Closing in favor of https://github.com/wevm/wagmi/pull/4843

PR #4853, opened 2025-11-03
docs: Correct useBlockNumber example in migration docs

2wheeh:
type of `blockNumber` is `bigint | undefined`, so this is correct.


tmm:
It's valid JS so keeping for sake of brevity.

2wheeh:
@tmm I wish it works this way, but it's not valid JS. It throws [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cant_convert_BigInt_to_number):

`Uncaught TypeError: Cannot mix BigInt and other types, use explicit conversions`

PR #4900, opened 2025-12-15
fix: vite-react template

stephancill:
When creating a new project with `pnpm create wagmi`, the new project is missing the `buffer` polyfill dependency and has a type error in `main.tsx` where it patches the global buffer.

This PR removes the `buffer` re-assignment since it is no longer required.


packages/create-wagmi/templates/vite-react/src/main.tsx
```diff
@@ -9,7 +9,7 @@ import { config } from './wagmi.ts'
 
 import './index.css'
 
-globalThis.Buffer = Buffer
```
tmm: Let's just remove this line entirely so no need for buffer at all

tmm (approved):

PR #4908, opened 2025-12-22
Feat: Add Foundry broadcasts to `deployments` mapping

jamesmccomish:
# Problem
The [deployments mapping](https://github.com/wevm/wagmi/blob/main/site/cli/api/plugins/foundry.md#deployments) in `FoundryConfig` is useful for attaching contracts to addresses. But often when developing locally in a monorepo I run anvil (or a local fork) and deploy my contracts there. The addresses might change as the contracts change, and keeping the deployments mapping updated is awkward, it seems like a small addition to `wagmi/cli` could make this a lot easier.

https://github.com/wevm/wagmi/blob/29f95df3aeaa9beab54b9bf91787bb10d44dbf26/site/cli/api/plugins/foundry.md?plain=1#L52-L57

# Update
This PR adds a flag `includeBroadcasts` which automatically populates the deployments mapping with any contracts deployed in a `run-latest.json` file in the `/broadcasts/` folder. 

https://github.com/wevm/wagmi/blob/80c777f501d1a7b2b455e73fdc1137304bd7b538/packages/cli/src/plugins/foundry.ts#L55-L62

With that flag set, running `wagmi generate` in a local development script after deploying any test contracts locally will take the latest deployment addresses and generate files that can be read by the frontend.

# Further Work
I think more info from these broadcast files could be useful to extract too. Things like `blockNumber` or deployment arguments might be useful for linking to indexers etc.

### DIFF
--- a/.changeset/crazy-moose-remain.md
+++ b/.changeset/crazy-moose-remain.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/cli": minor
+---
+
+Added `includeBroadcasts` option to Foundry plugin to automatically populate deployments mapping from `run-latest.json` files in the `broadcast/` directory.
--- a/packages/cli/src/plugins/foundry.test.ts
+++ b/packages/cli/src/plugins/foundry.test.ts
@@ -1,5 +1,5 @@
+import fs from 'node:fs/promises'
 import fixtures from 'fixturez'
-import { promises as fs } from 'fs'
 import { dirname, resolve } from 'pathe'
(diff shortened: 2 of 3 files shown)


lennardevertz:
yes please!

tmm:
@jamesmccomish this ready for review?

jamesmccomish:
@tmm yes. just got the branch up to latest and now also handle the new broadcast contracts when there's a change

PR #4948, opened 2026-01-12
fix #4352: prevent re-subscription in useWatchBlocks and useWatchBloc…

### LINKED ISSUE #4352: `useWatchBlocks` subscribes / unsubscribes on each new Block with WebSocket
### Check existing issues

### Describe the bug

When using `useWatchBlocks` with a WebSocket connection, the expected behavior is as follows:
1) Subscribe to new blocks
2) Receive (`onBlock`) callbacks without unsubscribing
3) Unsubscribe when the component using `useWatchBlocks` is unmounted / modified

The observed behavior is:
1) Subscribe to new heads
2) Receive a block, execute `onBlock`
3) Unsubscribe as a dependency seems to change in the implementation of `useWatchBlocks`, probably caused by  `useEffect` in the `useWatchBlocks` implementation
4) Re-Subscribe and repeat

It's important to be aware that subscribing / Unsubscribing can be an expensive call on certain API providers (e.g. 10 Compute Units on Alchemy), causing unnecessary spikes in API Usage. 

[Alchemy Compute Unit Costs](https://docs.alchemy.com/reference/compute-unit-costs)

_No response_

### Steps To Reproduce

1) Configure Wagmi with a WebSocket RPC endpoint
2) Use the snippets below, which only slightly modify the `useWatchBlocks` code to get console log outputs
You will observe that the subscription is closed after each block, and a new subscription is established.

Code to use the modified hook:
```
  useWatchBlocks({
    emitOnBegin: true,
    chainId: Constants.CHAIN_ID,
    onBlock: (b) => {
      console.log(b)
    }
  })
  ```
  
  The modified hook:
  ```
  'use client'

import {
  type Config,
  type ResolvedRegister,
  type WatchBlocksParameters,
  watchBlocks,
} from '@wagmi/core'
import type { UnionCompute, UnionExactPartial } from '@wagmi/core/internal'
import { useEffect } from 'react'
import type { BlockTag } from 'viem'

// Import wagmi config and constants for test
import { config } from '../AppKit'

export type EnabledParameter = {
  enabled?: boolean | undefined
}

export type ConfigParameter<config extends Config = Config> = {
  config?: Config | config | undefined
}

export type UseWatchBlocksParameters<
  includeTransactions extends boolean = false,
  blockTag extends BlockTag = 'latest',
  config extends Config = Config,
  chainId extends
  config['chains'][number]['id'] = config['chains'][number]['id'],
> = UnionCompute<
  UnionExactPartial<
    WatchBlocksParameters<includeTransactions, blockTag, config, chainId>
  > &
  ConfigParameter<config> &
  EnabledParameter
>

export type UseWatchBlocksReturnType = void

/** https://wagmi.sh/react/hooks/useWatchBlocks */
export function useWatchBlocks<
  config extends Config = ResolvedRegister['config'],
  chainId extends
  config['chains'][number]['id'] = config['chains'][number]['id'],
  includeTransactions extends boolean = false,
  blockTag extends BlockTag = 'latest',
>(
  parameters: UseWatchBlocksParameters<
    includeTransactions,
    blockTag,
    config,
    chainId
  > = {} as any,
): UseWatchBlocksReturnType {
  const { enabled = true, onBlock, config: _, ...rest } = parameters
  const chainId = parameters.chainId

  // TODO(react@19): cleanup
  // biome-ignore lint/correctness/useExhaustiveDependencies: `rest` changes every render so only including properties in dependency array
  useEffect(() => {
    if (!enabled) return
    if (!onBlock) return

    // Slightly modify for console logs
    console.log('WATCH')
    const unwatch = watchBlocks(config, {
      ...(rest as any),
      chainId,
      onBlock,
    })

    return () => {
      console.log('UNWATCH')
      unwatch()
    }
  }, [
    chainId,
    config,
    enabled,
    onBlock,
    ///
    rest.blockTag,
    rest.emitMissed,
    rest.emitOnBegin,
    rest.includeTransactions,
    rest.onError,
    rest.poll,
    rest.pollingInterval,
    rest.syncConnectedChain,
  ])
}
```

wagmi

2.12.17

### Viem Version

2.21.21

### TypeScript Version

5.6.0

### Anything else?

_No response_

mishop-15:
### Problem
`useWatchBlocks` and `useWatchBlockNumber` were re-subscribing on every block when using WebSocket transport. This caused unnecessary API calls and increased compute unit costs (e.g., 10 CU per subscribe/unsubscribe on Alchemy).

### Root Cause
The `onBlock` and `onBlockNumber` callbacks were included in the `useEffect` dependency array, causing the effect to re-run on every render when the callbacks changed.

### Solution
- Store callbacks in refs (`useRef`)
- Update refs in separate `useEffect` hooks
- Main subscription effect excludes callback dependencies
- Subscription remains stable while always calling the latest callback version

### Changes
- `packages/react/src/hooks/useWatchBlocks.ts`
- `packages/react/src/hooks/useWatchBlockNumber.ts`

### Testing
Tested with WebSocket transport on devnet - verified that subscription persists across blocks without re-subscribing.

### DIFF
--- a/packages/react/src/hooks/useWatchBlockNumber.ts
+++ b/packages/react/src/hooks/useWatchBlockNumber.ts
@@ -7,7 +7,7 @@ import {
   watchBlockNumber,
 } from '@wagmi/core'
 import type { UnionCompute, UnionExactPartial } from '@wagmi/core/internal'
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 
 import type { ConfigParameter, EnabledParameter } from '../types/properties.js'
 import { useChainId } from './useChainId.js'
@@ -39,6 +39,20 @@ export function useWatchBlockNumber<
   const configChainId = useChainId({ config })
   const chainId = parameters.chainId ?? configChainId
 
+  // store callbacks in refs to prevent re-subscriptions when they change.
+  // refs are updated separately (in useEffect below) so the subscription
+  // remains stable while always calling the latest callback version.
+  const onBlockNumberRef = useRef(onBlockNumber)
+  const onErrorRef = useRef(rest.onError)
+
+  useEffect(() => {
+    onBlockNumberRef.current = onBlockNumber
+  }, [onBlockNumber])
+
+  useEffect(() => {
+    onErrorRef.current = rest.onError
+  }, [rest.onError])
+
   // TODO(react@19): cleanup
   // biome-ignore lint/correctness/useExhaustiveDependencies: `rest` changes every render so only including properties in dependency array
   useEffect(() => {
@@ -47,15 +61,16 @@ export function useWatchBlockNumber<
     return watchBlockNumber(config, {
       ...(rest as any),
       chainId,
-      onBlockNumber,
+      onBlockNumber: (blockNumber, prevBlockNumber) => {
+        onBlockNumberRef.current?.(blockNumber, prevBlockNumber)
+      },
+      onError: (error) => onErrorRef.current?.(error),
     })
   }, [
     chainId,
     config,
     enabled,
-    onBlockNumber,
     ///
-    rest.onError,
     rest.emitMissed,
     rest.emitOnBegin,
     rest.poll,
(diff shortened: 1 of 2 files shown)


tmm:
Thanks for the PR! Going with https://github.com/wevm/wagmi/commit/b5d6911a863e35d677374acd744c054daacd663a instead

PR #4968, opened 2026-01-21
fix: bind WebAuthn response methods to fix illegal invocation error

gorried:
Password manager extensions (1Password, etc.) intercept WebAuthn calls and return proxy objects where response methods lose their `this` binding. Explicitly bind methods with `.call(response)` to prevent this error.

Verified via testing behavior against a project that uses WebAuthn calls


jxom (approved):
lgtm

PR #4992, opened 2026-02-04
fix(solid): restrict @tanstack/solid-query peer dep to avoid type reg…

2wheeh:
…ression

## Summary

- Restrict `@tanstack/solid-query` peer dependency to `>=5.0.0 <5.71.9` to avoid type inference regression

## Context

Starting from `@tanstack/solid-query@5.71.9`(https://github.com/TanStack/query/pull/8950), there's a type inference regression where generics are lost on deprecated `create*` APIs (e.g., `createQuery`, `createMutation`). 

This affects `@wagmi/solid` which still uses these APIs.

Once the upstream fix is released, we'll update the peer dependency constraint accordingly.

Related https://github.com/TanStack/query/pull/10093

### DIFF
--- a/.changeset/red-ways-sniff.md
+++ b/.changeset/red-ways-sniff.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/solid": patch
+---
+
+fix(solid): restrict @tanstack/solid-query peer dep to avoid type regression
--- a/.changeset/red-ways-sniff.md
+++ b/.changeset/red-ways-sniff.md
@@ -2,4 +2,4 @@
 "@wagmi/solid": patch
 ---
 
-fix(solid): restrict @tanstack/solid-query peer dep to avoid type regression
+Restrict `@tanstack/solid-query` peer dep version to avoid type regression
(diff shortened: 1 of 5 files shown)


.changeset/red-ways-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"@wagmi/solid": patch
+---
+
+fix(solid): restrict @tanstack/solid-query peer dep to avoid type regression
```
tmm:
```suggestion
Restrict `@tanstack/solid-query` peer dep version to avoid type regression
```

2wheeh:
@tmm https://github.com/TanStack/query/pull/10093 is merged, it's compatible with the latest (`>= 5.90.24`) now.

tmm:
Thanks! Sorry it took so long to respond. Going to close this since if folks install TanStack latest, they will get `>= 5.90.24` (versions before will have issue, but seems fine since package is alpha and we don't need to change peer dep range then).

PR #4994, opened 2026-02-05
feat(tempo): sign arbitrary hash during webAuthn connect

deodad:
Adds `capabilities.sign: { hash }` to the webAuthn connector's `connect` method. This lets callers get a WebAuthn signature over an arbitrary hash during the connect gesture, returned as `signature?: Hex` in the connect result.

Use case: external parties can obtain a signed key authorization (or any other digest) during connect, then submit it on-chain independently — without the connector needing to manage the access key lifecycle.

## Changes

- **`packages/core/src/tempo/Connectors.ts`** — extend webAuthn `connect` parameters with `capabilities.sign: { hash: Hex }` and return type with `signature?: Hex`. When `signHash` is provided, access key provisioning is skipped across all paths — the WebAuthn gesture is used exclusively for signing the hash.
  - *Create*: `createCredential` cannot sign, so `account.sign(hash)` produces the signature (2nd gesture)
  - *Stored credential*: uses `account.sign(hash)` (1 gesture, no WebAuthn prompt for credential)
  - *Discover credential*: folds hash into `WebAuthnP256.getCredential()` for single-gesture signing, returns serialized `SignatureEnvelope` directly
- **`packages/core/src/actions/connect.ts`** — infer `signature` type from connector's connect return type (same pattern as `capabilities` inference), forward `signature` in return value
- **`packages/react/src/hooks/useConnect.test-d.ts`** — add `signature?: unknown | undefined` to all `ConnectReturnType` assertions
- **`packages/react/src/tempo/Connectors.test.ts`** — two new browser tests using CDP virtual WebAuthn authenticators:
  - `connect with sign hash` — sign-up, disconnect, reconnect with hash, verify signature
  - `connect with sign hash (without grantAccessKey)` — same flow without `grantAccessKey` on connector

## Testing

```bash
pnpm --filter @wagmi/core check:types  # passes
pnpm --filter wagmi check:types        # passes
pnpm format                            # passes
```

Browser tests run in Playwright with CDP virtual WebAuthn authenticators.

### DIFF
--- a/.changeset/short-worms-yawn.md
+++ b/.changeset/short-worms-yawn.md
@@ -1,0 +1,6 @@
+---
+"@wagmi/core": patch
+"wagmi": patch
+---
+
+Added ability to sign arbitrary `hash` on `webAuthn#connect`.


packages/core/src/tempo/PLAN-connect-access-key.md
jxom: rm this one

PR #5018, opened 2026-03-05
feat: update tanstack query version to the latest

2wheeh:
~~closes #4992~~

## Background

The upstream fix (TanStack/query#10093 related to #4992) hadn't been merged for some time, so I decided to migrate `@wagmi/solid` to the `use*` APIs directly and bump all TanStack Query packages to the latest.

During the update, I found that the generics had already diverged between wagmi's type wrappers and TanStack Query's latest types.

## Changes

1. `@tanstack/*-query` version bumps
2.  removes dropped `TQueryData` from `InfiniteQueryObserverOptions` and all wrapper types that forwarded it
3. `MutateOptions` callback signature change:

```ts
// Before (v5.49.x)
onSuccess?: (data, variables, context: TContext) => void
```
```ts
// After (v5.90.x)
onSuccess?: (data, variables, onMutateResult: TOnMutateResult | undefined, context: MutationFunctionContext) => void
```

The old `context` (return value of `onMutate`) was renamed to `onMutateResult`, and a new `context`, type of which is [`MutationFunctionContext`](https://github.com/TanStack/query/blob/fe25edc2f9fb231004563fb65320983b5a5a905e/packages/query-core/src/types.ts#L1098), was added as an extra param.

4.  `client: QueryClient` added to `QueryFunctionContext`

Updated `getFeeHistory.test.ts` which manually constructs the context object.

5. Solid Query: `create*` → `use*` API migration

## Questions
- Should we bump peerDeps to `>=5.90.0` across the board to match the versions we're actually testing against?
- Should we also rename the context generic parameter to `onMutateResult` in wagmi's type wrappers to align with the upstream naming? (easy but quite a lot of changes)


2wheeh:
will resolve conflicts soon this week

tmm:
Closing this out since the original Solid Query compatibility issue was resolved upstream and #4992 was closed. We can revisit a TanStack Query upgrade separately if needed. Thanks for the PR!

2wheeh:
Hi @tmm, with viem v3 now in prerelease, is a corresponding major release of wagmi planned? 

If so, would you be open to including a TanStack Query upgrade, now that Solid v2 is in RC and Solid Query for the v2 depends on the latest query core?

I'd be happy to update this work against the current main and open a fresh PR if that would be helpful. Please let me know!

PR #5027, opened 2026-03-20
feat: add `crossDomainAccessKey` connector

figtracer:
**Description:**

adds a new `crossDomainAccessKey` connector to `wagmi/tempo` that lets third-party apps use the user's existing tempo wallet without per-domain passkeys.

### problem

webauthn passkeys are domain-bound. a passkey created on `tempo.xyz` can't be used on `varzim.figtracer.com`. every mpp app needs its own funded wallet, which is terrible ux.

### solution

generates a secp256k1 key locally and authorizes it via a popup to the domain where the user's passkey lives (e.g. `tempo.xyz/authorize`). follows the same access key pattern as the CLI.

**flow:**
1. app calls `connect()` → generates secp256k1 private key, stores in wagmi storage
2. opens popup to `authorizationUrl` with key address + authorization hash
3. user approves with their passkey on the relay page
4. signed `KeyAuthorization` comes back via `postMessage`
5. connector includes the authorization in `prepareTransactionRequest`
6. subsequent connects restore from storage (no popup)

**usage:**
```ts
import { crossDomainAccessKey } from '@wagmi/core/tempo'

const config = createConfig({
  connectors: [
    crossDomainAccessKey({
      authorizationUrl: 'https://tempo.xyz/authorize',
    }),
  ],
})
```

handles expiry checks, storage persistence, popup lifecycle, and origin verification.

companion PR: tempoxyz/tempo-ts — `authorizationRelay` handler https://github.com/tempoxyz/tempo-ts/pull/154

### DIFF
--- a/packages/core/src/tempo/Connectors.ts
+++ b/packages/core/src/tempo/Connectors.ts
@@ -854,3 +854,378 @@ export declare namespace dangerous_secp256k1 {
     account?: LocalAccount | undefined
   }
 }
+
+crossDomainAccessKey.type = 'crossDomainAccessKey' as const
+
+/**
+ * Connector for cross-domain access key authorization.
+ *
+ * Generates a secp256k1 key locally and authorizes it via a popup
+ * to the domain where the user's passkey wallet lives (e.g., tempo.xyz).
+ * This solves the WebAuthn rpId domain-binding limitation — users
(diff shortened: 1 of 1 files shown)


jrin0340-420 (commented):
rindfussjoshua169@gmail.com would be the same person that we're all one but you'll see I've got about six or seven f****** different emails

PR #5051, opened 2026-04-08
feat(tempo): connector exports

tmm:
Adds Tempo connector exports


packages/core/src/tempo/Connectors.ts
jxom, line 1: Think we might want to eventually deprecate `webAuthn` and `dangerous_secp256k1` (and the `KeyManager` module) in favor of Accounts SDK here too which could remove a lot of LoC.

tmm: yup, part of my plan

PR #5053, opened 2026-04-10
fix(core): clean up injected switchChain listener

pauloZion1:
This cleans up the pending `change` listener in `injected.switchChain()` whenever the switch request fails, and recreates the listener for the `wallet_addEthereumChain` fallback path so rejected switches do not leave stale subscriptions behind.

Adds focused regression tests for both direct `wallet_switchEthereumChain` rejection and `wallet_addEthereumChain` rejection.


tmm:
Closing as a duplicate of #5126, which includes the injected switchChain listener cleanup plus a changeset and 4902 retry coverage. Thanks for the PR!

PR #5070, opened 2026-04-22
refactor(connectors): upgrade @metamask/connect-evm

wenfix:
## Summary

- Upgrades `@metamask/connect-evm` to `1.0.0` (catalog + `@wagmi/connectors` peer range).
- Aligns the MetaMask connector with the SDK's new return shapes:
  - `connectAndSign()` now resolves to `{ accounts, chainId, signature }` (was a raw signature string).
  - `connectWith()` now resolves to `{ accounts, chainId, result }` (was a raw result).
  - The connector extracts `.signature` / `.result` before emitting.
- The `connectAndSign` and `connectWith` provider event payloads continue to expose the original `signResponse` and `connectWithResponse` keys — no breaking change for subscribers.

## Test plan

- [x] `pnpm --filter @wagmi/connectors check:types`
- [x] `pnpm vitest run --project connectors` (7/7 passing)
- [ ] Manual smoke test with the playground against `@metamask/connect-evm@1.0.0` (connect, connectAndSign, connectWith, disconnect, chain switch)

### DIFF
--- a/packages/connectors/src/metaMask.ts
+++ b/packages/connectors/src/metaMask.ts
@@ -19,24 +19,6 @@ import {
   withTimeout,
 } from 'viem'
 
-type LegacyConnectorEventPayloads = {
-  connectAndSign: {
-    accounts: readonly Address[]
-    chainId: Hex
-    signResponse: string
-  }
-  connectWith: {
-    accounts: readonly Address[]
-    chainId: Hex
-    connectWithResponse: unknown
-  }
-}
-
-type LegacyEmit = <K extends keyof LegacyConnectorEventPayloads>(
-  event: K,
-  payload: LegacyConnectorEventPayloads[K],
-) => boolean
-
 export type MetaMaskParameters = UnionCompute<
   ExactPartial<Omit<CreateEVMClientParameters, 'api' | 'eventHandlers'>> & {
     /**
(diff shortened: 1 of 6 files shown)


adonesky1 (approved):
LGTM

packages/connectors/src/metaMask.ts
```diff
@@ -111,13 +115,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
             chainId: numberToHex(currentChainId),
-            signResponse,
+            signature: signResponse,
           })
         else if (connectWithResponse)
           provider.emit('connectWith', {
             accounts,
             chainId: numberToHex(currentChainId),
-            connectWithResponse,
+            result: connectWithResponse,
```
jiexi: this technically doesn't have to be breaking at the Wagmi connector API if we keep the original event property names intact

adonesky1: Good point. I suppose there's not a real need to be breaking here on the WAGMI layer
I would support preserving the original event property names

wenfix: yeah great point: 63b93b96

adonesky1 (approved):
LGTM

packages/connectors/src/metaMask.ts
```diff
@@ -112,13 +116,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
             accounts,
             chainId: numberToHex(currentChainId),
             signResponse,
-          })
+          } as never)
```
tmm: would be nice to improve typing for `provider.emit` so we don't need to cast as `never` and lose some safety (in case the API changes in the future).

packages/connectors/src/metaMask.ts
```diff
@@ -112,13 +116,13 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
...
         else if (connectWithResponse)
           provider.emit('connectWith', {
             accounts,
             chainId: numberToHex(currentChainId),
             connectWithResponse,
-          })
+          } as never)
```
tmm: Same as above

packages/connectors/src/metaMask.ts
wenfix: thanks for the feedback! here: 63d1e4c4
was the leastest invasive solution I could find, lmk if this works!

packages/connectors/src/metaMask.ts
wenfix: replied above 🙏

PR #5077, opened 2026-04-23
fix(react): narrow generated simulate hook types

fuyua9:
## Summary
- narrow generated simulate hook call-site types to omit generated fields such as `abi`, `address`, `functionName`, and `args`
- add a dts regression for functions with matching argument shapes but different return types
- keep generated simulate hook result narrowing aligned with the recent generated read hook fix

## Testing
- pnpm bench:types packages/react/src/hooks/codegen/createUseSimulateContract.test-d.ts

### DIFF
--- a/.changeset/quiet-tigers-cheer.md
+++ b/.changeset/quiet-tigers-cheer.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/react': patch
+---
+
+Narrowed generated `useSimulateContract` hook result types for functions that share argument shapes.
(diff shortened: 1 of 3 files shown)


fuyua9:
Added a patch changeset for @wagmi/react in 965f93ed so the changeset bot can classify this generated simulate hook typing fix for release.

Local validation: git diff --check passed. I did not rerun the full type suite here because this follow-up only adds a changeset file and the worktree did not have dependencies installed.

PR #5080, opened 2026-04-24
fix(core): preserve connector signer context during prepare

deodad:
## Summary
- preserve connector signer context when preparing transactions
- align prepare-time account resolution with Wagmi's existing connector-aware transaction flows
- add regression coverage around richer local accounts and connector-driven preparation

## Why
Some connectors can provide more than just an address for the active signer. When that richer signer context is dropped during transaction preparation, downstream preparation and estimation can behave differently than the actual send path.

This change is intended to make preparation use the same signer identity that connector-aware transaction flows already rely on, so preparation stays consistent with execution for connectors that expose richer local account details.

## Notes
The goal is to preserve existing behavior for ordinary connectors while fixing the cases where the connector can provide a more specific signer representation.

### DIFF
--- a/.changeset/great-drinks-smoke.md
+++ b/.changeset/great-drinks-smoke.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/core': patch
+---
+
+Preserved connector signer context during transaction preparation so connectors that expose richer local account details prepare requests consistently with connector-aware transaction flows.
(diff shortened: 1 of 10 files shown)


jxom:
https://github.com/wevm/wagmi/commit/2058bd668ae897a3c5480cf565c5716c4d06449b should fix

Issue #3252, opened 2023-11-27
bug: useSwitchChain fails to add new chain to MetaMask


aquaticone:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Package Version

2.0.0-beta.5

### Current Behavior

1. User goes to switch to a new chain that is not configured in MetaMask
2. Request to `wallet_addEthereumChain` fails due to invalid RPC URLs

### Expected Behavior

1. User goes to switch to a new chain that is not configured in MetaMask
2. MetaMask prompts user to add the chain

### Anything else?

It looks like the chains were all updated with only `rpcUrls.default` defined (`rpcUrls.public` no longer defined).

Fix PR'd - https://github.com/wevm/wagmi/pull/3251

PR #5101, opened 2026-05-07
fix(connectors): add MetaMask chain fallback

euyua9:
## Summary

- add a MetaMask `switchChain` fallback for provider error `4902`
- request `wallet_addEthereumChain` with the configured chain metadata when the chain is missing from MetaMask
- cover the fallback with a connector unit test and add a patch changeset

This keeps the MetaMask connector aligned with the existing Base Account and Coinbase Wallet connector behavior for missing chains. Related historical reports: #3252 and #4223.

## Tests

- `pnpm test packages/connectors/src/metaMask.test.ts --run`
- `pnpm --filter @wagmi/connectors check:types`
- `pnpm exec biome check --write packages/connectors/src/metaMask.ts packages/connectors/src/metaMask.test.ts .changeset/metamask-add-chain-fallback.md`
- `git diff --check`

### DIFF
--- a/.changeset/metamask-add-chain-fallback.md
+++ b/.changeset/metamask-add-chain-fallback.md
@@ -1,0 +1,5 @@
+---
+'@wagmi/connectors': patch
+---
+
+Added a MetaMask switch-chain fallback that requests `wallet_addEthereumChain` when the target chain is missing.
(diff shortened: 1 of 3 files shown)


tmm:
MetaMask SDK handles this internally

https://github.com/MetaMask/connect-monorepo/blob/main/packages/connect-evm/src/connect.ts#L589

PR #5109, opened 2026-05-08
fix(connectors): `metaMask` — serve probe methods from EIP-6963 provider when SDK unloaded

adonesky1:
## Description

The `metaMask` connector currently routes every method through `getInstance()`, which dynamically imports `@metamask/connect-evm` on first call. Because `@wagmi/core`'s `reconnect()` calls `connector.getProvider()` (and `isAuthorized()` → `getAccounts()`) on **every** registered connector at page load, just having `metaMask()` in the config pays the SDK chunk download/parse cost on every page load — even for users who already have the MetaMask extension installed and never user-initiated a connection through this connector.

This PR teaches the read-only / probe methods (`getProvider`, `isAuthorized`, `getAccounts`, `getChainId`) to short-circuit through the EIP-6963 MetaMask provider that the connector itself observes, **only when the SDK has not yet started loading**. User-initiated `connect()` is unchanged — it still goes through the SDK, so one-tap flows (`connectAndSign`, `connectWith`), `switchChain` auto-fallback with `chainConfiguration`, MWP/mobile transports, analytics, and event wiring are preserved exactly. Once SDK loading starts (`metamaskPromise`) or completes (`metamask`), all subsequent calls go through the SDK — no provider-instance flip-flop, no event-listener splitting.

### Why this matters

This is the structural fix behind a class of bundle-size concerns we've heard from teams routing wagmi widgets to MetaMask (e.g. LiFi/Jumper). Today, a common workaround is to gate the `metaMask` connector out entirely when an injected MetaMask is detected — which means giving up everything `connect-evm` provides on the user-initiated path (one-tap `connectAndSign`/`connectWith`, automatic `wallet_addEthereumChain` fallback on unrecognized chains, a single MetaMask code path that gracefully transitions to MWP/mobile when the user opens the same widget on their phone, etc.). With this PR, dapps can keep `metaMask()` in the config without paying download cost on extension users until they actually connect via the SDK path.

#### Why #5109 is required (and trimming the chunk isn't enough)

Wagmi's `reconnect()` runs on every page load. For each registered connector, it calls `getProvider()` and then `isAuthorized()` → `getAccounts()` to ask "is this user already connected?" Today, those four methods on the `metaMask` connector all delegate to `getInstance()`, which `await import('@metamask/connect-evm')` on first call. So just having `metaMask()` in the wagmi config triggers the SDK chunk download + parse + execute on every page load — even for users who have the extension installed and never click Connect.

The related package-side work — [`MetaMask/connect-monorepo#244`](https://github.com/MetaMask/connect-monorepo/pull/244) (lazy-load MWP transport), [`MetaMask/connect-monorepo#289`](https://github.com/MetaMask/connect-monorepo/pull/289) (drop `@metamask/chain-agnostic-permission` and its `lodash`/`bn.js` chain), and [`#5107`](https://github.com/wevm/wagmi/pull/5107) (peer-dep widen so consumers pick those up) — reduces the *size* of that download but doesn't eliminate it. Extension users still pay a chunk fetch on every visit just to be told "yes, you're already authorized."

#### How #5109 accomplishes it

The connector keeps a local EIP-6963 cache: a `setup()` listener that records any `eip6963:announceProvider` event with `rdns === 'io.metamask'` (or `'io.metamask.mobile'`). The four probe methods (`getProvider`, `isAuthorized`, `getAccounts`, `getChainId`) check the cache *before* calling `getInstance()`:

- **Cache hit (extension installed)** → answer the probe directly from the injected EIP-1193 provider. `getInstance()` is never called, `import('@metamask/connect-evm')` never runs, no chunk fetch.
- **Cache miss (mobile-web, no extension)** → fall through to `getInstance()`, SDK loads as before (correct — the SDK is what handles MWP/QR).
- **User clicks Connect** → `connect()` is unchanged, always goes through `getInstance()` → SDK loads for real, all the SDK-only features (`connectAndSign`, switch-chain auto-fallback, mobile transport, event wiring) work exactly as before.

Once the SDK has started loading (`metamaskPromise` set) or finished (`metamask` set), all subsequent probes also go through the SDK — so there's no provider-instance flip-flop where some calls return the injected provider and others return the SDK-wrapped one.

**Net result for the extension-installed case**: `metaMask()` in a wagmi config costs zero `connect-evm` bytes on page load, and full SDK functionality the moment the user clicks Connect.

### How (implementation detail)

A connector-local module-level helper `watchInjectedMetaMaskProvider()` registers an `eip6963:announceProvider` listener (and dispatches `eip6963:requestProvider` to flush already-announced wallets) the first time it's called. Announcements with `rdns === 'io.metamask'` or `'io.metamask.mobile'` are cached in a `Map<rdns, EIP1193Provider>`. The listener is registered once per JS realm and is idempotent — multiple `metaMask()` connectors in the same config share the cache.

`setup()` primes the watcher; the probe methods call `findInjectedMetaMaskProvider()` (which is also defensively idempotent) to read from the cache. The probe-side gate is `!metamask && !metamaskPromise`, so the moment `connect()` starts the SDK import, all subsequent probes go through the SDK — no flip-flop window where some calls hit the injected provider and others hit the SDK-wrapped one.

Returns `undefined` on the server (`typeof window === 'undefined'`) or when no MetaMask EIP-6963 announcement has been received — in which case callers fall through to the existing `getInstance()` path, identical to today.

### Note for MetaMask reviewers — interaction with planned SDK-side EIP-6963 announcement

The MetaMask team intends to add EIP-6963 announcement to `@metamask/connect-evm` itself (mirroring the pattern already used by `@metamask/connect-solana`, which suppresses its own `wallet-standard` registration when a MetaMask extension is already registered). That work is **complementary** to this PR, not in conflict:

- This PR's shortcut prevents the `@metamask/connect-evm` chunk from loading at all on probe paths. The SDK-side announcement only matters once the SDK has loaded (e.g. after a user-initiated connect), and is about wallet-picker UX (don't double-announce).
- The wagmi `metaMask` connector already declares `rdns: ['io.metamask', 'io.metamask.mobile']`, which causes wagmi's `createConfig` to dedupe the EIP-6963 entry from the picker. So even when SDK-side announcement lands later, the picker will stay clean — wagmi already handles the dedupe based on connector `rdns`.

In short: shipping this PR doesn't paint either side into a corner. The two changes operate at different points in the lifecycle (probe vs. post-load) and rely on the existing wagmi dedupe behavior.

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| Extension installed, page-load `reconnect()` probe | `import('@metamask/connect-evm')` runs (chunk fetch + parse) | Probe answered from injected EIP-6963 provider; **no chunk load** |
| Extension installed, user clicks Connect | SDK loads, `connect()` runs through SDK | SDK loads, `connect()` runs through SDK (unchanged) |
| Extension uninstalled mid-session | SDK already loaded; `getProvider` returns SDK-wrapped provider | SDK already loaded; `getProvider` returns SDK-wrapped provider (unchanged) |
| Mobile-web (no extension), page load | SDK loads to handle MWP/QR | SDK loads to handle MWP/QR (unchanged — no `io.metamask` announcement) |
| SSR | SDK loads on hydration | SDK loads on hydration (unchanged — `window === undefined` short-circuit) |

## Diff shape

Surgical, no reformatting of unrelated code:

```
.changeset/metamask-mipd-shortcut.md  |   5 +
packages/connectors/src/metaMask.ts   |  63 +++++++++--/41 ----
packages/connectors/src/metaMask.test.ts | 147 +++++++++++++++++++++++++--/4 ----
```

## Test plan

- [x] `pnpm test --run packages/connectors/src/metaMask.test.ts` passes, with explicit assertions that `createEVMClient` is **not** called on probe-only paths:
  - pre-connect probes use the announced EIP-6963 provider (no SDK created)
  - unauthorized `reconnect()` does not load the SDK
  - falls back to SDK when no EIP-6963 MetaMask provider is announced
  - explicit `connect()` still creates and uses the SDK client
- [x] `pnpm exec biome check packages/connectors/src/metaMask.ts packages/connectors/src/metaMask.test.ts .changeset/metamask-mipd-shortcut.md` passes.
- [x] `git diff --check` passes.
- [ ] `pnpm exec tsc --noEmit -p packages/connectors/tsconfig.json` — the existing `packages/core/src/tempo` / `viem` type mismatch (introduced by #5103) still fails in this workspace; not caused by this PR.

## References

- Related to [`MetaMask/connect-monorepo#244`](https://github.com/MetaMask/connect-monorepo/pull/244) (lazy-MWP) and [`MetaMask/connect-monorepo#289`](https://github.com/MetaMask/connect-monorepo/pull/289) (drop chain-agnostic-permission) — they shrink the loaded chunk; this PR avoids loading it on the probe path entirely.
- Builds on [`#5107`](https://github.com/wevm/wagmi/pull/5107) (peer-dep widening, merged).

## Checklist

- [x] I've added tests for new functionality
- [x] I've added a changeset for changed functionality


wenfix:
Updated this to avoid relying on `config._internal.mipd`.

Connector factories only receive the connector setup slice (`emitter`, `chains`, `storage`, `transports`), not the full wagmi `Config`, so `_internal.mipd` is not available from inside `metaMask()`. The shortcut now uses a connector-local EIP-6963 listener/cache instead.

Behavior after the update:

  - `setup()` starts listening for MetaMask EIP-6963 announcements and requests providers.
  - Pre-connect probe methods (`getProvider`, `isAuthorized`, `getAccounts`, `getChainId`)
  use the announced `io.metamask` / `io.metamask.mobile` provider when the SDK has not
  started loading.
  - Once SDK loading starts (`metamaskPromise`) or completes (`metamask`), all calls stay on
  the SDK path to avoid provider flip-flop.
  - User-initiated `connect()` still always goes through `@metamask/connect-evm`.
  - If no MetaMask EIP-6963 provider is announced, the connector falls back to the existing
  SDK path.

  I also updated the changeset wording to remove the `multiInjectedProviderDiscovery: false` claim, since connectors cannot observe that core config option without changing wagmi core.

  Validation:

  - `pnpm test --run packages/connectors/src/metaMask.test.ts` passes with coverage for:
    - pre-connect probes using the announced EIP-6963 provider without creating the SDK
  client
    - unauthorized `reconnect()` avoiding SDK creation
    - fallback to SDK when no EIP-6963 MetaMask provider is announced
    - explicit `connect()` still creating/using the SDK client
  - `pnpm exec biome check packages/connectors/src/metaMask.ts packages/connectors/src/
metaMask.test.ts .changeset/metamask-mipd-shortcut.md` passes
  - `git diff --check` passes

`pnpm exec tsc --noEmit -p packages/connectors/tsconfig.json` still fails in this workspace due to the existing `packages/core/src/tempo` / `viem` type mismatch, not from the MetaMask connector changes.

packages/connectors/src/metaMask.ts
```diff
@@ -50,6 +50,53 @@ export type MetaMaskParameters = UnionCompute<
...
+          }) => { provider: unknown } | undefined
+        }
+      }
+    }
+  )._internal?.mipd
+  if (!mipd) return undefined
+  // Match the connector's own `rdns` array. Order matters: prefer the
+  // desktop extension over the mobile in-app browser when both somehow
+  // announce on the same surface.
+  for (const rdns of ['io.metamask', 'io.metamask.mobile']) {
```
ffmcgee725: should we also have the rdns for `flask` here ?

jiexi: I was also wondering this when we rewrote the connector. Specifically line 115 in this file https://github.com/adonesky1/wagmi/blob/faee00a6ce03b79b11db1a5e1cc3837a6a88e86e/packages/connectors/src/metaMask.ts#L115

packages/connectors/src/metaMask.ts
```diff
@@ -50,6 +50,72 @@ export type MetaMaskParameters = UnionCompute<
...
+
+  // Match the connector's own `rdns` array. Order matters: prefer the desktop
+  // extension over the mobile in-app browser when both somehow announce on the
+  // same surface.
+  for (const rdns of metaMaskRdns) {
+    const provider = injectedMetaMaskProviders.get(rdns)
+    if (provider) return provider
+  }
+  return undefined
+}
```
jiexi: `watchInjectedMetaMaskProvider` is async and so there's no guarantee on first call of `findInjectedMetaMaskProvider` that an injected metamask instance will be set on `injectedMetaMaskProviders` in time

jiexi: nevermind. i see `setup()` below. Does `watchInjectedMetaMaskProvider()` need to be here then?

adonesky1: yeah good call — `setup()` already primes it so the call inside `findInjectedMetaMaskProvider` is redundant. cleaner to drop it. lemme push.

adonesky1:
closing in favor of https://github.com/wevm/wagmi/pull/5128

PR #5124, opened 2026-05-16
docs: fix article and grammar errors across guides

Kropiunig:
Cleans up small grammar issues spotted while reading through the docs. Each fix is verified against parallel content in the same repo.

- `an connection` -> `a connection` in the React and Vue connect-wallet guides
- `an discussion` -> `a discussion` in the Vue viem guide (the corresponding React and core guides already use `a`)
- `a accent` -> `an accent` in the VitePress theme stylesheet comment
- `a async function` -> `an async function` in the CLI `configuring-cli` guide (both occurrences in the same sentence)
- `Vue Plugins does not work well` -> `Vue Plugin does not work well` in the Vue TypeScript guide. The same file uses singular `Vue Plugin` lower down, and the React/Solid equivalents use singular `React Context` / `Solid Context` for the parallel sentence

No source/code changes, so no changeset added (matches #4989, #4990, #5120).

### DIFF
--- a/site/.vitepress/theme/style.css
+++ b/site/.vitepress/theme/style.css
@@ -24,7 +24,7 @@
  *
  *   The soft color must be semi transparent alpha channel. This is crucial
  *   because it allows adding multiple "soft" colors on top of each other
- *   to create a accent, such as when having inline code block inside
+ *   to create an accent, such as when having inline code block inside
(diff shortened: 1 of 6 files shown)


tmm:
Superseded by 69e287c41 on main.

PR #5132, opened 2026-05-23
fix(cli): support shared foundry artifact deployments

### LINKED ISSUE #4396: bug: foundry plugin multiple addresses referencing the same ABI
### Check existing issues

### Describe the bug

At Zora, we use the [foundry plugin](https://wagmi.sh/cli/api/plugins/foundry) and love it. However, if you have multiple addresses that need to share the same ABI (e.g. different ERC20 tokens) it doesn't work.

If I am reading in a deployment file and writing the addresses like in this example: https://github.com/ourzora/zora-721-contracts/blob/main/wagmi.config.ts

Example extending that:
```
type ContractNames =
  // ...
  | 'ERC20';

type Address = `0x${string}`;

const contractFilesToInclude: ContractNames[] = [
  // ...
  'ERC20'
];

  for (const addressesFile of addressesFiles) {
    const jsonAddress = JSON.parse(
      readFileSync(`./addresses/${addressesFile}`, 'utf-8')
    ) as {
      // ... 
      WETH: Address;
      DAI: Address;
    };

    addAddress(
      'ZoraNFTCreatorV1',
      chainId,
      jsonAddress.ZORA_NFT_CREATOR_PROXY
    );
    addAddress('ERC20', chainId, jsonAddress.DAI);
    addAddress('ERC20', chainId, jsonAddress.WETH);
```
only the `daiConfig` / `daiAddress` gets created, but not the WETH config.

https://gist.github.com/0x6080/126272e246be09b24274acd3aae55bd5

### Steps To Reproduce

If I am reading in a deployment file and writing the addresses like in this example: https://github.com/ourzora/zora-721-contracts/blob/main/wagmi.config.ts

Example extending that:
```ts
type ContractNames =
  // ...
  | 'ERC20';

type Address = `0x${string}`;

const contractFilesToInclude: ContractNames[] = [
  // ...
  'ERC20'
];

  for (const addressesFile of addressesFiles) {
    const jsonAddress = JSON.parse(
      readFileSync(`./addresses/${addressesFile}`, 'utf-8')
    ) as {
      // ... 
      WETH: Address;
      DAI: Address;
    };

    // ... 
    addAddress('ERC20', chainId, jsonAddress.DAI);
    addAddress('ERC20', chainId, jsonAddress.WETH);
```
only the `daiConfig` / `daiAddress` gets created:
```ts
/**
 *
 */
export const daiAddress = {
  31337: '0xF0794f5a1318CAbf8270AFeeA33cB2b7543d1839',
} as const

/**
 *
 */
export const daiConfig = {
  address: daiAddress,
  abi: erc20Abi,
} as const
```
but **not the WETH config**.

@wagmi/cli

2.12.24

### Viem Version

2.21.35

### TypeScript Version

20

### Anything else?

Ideal behavior would be for both of them to reference the same abi, for example:
```ts
export const wethConfig = {
  address: wethAddress,
  abi: erc20Abi,
} as const
```

samsamtrum:
## Summary

Adds support for multiple named deployments sharing one Foundry ABI artifact in the CLI Foundry plugin.

## Problem

When several deployed contracts share the same ABI artifact (for example `DAI` and `WETH` both using `ERC20.json`), the Foundry plugin can currently only emit the artifact contract name. That means only one generated config/address pair is available for the shared ABI use case.

## Changes

- Add a `deploymentArtifacts` mapping from deployment name to Foundry artifact name.
- When resolving an artifact, emit additional contract entries for deployment names mapped to that artifact.
- Keep the existing `deployments` address mapping behavior intact.
- Add test coverage for two deployment names sharing one `ERC20` ABI artifact.

Example:

```ts
foundry({
  deployments: {
    DAI: '0x1111111111111111111111111111111111111111',
    WETH: '0x2222222222222222222222222222222222222222',
  },
  deploymentArtifacts: {
    DAI: 'ERC20',
    WETH: 'ERC20',
  },
})
```

This allows generated exports such as both `daiConfig` and `wethConfig` while reusing `erc20Abi`.

## Verification

- `corepack pnpm install --ignore-scripts --frozen-lockfile --filter @wagmi/cli...`
- `corepack pnpm exec biome check --write packages/cli/src/plugins/foundry.ts packages/cli/src/plugins/foundry.test.ts`
- `corepack pnpm exec vitest run packages/cli/src/plugins/foundry.test.ts -t 'contracts supports multiple deployment names sharing one ABI artifact'`
- `corepack pnpm --filter @wagmi/cli run check:types`

Closes #4396

### DIFF
--- a/packages/cli/src/plugins/foundry.test.ts
+++ b/packages/cli/src/plugins/foundry.test.ts
@@ -407,3 +407,54 @@ test('watch callbacks use broadcast deployments', async () => {
     },
   })
 })
+
+test('contracts supports multiple deployment names sharing one ABI artifact', async () => {
+  const dir = await createTempDir()
+  const spy = vi.spyOn(process, 'cwd')
+  spy.mockImplementation(() => dir)
+
+  const artifactsDir = resolve(dir, 'out')
+  await fs.mkdir(artifactsDir, { recursive: true })
+  const erc20Artifact = {
+    abi: [
+      {
+        inputs: [],
+        name: 'totalSupply',
+        outputs: [{ internalType: 'uint256', name: '', type: 'uint256' }],
+        stateMutability: 'view',
+        type: 'function',
+      },
+    ],
+  }
+  await fs.writeFile(
+    resolve(artifactsDir, 'ERC20.json'),
+    JSON.stringify(erc20Artifact, null, 2),
+  )
+
+  await expect(
+    foundry({
+      forge: { build: false, clean: false },
+      deployments: {
+        DAI: '0x1111111111111111111111111111111111111111',
+        WETH: '0x2222222222222222222222222222222222222222',
+      },
+      deploymentArtifacts: {
+        DAI: 'ERC20',
+        WETH: 'ERC20',
+      },
+    }).contracts?.(),
+  ).resolves.toMatchObject([
+    {
+      address: undefined,
+      name: 'ERC20',
+    },
+    {
+      address: '0x1111111111111111111111111111111111111111',
+      name: 'DAI',
+    },
+    {
+      address: '0x2222222222222222222222222222222222222222',
+      name: 'WETH',
+    },
+  ])
+})
--- a/packages/cli/src/plugins/foundry.ts
+++ b/packages/cli/src/plugins/foundry.ts
@@ -62,6 +62,8 @@ export type FoundryConfig = {
   includeBroadcasts?: boolean | undefined
   /** Mapping of addresses to attach to artifacts. */
   deployments?: { [key: string]: ContractConfig['address'] } | undefined
+  /** Mapping of deployment names to Foundry artifact names. */
+  deploymentArtifacts?: { [key: string]: string } | undefined
   /** Artifact files to exclude. */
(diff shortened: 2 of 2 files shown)


promiseeuler (changes requested):
While testing the new shared-artifact path locally on commit `86c111e`, I found that watch mode only updates the original artifact contract, not the deployment aliases.

The initial `contracts()` call correctly returns `ERC20`, `DAI`, and `WETH`. After rewriting `ERC20.json`, however, `plugin.watch.onChange(artifactPath)` returns only:

```ts
{
  name: 'ERC20',
  abi: [{ name: 'after', /* ... */ }],
}
```

This comes from both callbacks selecting `(await getContract(path))[0]`. In `generate --watch`, that updates only the `ERC20` entry in `contractMap`; the generated `DAI` and `WETH` configs keep the previous ABI until the process is restarted.

Could the watch contract support returning all resolved aliases (and remove all of them on unlink), with a regression test that changes the shared artifact after initial resolution? The current focused test passes, as does `@wagmi/cli check:types`; the additional watch regression fails because it receives the single `ERC20` object instead of the three expected updates.

PR #5151, opened 2026-06-08
chore: upgrade @metamask/connect-evm to v2

wenfix:
## Description

Upgrades `@metamask/connect-evm` from `1.3.1` to `2.1.0` (a major release) in `@wagmi/connectors`, and sets `skipAutoAnnounce: true` on the `metaMask` connector.

`skipAutoAnnounce: true` prevents the MetaMask SDK from announcing its own EIP-1193 provider via EIP-6963. wagmi already discovers the native MetaMask provider through MIPD/EIP-6963 (the connector's pre-connect fast path relies on it), so letting the SDK announce a second time would register a duplicate, competing provider entry. Forcing it on is the correct behavior for this connector.

## Changes

- **`@metamask/connect-evm` → `2.1.0`** — catalog pin (`pnpm-workspace.yaml`) and peer range (`packages/connectors/package.json`) bumped to `^2.1.0`.
- **`skipAutoAnnounce: true`** set on the `metaMask` connector's `createEVMClient(...)` call, placed after the `...parameters` spread so it's always enforced and can't be overridden by callers.
- **`getProvider()` cast fix** — connect-evm v2 changes the exported `EIP1193Provider` from a structural type to a concrete class (`extends EventEmitter`, private fields), so the structurally-compatible injected EIP-6963 provider now needs a cast through `unknown`.
- Added a changeset (`major` for `@wagmi/connectors`).

## Breaking changes handled (connect-evm v2)

| Breaking change | Impact | Handling |
| --- | --- | --- |
| `EIP1193Provider` is now a concrete class | `getProvider()` cast no longer type-checks | Cast via `unknown` |
| Removed `transport.onNotification` from `createEVMClient()` | None — connector doesn't use it; `MetaMaskParameters` is derived from `Parameters<typeof createEVMClient>[0]`, so it drops out automatically | No action |
| Removed public `transport` accessor | None — connector doesn't read `instance.transport` | No action |

## Note on `@metamask/connect-multichain`

connect-evm `2.0.0` briefly made `@metamask/connect-multichain` a **required peer dependency**. That was reverted in [MetaMask/connect-monorepo#323](https://github.com/MetaMask/connect-monorepo/pull/323) and shipped in **`2.1.0`**, where connect-multichain is once again a bundled (regular) dependency of connect-evm. So consumers of the `metaMask` connector only need `@metamask/connect-evm` — no separate peer install.

> An earlier revision of this PR (targeting `2.0.0`) declared connect-multichain as an optional peer dependency and added catalog/knip/docs wiring for it. That has been removed now that `2.1.0` bundles it again.

## Consumer migration

Apps using the `metaMask` connector install only:

```bash
npm install @metamask/connect-evm@^2.1.0
```

## Notes for reviewers

- `skipAutoAnnounce` is hard-forced and omitted from `MetaMaskParameters` (via `Omit`), so callers cannot override it. wagmi already surfaces the native MetaMask provider via MIPD — this is intentional.
- The `injected as unknown as EIP1193Provider` double-cast is required by the v2 class-based provider type; it's commented inline.
- Lockfile: `connect-evm@2.1.0` pulls in `@metamask/connect-multichain@1.1.0` transitively (along with its `centrifuge`/`protobufjs` mobile-wallet-protocol transport deps). `@metamask/connect-multichain` is kept in `minimumReleaseAgeExclude` so the freshly-published `1.1.0` resolves.

## Testing

- `pnpm check:types` (connectors)
- `pnpm vitest run --project connectors metaMask` (the `forces skipAutoAnnounce` test covers the connector behavior)

Connector source is unchanged from the previously-passing revision; only the dependency version moved `2.0.0` → `2.1.0` (same v2 API surface). CI re-verifies.


.changeset/bump-connect-evm-2-0-0.md
```diff
@@ -0,0 +1,11 @@
+---
+"@wagmi/connectors": minor
```
adonesky1: I think this is `major`?

.changeset/bump-connect-evm-2-0-0.md
```diff
@@ -0,0 +1,11 @@
...
+---
+
+Updated `@metamask/connect-evm` to `2.0.0` and set `skipAutoAnnounce: true` on the `metaMask` connector to prevent the SDK from announcing its EIP-1193 provider via EIP-6963.
+
+`@metamask/connect-evm` v2 makes `@metamask/connect-multichain` a required peer dependency (it was previously installed transitively). If you use the `metaMask` connector, install it alongside `@metamask/connect-evm`:
```
adonesky1: This is breaking

packages/connectors/src/metaMask.ts
```diff
@@ -326,6 +329,7 @@ export function metaMask(parameters: MetaMaskParameters = {}) {
 
           metamaskPromise = createEVMClient({
             ...parameters,
+            skipAutoAnnounce: true,
```
adonesky1, line 339: Since this is hard-forced after the `...parameters` spread, a caller who passes `skipAutoAnnounce: false` gets silently ignored. I think this is what we want based on our chats... But it's still a typed, settable option — `MetaMaskParameters` only omits `'api' | 'eventHandlers'` from `CreateEVMClientParameters` (line 23), so `skipAutoAnnounce` stays in the public type as a no-op.

Worth adding it to the `Omit` so the type matches the actual behavior?

 ```ts
ExactPartial<Omit<CreateEVMClientParameters, 'api' | 'eventHandlers' | 'skipAutoAnnounce'>>
 ```

.changeset/bump-connect-evm-2-0-0.md
jiexi: indeed this is a breaking change

.changeset/bump-connect-evm-2-0-0.md
wenfix: 19cbd57d8e3de2ac1fdaee08cdce359719049aa0

.changeset/bump-connect-evm-2-0-0.md
wenfix: 19cbd57d8e3de2ac1fdaee08cdce359719049aa0

packages/connectors/src/metaMask.ts
wenfix: 9c8056bfd919e35868fd282a4635775855b13443

jiexi (approved):

adonesky1 (approved):

.changeset/bump-connect-evm-2-0-0.md
tmm: Marked as patch since I would consider the internal behavior changes a fix for undesirable behavior. The major version dep bump normally would be breaking, but doesn't look like there would be runtime issues if someone is using an older version and we alert folks to this in the docs.

PR #5160, opened 2026-06-20
Fix accounts peer dependency range

### LINKED ISSUE #5145: accounts peer dependency conflict with accounts@^0.14
## Description

`npm install` fails when project has `accounts@^0.14.5` alongside `wagmi@^3.6.16`.

## Error
```
ERESOLVE: @wagmi/core@3.5.0 peerOptional accounts@~0.12
Found: accounts@0.14.5
```

## Environment
- wagmi: ^3.6.16
- @wagmi/core: 3.5.0
- accounts (root): ^0.14.5

## Question
Should @wagmi/core support accounts 0.14, or document that accounts must not be added separately at incompatible versions?

yintaisha:
Fixes #5145.

This updates the optional `accounts` peer dependency range for `@wagmi/core` and `@wagmi/connectors` to allow `accounts@0.14`.

Projects that install `accounts@^0.14` alongside wagmi currently hit npm peer resolution errors because `@wagmi/core` only accepts `~0.12` and `@wagmi/connectors` only accepts `~0.10`.

### DIFF
--- a/.changeset/tidy-berries-cheer.md
+++ b/.changeset/tidy-berries-cheer.md
@@ -1,0 +1,6 @@
+---
+"@wagmi/connectors": patch
+"@wagmi/core": patch
+---
+
+Updated the optional `accounts` peer dependency range to allow `accounts@0.14`.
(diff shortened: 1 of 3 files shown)


tmm:
Went with https://github.com/wevm/wagmi/commit/3e12a5e43e050101123afeedf44333373f03e9bf instead since older versions of `accounts` package can drop compatibility.

PR #5202, opened 2026-07-17
fix(tempo): await accounts import instead of chaining .catch()

### LINKED ISSUE #5201: tempo: `getAccountsModule` chains `.catch()` on `import()`, breaking under bundlers where dynamic import isn't a spec Promise (e.g. Metro/React Native)
### Describe the bug

The Tempo `webAuthn`/`tempoWallet`/`dangerous_secp256k1` connectors in `@wagmi/core`'s `tempo` module lazily load the peer-installed `accounts` package via:

```ts
async function getAccountsModule() {
  return await import('accounts').catch(() => {
    throw new Error('dependency "accounts" not found')
  })
}
```

Chaining `.catch()` directly on the return value of a dynamic `import()` assumes that value is always a spec-compliant native `Promise`. That's not guaranteed in every bundling environment — notably under Metro (React Native's bundler), particularly with eager/non-lazy bundling (`?lazy=false`, which some setups force to work around unrelated issues), the object returned by a transformed `import()` call reliably has `.then()` but does **not** have a `.catch()` method.

The result is a very misleading failure: instead of surfacing "dependency accounts not found" (or succeeding, since `accounts` *is* installed and resolvable), the app crashes with:

```
TypeError: undefined is not a function
```

thrown from evaluating `import('accounts').catch` itself — before the import even settles, and before the intended `catch` handler's custom error message ever has a chance to run. This makes the connector effectively unusable in this class of environment, and the actual error message is unhelpful for diagnosing why.

Confirmed via runtime introspection in the affected environment:

```js
const importResult = import('accounts')
typeof importResult        // "object"
typeof importResult.then   // "function"
typeof importResult.catch  // "undefined"   <-- crashes here
```

### Link to Reproduction

No hosted repro (this reproduces specifically under Metro/React Native bundling, which isn't something new.wagmi.sh can host). Minimal repro steps below reproduce it in any React Native + Expo + Metro project.

### Steps To Reproduce

1. In a React Native (Expo) app, install `wagmi`, `@wagmi/core`, and `accounts`, and configure a Tempo connector, e.g.:
   ```ts
   import { createConfig } from 'wagmi'
   import { webAuthn } from 'wagmi/tempo'

   export const config = createConfig({
     chains: [tempoModerato],
     connectors: [webAuthn({ ceremony: someCeremony })],
     transports: { [tempoModerato.id]: http() },
   })
   ```
2. Run the app on a real device/simulator via Metro (`expo run:ios` / `expo run:android`), not a web/Node target.
3. On startup, `createConfig` → `createStore` → connector `setup()` → `getProvider()` → `getAccountsModule()` throws `TypeError: undefined is not a function` instead of resolving the `accounts` module (or throwing the intended, readable error).

### Affected package(s)

@wagmi/core

### Package version(s)

`@wagmi/core@3.4.4` through the current latest `@wagmi/core@3.6.3` — the code at `packages/core/src/tempo/Connectors.ts`'s `getAccountsModule` is unchanged in this respect as of PR #5165 (which touched the same function for an unrelated Turbopack issue, #5144, but didn't change the `.catch()` chaining).

### Current behavior

Chaining `.catch()` directly on `import('accounts')` throws `TypeError: undefined is not a function` in bundling environments (e.g. Metro/React Native, especially eager/non-lazy bundling) where the value returned by a transformed dynamic `import()` has `.then()` but no `.catch()` method — masking both the success path and the intended "dependency not found" error message.

### Expected behavior

`getAccountsModule` (and any other `import(...).catch(...)` call sites in the Tempo connectors) should not assume the dynamic `import()` return value is a fully spec-compliant `Promise`. Wrapping in `try { return await import('accounts') } catch { throw new Error('dependency "accounts" not found') }` instead of chaining `.catch()` fixes this — `await` only requires `.then()`, which is present in all the environments observed (including Metro), and behaves identically to the current code in standard Node/browser/Vite/webpack/Turbopack environments where `import()` does return a real `Promise`.

### Anything else?

Environment where this was diagnosed: Expo (React Native) app bundled by Metro, both with the default lazy bundling and with eager (`lazy=false`) bundling forced via a custom `rewriteRequestUrl` — reproduces in both, and is not specific to any particular Expo SDK/React Native version tested (confirmed across two consecutive SDK majors). Not a TypeScript type-checking issue — types report the code as correct; this is a runtime-only failure specific to how `import()` gets transformed by non-Node/browser bundlers.

singhlovepreet9:
Fixes #5201.

### Problem

The Tempo connectors (`webAuthn` / `tempoWallet` / `dangerous_secp256k1`) in `@wagmi/core` lazily load the peer-installed `accounts` package with:

```ts
async function getAccountsModule() {
  return await import(
    /* turbopackOptional: true */
    'accounts'
  ).catch(() => {
    throw new Error('dependency "accounts" not found')
  })
}
```

Chaining `.catch()` directly on the result of `import()` assumes that value is always a spec-compliant native `Promise`. That isn't guaranteed in every bundling environment — notably under Metro (React Native's bundler), a transformed `import()` reliably exposes `.then()` but **not** `.catch()`. Evaluating `import('accounts').catch` then throws `TypeError: undefined is not a function` before the import settles and before the intended `'dependency "accounts" not found'` message can run, making the connectors unusable there with a misleading error.

### Fix

Await the dynamic import inside a `try`/`catch` instead of chaining `.catch()` on it. `await` only requires a thenable's `.then()`, so it works with Metro's non-spec import result while remaining identical for spec Promises. The `/* turbopackOptional: true */` magic comment and the rethrown error message are preserved.

### Verification

- `biome check` on the changed file — clean.
- `tsc --noEmit` (core) and the `build:esm+types` build — both pass.
- Mechanism reproduced standalone (a thenable with `.then` but no `.catch`, mirroring Metro):
  - old path `thenable.catch(...)` → `TypeError: ...catch is not a function`
  - new path `await thenable` → resolves normally

### Note for reviewers (tests)

The template asks for a test that fails without the fix. I was unable to add one that's meaningful in CI: under Node/Vitest, `import()` always returns a spec-compliant `Promise` (with `.catch`) and `accounts` is installed in the workspace, so both the old and new code behave identically in the test environment — the divergence only appears under Metro's non-spec `import()` transform, which the repo's test runner can't reproduce. The change is a behavior-preserving 3-line refactor for spec Promises; happy to add a test if you can suggest a way to simulate the Metro import shape within the existing suite.

### DIFF
--- a/.changeset/tempo-accounts-import-catch.md
+++ b/.changeset/tempo-accounts-import-catch.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/core": patch
+---
+
+Fixed the Tempo connectors crashing with `TypeError: undefined is not a function` when lazily loading the `accounts` peer dependency under bundlers whose transformed `import()` does not return a spec-compliant `Promise` (e.g. Metro/React Native). The result of `import()` is now awaited instead of having `.catch()` chained onto it.
--- a/packages/core/src/tempo/Connectors.ts
+++ b/packages/core/src/tempo/Connectors.ts
@@ -188,12 +188,14 @@ function _setup(parameters: setup.Parameters) {
     let disconnect: ((error?: Error | undefined) => void) | undefined
 
     async function getAccountsModule() {
-      return await import(
-        /* turbopackOptional: true */
-        'accounts'
-      ).catch(() => {
+      try {
+        return await import(
+          /* turbopackOptional: true */
+          'accounts'
+        )
+      } catch {
         throw new Error('dependency "accounts" not found')
-      })
+      }
     }
 
     async function getProvider() {


ostapondo:
I independently hit this and ended up with the exact same fix before noticing this PR (closed mine, #5209, as a duplicate). While verifying it I wrote a regression test that reproduces the Metro failure in the browser-mode suite — it simulates the then-only thenable by removing `Promise.prototype.catch` during the synchronous evaluation of the import, and fails on `main` with the same `TypeError`. Feel free to pull it into this PR if useful: https://github.com/ostapondo/wagmi/blob/b06910f45909ade6f98d258873e861e88fa7d20c/packages/core/src/tempo/Connectors.test.ts#L109-L142. One caveat worth knowing: `createConfig` calls `connector.setup()` eagerly, so the prototype patch has to wrap `connectors.setup(...)` itself.

tmm:
Superseded by #5216, which includes the same fix plus a Metro regression test.

PR #5210, opened 2026-07-30
Fix async injected provider cleanup

grandpig:
## Summary

- consolidate the async injected provider event and timeout paths into a single settlement path
- remove the exact `ethereum#initialized` listener and clear the timeout once detection settles
- prevent the losing path from querying the provider after authorization has already completed
- add regression tests for both timeout-first and event-first behavior

Previously, the listener registered for `ethereum#initialized` was an anonymous wrapper, while cleanup attempted to remove a different function reference. If the timeout settled first and the event was never emitted, the listener remained attached. If the event settled first, the uncancelled timeout could later perform another provider lookup.

The new settlement path cleans up the listener and timeout before querying the provider and guards against concurrent settlement.

I have tested by the following cmd:

- `pnpm vitest run packages/core/src/connectors/injected.test.ts --project core`
- `pnpm --filter @wagmi/core check:types`
- `pnpm biome check packages/core/src/connectors/injected.ts packages/core/src/connectors/injected.test.ts .changeset/calm-ravens-listen.md`
- `pnpm changeset status`

### DIFF
--- a/.changeset/calm-ravens-listen.md
+++ b/.changeset/calm-ravens-listen.md
@@ -1,0 +1,5 @@
+---
+"@wagmi/core": patch
+---
+
+Fixed async injected provider detection retaining its initialization listener and timeout after authorization settled.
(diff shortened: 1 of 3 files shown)


grandpig:
@tmm Hi! Would you mind taking a look at this when you have a chance? It fixes cleanup of the async injected-provider detection path and adds regression tests for both timeout-first and event-first cases. Thanks!

grandpig:
@tmm Could you please take a look when you have a chance? Thanks!

PR #5226, opened 2026-08-20
fix(react): forward includeTransactions and blockTag to useBlock watcher

Kropiunig:
### Problem

`useBlock` builds its query key with the caller's `includeTransactions` and `blockTag`, but only forwards `config` and `chainId` into `useWatchBlocks`:

```ts
const options = getBlockQueryOptions(config, {
  ...parameters,
  chainId: parameters.chainId ?? chainId,
})
useWatchBlocks({
  ...({
    config: parameters.config,
    chainId: parameters.chainId!,
    ...(typeof parameters.watch === 'object' ? parameters.watch : {}),
  } as UseWatchBlocksParameters),
  ...
  onBlock(block) {
    queryClient.setQueryData(options.queryKey, block)
  },
})
```

The watcher therefore fetches with viem's defaults (`includeTransactions: false`, `blockTag: 'latest'`) and writes that block straight into a query key that promised something else.

For `useBlock({ includeTransactions: true, watch: true })` this means the initial fetch returns full `Transaction` objects, and then the very next block silently replaces `data.transactions` with an array of hash strings — while the return type still says `Transaction[]`. The same mismatch applies to `blockTag: 'safe' | 'finalized'` combined with `watch`.

### Fix

Forward `blockTag` and `includeTransactions` into `useWatchBlocks`, placed before the `watch` object spread so an explicit `watch` object still takes precedence:

```ts
blockTag: parameters.blockTag,
includeTransactions: parameters.includeTransactions,
```

viem's `watchBlocks` resolves both defaults internally (`includeTransactions_ ?? false`, `blockTag = 'latest'`), so passing `undefined` keeps the current behaviour when the parameters are not set.

### Test

Added `parameters: watch (includeTransactions)` to `packages/react/src/hooks/useBlock.test.ts`. The existing `parameters: watch` test only covers `watch: true` without `includeTransactions`, so this case was not exercised.

On `main` the new test fails with:

```
AssertionError: expected 'string' not to be 'string'
 at packages/react/src/hooks/useBlock.test.ts:98:37
```

With the fix applied it passes.

### DIFF
--- a/.changeset/hungry-donuts-shave.md
+++ b/.changeset/hungry-donuts-shave.md
@@ -1,0 +1,5 @@
+---
+"wagmi": patch
+---
+
+Forwarded `includeTransactions` and `blockTag` to the block watcher in `useBlock` so watched updates match the query parameters
--- a/packages/react/src/hooks/useBlock.test.ts
+++ b/packages/react/src/hooks/useBlock.test.ts
@@ -1,5 +1,6 @@
-import { testClient } from '@wagmi/test'
+import { accounts, testClient } from '@wagmi/test'
 import { renderHook } from '@wagmi/test/react'
+import { parseEther } from 'viem'
 import { expect, test, vi } from 'vitest'
 
 import { useBlock } from './useBlock.js'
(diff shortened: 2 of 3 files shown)


promiseeuler (commented):
While reviewing this fix on commit `9cdca015`, I confirmed that forwarding `includeTransactions` and `blockTag` before the `watch` object spread preserves explicit watcher overrides and keeps watched cache updates aligned with the query parameters. The two changed TypeScript files also pass Biome and `git show --check` is clean.

I could not complete the focused browser test locally: the repository test endpoint returned HTTP 400 for `anvil_mine`, and the PR currently has failing `Verify / Test (3, 4)` and `Verify / Check` jobs. Leaving this as a comment rather than an approval until those required checks are green.

PR #5227, opened 2026-08-22
docs: clarify paymaster receipt identifiers

### LINKED ISSUE #4626: useWaitForTransactionReceipt returning null for transaction using paymasterService
### Check existing issues

- [x] I checked there isn't [already an issue](https://github.com/wevm/wagmi/issues) for the bug I encountered.

### Describe the bug

I have a checkout flow that supports sponsoring the gas using the paymasterService. 
When I initiate a transaction using `writeContractsAsync` I extract the hash afterwards and add it to my transaction object, which is then used to feed the transaction hash to the `useWaitForTransactionReceipt` hook. 

After that's done, the transaction is stuck pending, never returns an error. Network tab confirms it by endless null results from the RPC calls. 

I have made sure that the hash is valid (looking it up on basescan was successful), and that the chain given to the hook matches the one transaction was done on.

**Example of the RPC calls:**

Image
Image

### Expected behaviour
Executing transactions using `walletClient.writeContract` works perfectly fine, and would expect to work the same when using `writeContractAsync`

https://stackblitz.com/edit/new-wagmi-fkukg35a

### Steps To Reproduce

Connect to an account abstracted wallet (I use coinbase smart wallet).

Make sure you have added a valid paymaster link and have verified the 0x41Aa48Ef3c0446b46a5b1cc6337FF3d3716E2A33 contract for it.

Initiate the transaction and check the console where the states from useWaitForTransactionReceipt will be displayed, and check the network tab to see the coresponding RPC calls-

wagmi

2.14.16

### Viem Version

2.26.3

### TypeScript Version

5.3.3

### Anything else?

_No response_

ayaanoncrypto:
Fixes #4626

## What is this PR solving?

`useWaitForTransactionReceipt` and `waitForTransactionReceipt` expect an on-chain transaction hash. EIP-5792 `writeContracts` returns a call bundle identifier, and paymaster-backed flows may return a UserOperation identifier. Passing either identifier to the transaction receipt APIs leads to repeated `null` results because the identifier is not an L1 or L2 transaction hash.

This PR adds matching warnings to the React hook and core action documentation. The warning directs users to `useWaitForCallsStatus` or `waitForCallsStatus`, whose `receipts` field exposes the on-chain transaction receipts for batched calls.

## Scope

The change is documentation-only. It does not change receipt polling behavior or runtime APIs. It records the supported distinction between transaction hashes and call bundle or UserOperation identifiers.

## Validation

- `git diff --check` passes.
- - The two documentation pages use the repository’s existing warning admonition syntax.
- - The commit is GPG verified.
-

### DIFF
--- a/site/core/api/actions/waitForTransactionReceipt.md
+++ b/site/core/api/actions/waitForTransactionReceipt.md
@@ -8,6 +8,10 @@ const typeName = 'WaitForTransactionReceipt'
 
 Action that waits for the transaction to be included on a block, and then returns the transaction receipt. If the transaction reverts, then the action will throw an error. Replacement detection (e.g. sped up transactions) is also supported.
 
+::: warning
+This action expects an on-chain transaction hash. It does not accept an EIP-5792 call bundle ID or a UserOperation hash returned by `writeContracts`, including calls submitted with a paymaster. For batched calls, use [`waitForCallsStatus`](/core/api/actions/waitForCallsStatus) with the returned ID. Its `receipts` field contains the on-chain transaction receipts.
+:::
+
 ## Import
 
 ```ts
--- a/site/react/api/hooks/useWaitForTransactionReceipt.md
+++ b/site/react/api/hooks/useWaitForTransactionReceipt.md
@@ -15,6 +15,10 @@ const TError = 'WaitForTransactionReceiptErrorType'
 
 Hook that waits for the transaction to be included on a block, and then returns the transaction receipt. If the transaction reverts, then the action will throw an error. Replacement detection (e.g. sped up transactions) is also supported.
 
+::: warning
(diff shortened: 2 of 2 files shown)


promiseeuler (approved):
While reviewing this documentation change locally on commit `8f2c7b11`, I verified that both linked targets exist and that the `waitForCallsStatus` documentation exposes the stated `receipts: TransactionReceipt[]` result. The warning is consistent across the core action and React hook and directly addresses the identifier mismatch reported in #4626.

Validation:
- `git show --check HEAD` passed.
- `pnpm --filter site build` completed in 104.68s.
- The site build emitted existing unrelated sidebar-link warnings and floating-vue SSR diagnostics outside these two changed pages; neither new link was among the missing targets.

PR #5240, opened 2026-09-06
fix(core): do not strand reconnect when isAuthorized rejects

### LINKED ISSUE #5233: reconnect() strands status at "reconnecting" when connector.isAuthorized() rejects
### Summary

`reconnect()` awaits `connector.isAuthorized()` without a `.catch()`. If that promise rejects, the rejection escapes the function. The store is left at `status: 'reconnecting'` permanently, and the module-level `isReconnecting` guard is never reset, so every later `reconnect()` call returns `[]` and can never recover.

The two awaits either side of it are already guarded, so this looks like an oversight rather than a deliberate choice:

```js
const provider = await connector.getProvider().catch(() => undefined)
...
const isAuthorized = await connector.isAuthorized()      // not guarded
...
const data = await connector.connect({ isReconnecting: true }).catch(() => null)
```

A WalletConnect connector can reject here on a stale session or an unreachable relay, so this is reachable in normal use.

### Why it does not recover

`isReconnecting` is a module-scoped variable. It is set to `true` on entry and reset to `false` only on the success path. A throw skips the reset, so the early return at the top of `reconnect()` matches forever afterwards. Callers cannot reach that variable, so no retry helps.

`config.state.current` is still set from storage, so `getAccount()` keeps returning an address. In the `'reconnecting'` branch `getAccount` reports `isConnected: !!address`, which means consumers see a connected account that never finishes connecting.

### Second site

`createConfig`'s `_internal.revalidate()` has the same unguarded call. A rejection there skips the closing `store.setState`, so stale connections are never cleaned up.

### Reproduction

```ts
const config = createConfig({
  chains: [sepolia],
  connectors: [() => ({
    // ...standard connector shape
    isAuthorized: async () => {
      throw new Error('stale session')
    },
  })],
  transports: { [sepolia.id]: custom({ request: async () => null }) },
})

config.setState((x) => ({ ...x, current: 'flaky-uid' }))

await reconnect(config)                 // rejects
config.state.status                     // 'reconnecting', and stays there

// a brand new healthy config in the same module instance:
await reconnect(healthyConfig)           // returns [], never reaches 'connected'
```

`connect()` is unaffected, so an explicit user-initiated connect still works. That is the only recovery path.

### Versions

Reproduced on `@wagmi/core` 2.22.1. The same two unguarded calls are present in 3.6.4, at `dist/esm/actions/reconnect.js` and `dist/esm/createConfig.js`.

### Impact for us

This froze a payment page on its loading skeleton. Our balance query was gated on the connection state, and the page had no way out because reconnect could not recover. We are running a local patch that adds `.catch(() => false)` to both calls.

### Suggested fix

Treat a rejection the same way an unauthorized result is treated:

```js
const isAuthorized = await connector.isAuthorized().catch(() => false)
```

Resetting `isReconnecting` in a `finally` block would also help, so one bad connector cannot disable reconnect for the page lifetime.

devtechedge:
## Summary

reconnect() already guards getProvider() and connect() with .catch(), but isAuthorized() was awaited bare. If that promise rejects, two things go wrong:

1. The rejection escapes reconnect(), so the status cleanup at the bottom never runs and the store can stay on reconnecting / connecting.
2. The module-level isReconnecting flag is never cleared, so every later reconnect() call returns [] immediately and cannot recover.

This showed up in #5233.

## Changes

- Treat a rejected isAuthorized() like an unauthorized connector (false), matching the getProvider() pattern.
- Clear isReconnecting in a finally block so the guard resets even if something else throws later.
- Add a unit test that mocks isAuthorized to reject, asserts status ends disconnected, then confirms a second reconnect() can run again.

## Test plan

- [x] Added behavior test in packages/core/src/actions/reconnect.test.ts
- [ ] pnpm test for @wagmi/core reconnect tests (please run in CI)

Fixes #5233


devtechedge:
Just floating this back up 👀 in case it slipped through.

It keeps reconnect from stranding when isAuthorized rejects.

Open since 6 September with green checks and no conflicts. Open to any changes you want.

